PR #114, opened 2014-01-24
add Hebrew po

GreenLunar:
Hebrew translation for Converse.js.
Only .po file is provided herein.


GreenLunar:
> Edit index.html to use **he** and not **en** and then open it in your browser.

It does not work for me anymore.

jcbrand:
@GreenLunar Please define "does not work". No he translations at all, or only not the new ones? The new ones are not yet available in the master branch (will be soon).

GreenLunar:
converse.js does not appear.

I use index.html (just that file alone) with "en" renamed to "he".

PR #302, opened 2015-01-05
Add fullname and jid to contact's tooltip in roster

gbonvehi:
It seems title attributes are limited to 512 chars, we will have
to live with that unless a custom tooltip is used.


gbonvehi:
If you're ok with the merge I'll modify the changes file too :)

jcbrand:
Looks good! Please update the changes file and then I'll merge.

gbonvehi:
Done

PR #304, opened 2015-01-12
Polish translation of version 0.8.6.

ser:
Thanks for Converse.js, Jan-Carel <3


jcbrand:
You're welcome and thanks for the translations :)

PR #301, opened 2015-01-05
Added new Makefile.win, it only has dev/build/check targets

gbonvehi:
Gruntfile now appends .cmd extension to r.js if platform is win32
Made touch-npm/bower grunt task/targets to be multiplatform (depends on node/grunt)
make -f Makefile.win to build on Windows

PR #308, opened 2015-01-22
Added new Makefile.win, it only has dev/build/check targets

gbonvehi:
Gruntfile now appends .cmd extension to r.js if platform is win32
Made touch-npm/bower grunt task/targets to be multiplatform (depends on node/grunt)
make -f Makefile.win to build on Windows

This is PR #301 with docs added and moved to a new branch.


jcbrand:
Super, thanks!

PR #309, opened 2015-01-22
Strophe.log and .error now calls converse.log, this honors converse.debu...

gbonvehi:
Currently converse.js is very verbose as Strophe.log is called everywhere and it simply outputs the content to console even if converse.debug === false


jcbrand:
Thanks!

Issue #305, opened 2015-01-16
presence/show text in XMPP request isn't allowed by specification


yurivict:
I spotted the request when show text is 'online', but section '2.2.2.1. Show' in rfc#3921 only allows away/chat/dnd/xa values there.

    <body rid='2984872344' xmlns='http://jabber.org/protocol/httpbind' sid='b818817d-9882-dc5c-83ef-5d0010b95ac1'>
       <presence xmlns='jabber:client'>
         <show>online</show>
         <status>xx</status>
      </presence>
      <iq type='get' xmlns='jabber:client' id='4:sendIQ'>
        <vCard xmlns='vcard-temp'/>
      </iq>
    </body>

gbonvehi:
@yurivict which version are you using? I've just tested sendPresence method and it works correctly, you can see the relevant code here https://github.com/jcbrand/converse.js/blob/master/converse.js#L4295 . As shown there, if type === 'online' it will just create the presence without adding a show node.

gbonvehi:
Actually, just confirmed this bug, I'll debug where it's setting that show node.

jcbrand:
Thanks for fixing this @gbonvehi

PR #311, opened 2015-01-23
Fix show online presence issue #305

gbonvehi:
It fixes the problem by calling sendPresence method instead of using it's own presence message.
Added a new test file, let me know if this should be instead inside of an already existing one.


jcbrand:
Great job, thanks!

PR #340, opened 2015-03-17
Hide 'Server' Label when hiding server input

pzia:
hide_muc_server hides the input box but not the 'Server' label.
Of course, it's possible to patch the css for this purpose, but the gui should conform to the hide_muc_server directive.

Note : I didn't test it as I don't have time to setup a build environnement, but the change is simple enough.


pzia:
Screenshot for the record (see the meaningless 'Server' label when it is alone)

capture du 2015-03-17 11 35 03

jcbrand:
Nice one, thanks!

PR #349, opened 2015-03-23
Color for visitors

c143:
I added a color for visitors and deleted the min css file.


jcbrand:
Thanks. I've cherry-picked the commit with the color change. I'm not going to remove the minified CSS file.

PR #354, opened 2015-03-29
contacts.add API to register contacts

pzia:
- Basic test if the given jid contains @
- use jid as name if undefined

This is just a follow up for having chats.open() API.

### DIFF
--- a/converse.js
+++ b/converse.js
@@ -5484,7 +5484,15 @@
                 return _.map(jids, _transform);
             },
             'add': function (jid, name) {
-                if (jid.indexOf('@') < 0) return false;
+                if (typeof jid === "undefined") {
+                    throw new Error("Error: you must supply a jid");
+                }
+                if (typeof jid !== "string") {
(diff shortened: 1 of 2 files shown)


pzia:
Builds fails, but this doesn't seems related to the code itself.

jcbrand:
Thanks for the pull request.

Instead of returning ``false`` I would rather throw an exception when an invalid jid is provided. The ``if`` statement should also check if ``jid`` is a string before using ``indexOf``.

I think the build is failing because Github is under DDOS attack so connectivity to stuff hosted there is not great.

pzia:
OK, I will cancel this, and make a better one.

jcbrand:
@piza: I don't think you have to cancel it. You can add more commits to this pull request. Just continue pushing to your branch.

pzia:
OK, done (just discovering github magic).

There are some inconsistencies in throwing exception. Some parts of the
code use throw("... "),  and some use throw new Error(...).
I went with the second one.

jcbrand:
Thanks!

PR #356, opened 2015-04-01
Fix the plugin extend function

floriancargoet:
I tried to use the `converse.plugins.extend` function but there was a bug. To check if a function was been added to the extended object, a `typeof key ===  'function'` was tested instead of `typeof value === 'function'`.


jcbrand:
Nice one, thanks!

PR #357, opened 2015-04-01
Fix the known bug where a state notification reopens a chat box

floriancargoet:
When a state notification is received, a chat box is opened. If the notification is the result of the other person closing the chat box, it's annoying. User 1 closes a box, user 2 sees a box auto opening. User 2 closes the new box, user 1 sees a box auto opening. This has no end.

It's a known bug since there's a FIXME comment in the code.

I propose this fix: only open a chat box when the message has a body (i.e. a real message).

### DIFF
--- a/converse.js
+++ b/converse.js
@@ -3119,12 +3119,15 @@
 
                 chatbox = this.get(contact_jid);
                 if (!chatbox) {
-                    /* FIXME: there is a bug here. If chat state notifications
-                     * (because a roster contact closed a chat box of yours
-                     * they had open) are received and we don't have a chat with
-                     * the user, then a chat box is created here which then
-                     * opens automatically :(
+                    /* If chat state notifications (because a roster contact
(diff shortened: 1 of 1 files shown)


jcbrand:
Thanks a lot for your pull requests.

I like the fact that you've come up with a simple solution to solve this particular issue. However there is a bigger architectural issue that I want to resolve which should then also fix this bug.

At the moment, as soon as a chat box is created, it's also opened and appears to the user. I want to change the code so that creating a chat box doesn't automatically open it as well. This should resolve certain issues with the API as well.

I'll first try that and see how it goes before deciding whether it then still makes sense to merge this particular pull request or not.

jcbrand:
OK, after doing some refactoring as explained above, I realize that your change is still perfectly valid and compatible.

floriancargoet:
Thank you for merging these PRs so quickly.

PR #358, opened 2015-04-01
Fix the ChatRoomOccupants storage id

floriancargoet:
The storage id generated for the occupants of a chat room is the same for all chat rooms. The results is that all chat rooms fetches the same data and all chat rooms appear to have the same occupants.

This is fixed by adding the room jid in the generated id.


jcbrand:
Thanks!

PR #369, opened 2015-04-14
Fixing when the path has space, it closes #368

### LINKED ISSUE #368: Error when the jsmin Gulp task is runned on Windows
When the path has space on it the taks is not executed properly. The solution is just to add the path into quotes.
http://www.evernote.com/shard/s100/sh/d0bf5894-baa3-44b0-b94a-bbc3fa54d458/51c69c93a686379c92606d5ac9ce5933

theoziran:


jcbrand:
Thanks!

PR #375, opened 2015-04-17
Fix collapsed group visibility on reload

### LINKED ISSUE #374: Collapsed groups disappear on page reload
If you collapse a group and reload the page, the group will not be displayed.
I am investigating.

floriancargoet:
Fixes #374 

In RosterGroupView.addContact()
``` js
                    if (this.model.get('state') === CLOSED) {
                        if (view.$el[0].style.display !== "none") { view.$el.hide(); }
                        if (this.$el[0].style.display === "none") { this.$el.show(); } // HERE
                    } else {
                        if (this.$el[0].style.display !== "block") { this.show(); }
                    }
```
The problem is that on initial page load, the `<dt>` are hidden by a CSS rule but not directly in the style element. Therefore, the `style.display === "none"` test fails but the element is hidden.

A better way of checking visibility is to use jQuery's is(':visible') method :

``` js
if (!this.$el.is(':visible')) { this.$el.show(); }
```

I'll let you decide if the other visibility tests should be updated.

### DIFF
--- a/converse.js
+++ b/converse.js
@@ -3978,7 +3978,7 @@
                 if (contact.showInRoster()) {
                     if (this.model.get('state') === CLOSED) {
                         if (view.$el[0].style.display !== "none") { view.$el.hide(); }
-                        if (this.$el[0].style.display === "none") { this.$el.show(); }
+                        if (!this.$el.is(':visible')) { this.$el.show(); }
                     } else {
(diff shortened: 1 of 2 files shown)


jcbrand:
Thanks. I think ``:visible`` is a better check in general.

PR #380, opened 2015-04-26
Note for debian & ubuntu users

pzia:
Quick note for packages required by Debian users.
Main point is `nodejs-legacy`.


jcbrand:
I've now updated the Makefile so it also installs bundler. This means that you don't need to install bundler via ``apt-get`` any more.

Would be nice if you could double check that it works and then remove ``bundler`` from the list of dependencies.

I've changed the Makefile now to also explicitly require the user to type ``make clean`` before running ``make dev`` if they want to start from scratch.

pzia:
Something is missing : 
```
~/Vcs/converse.js.git$ make build
make: *** No rule to make target 'stamp-node', needed by 'build'. Arrêt.
```

jcbrand:
Ah yes, that should be ``stamp-npm`` as you just figured out.

jcbrand:
Thanks!

PR #387, opened 2015-05-07
Fixed an error that appears when unload event is triggered

gbonvehi:
This happens when you load a page with converse and do not login to
the chat. It can be tested in converse.org preserving logs while
navigating to another page.
Inside clearSession there's a call to roster which is not initalized
and also another one to controlbox.save which is not set so it fails
with an url error during sync.

### DIFF
--- a/converse.js
+++ b/converse.js
@@ -569,10 +569,13 @@
         };
 
         this.clearSession = function () {
-            this.roster.browserStorage._clear();
+            if (this.roster) {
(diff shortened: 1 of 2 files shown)


gbonvehi:
I'm open to others solutions, this is a quick fix that I've tested that works.

jcbrand:
Thanks! Looks good. Just a tip, in future also include the pull request or issue number in the changelog.

PR #389, opened 2015-05-08
Added login-panel placeholder and roster-item Name translations

gbonvehi:
This will allow to translate 'user@server' and 'password' in the login
screen and also in the roster-item tooltip, there's 'Name:' which can
now be translated


jcbrand:
Thanks.

Issue #144, opened 2014-04-25
XEP-0199 Ping Pong 


manishjodhpur:
Hello Jcbrand,

There is a idle connection timeout set on my openfire of 30 seconds, so if user become idle, the connection disconnects after 30 second, this is due to reason that server sends a ping stanza and i should send a pong to server.
can you please help me to send a pong stanza as a reply of ping

thanks

thierrytiti:
I think this could be closed

jcbrand:
True, I thought it was already closed.

PR #390, opened 2015-05-10
Ping Pong

thierrytiti:
Fix to issue #144:
- Add of StropheJS ping plugin
- Add Ping functionality
- Add Pong handler


jcbrand:
Hi @thierrytiti, thanks for the pull request. 

I intend to merge your work, however I think we also need to [check whether the server supports XEP-0199](https://xmpp.org/extensions/xep-0199.html#support).

That would be a line like this:
``` javascript
if (converse.features.findWhere({'var': 'urn:xmpp:ping'}).length) {
    // The server supports XEP-0199 Ping
}
```

Here's an example where's it's being done for MUC:
https://github.com/jcbrand/converse.js/blob/a69ceba813c780943977df771329cb6687c99e3e/converse.js#L2016

converse.js
```diff
@@ -634,11 +635,35 @@
...
+        };
+
+        this.pong = function (ping){
+		    converse.connection.ping.pong(ping);
+	        return true;
+    	};
+
+	    this.registerPongHandler = function (){
+			converse.connection.ping.addPingHandler( this.pong );
+    	};
```
jcbrand: I think here we need to advertise that converse.js supports XEP-0199.

Something like this (I think in the `registerPongHandler` method):
```javascript
    converse.connection.disco.addFeature('urn:xmpp:ping');
```

converse.js
```diff
@@ -634,11 +635,35 @@
             },this), 200));
         };
 
+        this.ping = function (jid, timeout, success, error){
+            if (typeof jid === 'undefined'     || jid == null)     { jid = this.bare_jid; }
+                if (typeof timeout === 'undefined' || timeout == null) { timeout = 45; }
+                if (typeof success === 'undefined' ) { success = null; }
+                if (typeof error === 'undefined' ) { error = null; }
+                if (this.connection) this.connection.ping.ping( jid, success, error, timeout );
+        };
```
jcbrand: We should add add a callback method for `error` so that we can call `converse.reconnect` or `converse.renderLoginPanel`.

The code would be the same as here: https://github.com/jcbrand/converse.js/blob/a69ceba813c780943977df771329cb6687c99e3e/converse.js#L483

So I think we can refactor that out into a separate method and make it available as the error handler for `converse.ping`.

converse.js
```diff
@@ -634,11 +635,35 @@
...
+
+	    this.registerPongHandler = function (){
+			converse.connection.ping.addPingHandler( this.pong );
+    	};
+
+	    this.registerPingHandler = function (){
+            if (this.ping_interval>0){
+		        window.setInterval( function() { converse.ping(); }, this.ping_interval*1000);
+                }
+    	};
```
jcbrand: With this code the server will be pinged every `ping_interval` amount of seconds. Even when there is already lots of other stanzas being sent out between converse.js and the server.

An enhancement might be to rather wait `ping_interval` seconds from the last stanza sent to the server (no matter what stanza that was). This avoids pinging the server when it's not really necessary (i.e. when there's already lots of traffic between client and server).

I'm however not yet sure how we could go about doing that.

thierrytiti: If we change every converse.connection.send and converse.connection.sendIQ to internal functions converse.send and converse.sendIQ, it should be possible to keep a 'lastMessageTimeStamp' to compare to clock every X seconds.

There is also http://strophe.im/strophejs/doc/1.2.0/files/strophe-js.html#Strophe.Connection.addTimedHandler instead of window.setInterval.

converse.js
thierrytiti: Normaly , the connection would change and onConnStatusChanged should already be called.
my opinion is there is no need to force error callback to do this checks.
There is also servers that don't manage ping and will send errors but it will keep the connection.

jcbrand: Ok, I understand and agree.

converse.js
thierrytiti: With addTimedHandler, we need to re-register Ping on reconnection, because it's attached to connection object.

converse.js
thierrytiti: If i add a test to detect if the server supports XEP-0199 Ping, it's the same problem, it will not work if Ping is not supported, but it would keep the connection anyway.
Do you prefer a proper way (with the test) or a non standart way ( but  keeping connections) ?

converse.js
jcbrand: `addTimedHandler` doesn't seem to be doing what I have in mind though. I'm thinking of something that does an action only after there has been X amount of time where no stanzas were sent.

converse.js
jcbrand: Difficult to say :) 

I guess we don't do the check then.

converse.js
thierrytiti: A Strophe.TimedHandler encapsulates a user provided callback that should be called after a certain period of time or at regular intervals.
No reference to time between stanzas in the documentation (neither in the code)

The tests that i've done show it similar to setInterval but it's removed on disconnection.

thierrytiti:
I known my changes are very intrusive.
With a connection.addHandler it will be proper, please ignore my commit for the moment

thierrytiti:
This time it's better
Do you prefer I pull in only 1 commit ?

jcbrand:
> Do you prefer I pull in only 1 commit ?

Not necessarily, but I would appreciate more descriptive commit messages. `Corrections` doesn't really say what it is that you are doing :)

I don't have time to look into your changes right now, but I'll try to do so over the coming days.

alexdrupal:
Please, please, merge this. Really needed feature :+1:

thierrytiti:
Should I submit a pull request re-synchronized with last commit or there is  another reason ?

jcbrand:
@thierrytiti I would appreciate it if you synchronized with the last commit.

I would like to investigate whether it's possible to only send ping messages when the connection has been idle for `ping_interval` but I haven't had the time yet, which is why I haven't merged this yet.

thierrytiti:
@jcbrand It's already in the code ! 
I've managed it with keeping the date of the last sent message and sending only if ( last sent - now ) is greater than ping_interval 
the last commits make the code the less intrusive than possible 

I will close this pull request and make it synchronized, it will be clearer 

----- Mail original -----

De: "JC Brand" <notifications@github.com> 
À: "jcbrand/converse.js" <converse.js@noreply.github.com> 
Cc: "thierrytiti" <thierry.perpere@u-psud.fr> 
Envoyé: Mercredi 27 Mai 2015 17:36:06 
Objet: Re: [converse.js] Ping Pong (#390) 



@thierrytiti I would appreciate it if you synchronized with the last commit. 

I would like to investigate whether it's possible to only send ping messages when the connection has been idle for ping_interval but I haven't had the time yet, which is why I haven't merged this yet.

PR #393, opened 2015-05-11
Language Detection

thierrytiti:
Replace default language to autodetection if no i18n setting is set.

Add autodetection of the user language
default to locales.en if no matching found


jcbrand:
Cool feature, thanks!

PR #396, opened 2015-05-13
Auto away

thierrytiti:
Add functions for changing user presence to ``away`` or ``xa`` after timeouts

### DIFF
--- a/converse.js
+++ b/converse.js
@@ -354,17 +354,29 @@
         // Module-level functions
         // ----------------------
 		
-	this.autoAwayReset=function(){
+        this.autoAwayReset=function(){
             if (converse._idleCounter > 0) {
                 converse._idleCounter = 0;
                 if (converse._autoAway>0) {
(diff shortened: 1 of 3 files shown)


jcbrand:
Thanks. This is now merged.

I cleaned up the code a bit and also fixed some bugs, see https://github.com/jcbrand/converse.js/commit/47e00ae005fc4034836595d4267953b46a869cb7.

You might want to double-check that everything is still working properly for you.

jcbrand:
@thierrytiti I refactored the code to separate the CSI and auto away features. There is now a new config option ``csi_waiting_time`` specifically for CSI.

I also added some tests for both features.

The relevant commit is here:
https://github.com/jcbrand/converse.js/commit/868435173fceedef57fa14ad4ad19fd0daede16b

I did a quick live check and it looks to be working well. Would be good if you could see if it works fine for you as well.

PR #399, opened 2015-05-15
Bugfix. fullname is null

thierrytiti:
wrong callback argument mapping in XmppStatus initialize: fullname is
null


jcbrand:
Thanks!

PR #401, opened 2015-05-15
French Translation

thierrytiti:
Updated French Translation


jcbrand:
Thanks. BTW, for next time, you can generate the JSON file with the translations (which converse.js needs) by running `make po2json`.

PR #404, opened 2015-05-20
Css fix: position and width of #conversejs

thierrytiti:
- left and width: The last 35px at bottom of the page is not clickable
- right: in some case (e.g. frame in div), the div is over the vertical scrollbar

### DIFF
--- a/css/converse.css
+++ b/css/converse.css
@@ -32,14 +32,15 @@
   bottom: 0;
   direction: ltr;
   height: 35px;
-  left: 0;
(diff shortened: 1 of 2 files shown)


jcbrand:
Sorry @thierrytiti but the CSS is generated from Sass files. You need to make the changes there.

See here: https://github.com/jcbrand/converse.js/tree/master/sass

Then afterwards you can generate the CSS files by running `make css`.

PR #405, opened 2015-05-20
Css fixes

thierrytiti:
CSS rebuild with sass
main div conversejs
- left and width: The last 35px at bottom of the page is not clickable
- right: in some case (e.g. frame in div), the div is over the vertical
scrollbar

room-info:
- wrong placement on hover after information fetched on a room

### DIFF
--- a/css/converse.css
+++ b/css/converse.css
@@ -1,3 +1,4 @@
+@charset "UTF-8";
 /*!
  * Converse.js (Web-based XMPP instant messaging client)
  * http://conversejs.org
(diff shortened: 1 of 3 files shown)


css/converse.css
```diff
@@ -793,7 +796,7 @@
     #conversejs dd.available-chatroom:hover {
       background-color: #E3C9C1; }
       #conversejs dd.available-chatroom:hover .room-info {
-        display: inline-block;
+        display: block;
         font-size: 14px; }
```
jcbrand: Why did you make this change?

jcbrand:
Thanks!

PR #407, opened 2015-05-22
CSS: Fonts Path: editable $font-path via sass/variables.scss

thierrytiti:
Add a variable to build CSS with another path for the fonts
import added in  ``sass/converse.scss``: Undefined variable $font-path with ``make css``


jcbrand:
Thanks!

jcbrand:
k

PR #408, opened 2015-05-26
MUC: missing toggle Call handler and function

thierrytiti:
the Call button is visible but no action is done

### DIFF
--- a/converse.js
+++ b/converse.js
@@ -2429,14 +2429,6 @@
                 return this;
             },
 			
-            toggleCall: function (ev) {
-                ev.stopPropagation();
-                converse.emit('callButtonClicked', {
-                    connection: converse.connection,
-                    model: this.model
-                });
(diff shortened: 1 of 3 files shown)


converse.js
```diff
@@ -2427,7 +2428,15 @@
...
-
+			
+            toggleCall: function (ev) {
+                ev.stopPropagation();
+                converse.emit('callButtonClicked', {
+                    connection: converse.connection,
+                    model: this.model
+                });
+            },
+
```
jcbrand: I don't think this function is necessary because `ChatRoomView` extends `ChatBoxView` and therefore inherits the `toggleCall` method from there.

thierrytiti:
I've removed the unnecessary  toggleCall function

jcbrand:
Thanks! Merged.

PR #413, opened 2015-05-27
I18N: detect language for MomentJS

thierrytiti:
user language detection in order to format date and time in correct locale

### DIFF
--- a/converse.js
+++ b/converse.js
@@ -215,35 +215,52 @@
 
         // Translation machinery
         // ---------------------
-        this.isAvailableLocale = function (locale) {
+        //test for MomentJS
+        this.testMomentLocale = function(locale){
+            oldlang=moment.locale();
+            newlang=moment.locale(locale);
+            return (newlang != oldlang);
+        }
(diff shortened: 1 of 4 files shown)


thierrytiti:
Do you think it will be merged or there is something wrong ?

jcbrand:
I will eventually merge it, I just don't currently have time to thoroughly check it.

Non-trivial changes can sometimes take a surprising amount of time to properly integrate and test.

jcbrand:
I've merged this now, but also refactored the code quite a bit.

Can you please check whether it still works for you as intended?

Thanks.

thierrytiti:
There is a typo on the default return at the end of `` detectLocale `` function. 
This should be `` return locale || 'en';`` instead of `` return locale || locales.en;``

jcbrand:
Thanks, fixed here: https://github.com/jcbrand/converse.js/commit/0c67d2117bdc5ca662f79f6694381cb8bbcd24ab

PR #422, opened 2015-06-05
More small improvements to german locale

1st8:
My colleague @dino115 pointed these out.

Have a nice weekend :beers:


jcbrand:
Danke schön!

PR #426, opened 2015-06-16
AutoAway disabled while connection is lost

thierrytiti:
if connection is lost, the auto away mode must not change status.


jcbrand:
I only noticed this now after I refactored the code quite extensively.  :-/

jcbrand:
I've added a fix for this. See here: https://github.com/jcbrand/converse.js/commit/21a3b4405080a3f20470e2073dbe6a8587e892ac

thierrytiti:
Thanks

PR #434, opened 2015-07-01
Cleanup references to strophe.roster.js

thierrytiti:
Cleanup references to strophe.roster.js in non_amd.html and main.js


jcbrand:
Nice, thanks!

Issue #306, opened 2015-01-19
XEP-0313: Message Archive Management


phragment:
[![Bountysource](https://www.bountysource.com/badge/issue?issue_id=7883988)](https://www.bountysource.com/issues/7883988-xep-0313-message-archive-management?utm_source=7883988&utm_medium=shield&utm_campaign=ISSUE_BADGE)

It would be great to have support for [MAM](http://xmpp.org/extensions/xep-0313.html).
This would allow users to seamlessly continue their chats switching between converse.js and other XMPP clients, since messages received on another client while converse.js is offline could be retrieved from the server.

For example, this does work using prosody, gajim (Linux/Windows) and conversations (Android).

jacekgrzybowski:
+1

fpytloun:
+1 ;-)

devendiran:
+1

nickbroon:
:+1:

nickbroon:
ejabbard and prosody both now have support for MAM http://modules.prosody.im/mod_mam.html and it being used by kaiwa https://github.com/digicoop/kaiwa#message-history-syncing

jacekgrzybowski:
@jcbrand do you plan to implement this feature?

jcbrand:
@jacekgrzybowski I prioritize what people pay me to do or what I'd like to work on for myself. I don't have a particular need for this feature so it's not high on my list of priorities.

fpytloun:
And do you have some button like "Donate for XEP-0313"? :-)
If so, I think some people would support your project, especially this feature, expected by every modern-day chat.

jcbrand:
@fpytloun You could post a bounty on [bountysource](https://www.bountysource.com). :)

fpytloun:
Here it goes.. :-)

jcbrand:
Thanks @fpytloun :)

styoung:
@jcbrand please let me know how big this bounty would have to be in order to prioritize?

jtrachtenberg:
@jcbrand - another request - please let @styoung and I know

jcbrand:
@jtrachtenberg I appreciate it if people want to make smaller contributions towards getting something done, and I understand that not everyone can afford to pay a developer to implement a full feature. However, if you want to know what my quote is for a feature, then you can contact me personally, and not via the github issue tracker.  https://opkode.com/contact.html

jcbrand:
This has feature now been implemented and merged into master.

PR #442, opened 2015-07-23
Message Archive Management

jcbrand:
This pull request implements support for MAM, as asked for in #306.

The following new configuration options have been added:

- [muc_history_max_stanzas](https://github.com/jcbrand/converse.js/blob/mam/docs/source/configuration.rst#muc-history-max-stanzas)
- [archived_messages_page_size](                                                                                                                                                                                  
https://github.com/jcbrand/converse.js/blob/mam/docs/source/configuration.rst#archived-messages-page-size)
- [message_archiving](https://github.com/jcbrand/converse.js/blob/mam/docs/source/configuration.rst#message-archiving)                                                                                                                                                                                                                  

The **message_archiving** option allows you to specify your archiving preferences, but these preferences may or may not be ignored by the XMPP server, depending on how it's configured.
If the server is configured to always ignore what the client says, then that option makes no difference.                                                                                                                                                                                                    
Otherwise, if the server by default doesn't log and listens to what the client prefers, then it is necessary.

Besides showing archived messages, there is now also an API for archived messages:
https://github.com/jcbrand/converse.js/blob/mam/docs/source/development.rst#the-archive-grouping

CC: @fpytloun


fpytloun:
I have deployed it and it seems to behave just as I have expected.
It's great, thank you!

jcbrand:
Good to hear! Thanks.

PR #444, opened 2015-07-31
Excessive double quote in french translation file

skope:
Excessive double quote in ./locale/fr/LC_MESSAGES/converse.po caused "make po" to fail:

./locale/fr/LC_MESSAGES/converse.po:286: end-of-line within string


jcbrand:
Thanks!

PR #458, opened 2015-09-01
Fix no messages being shown in chat 

1st8:
Hey,

when upgrading my custom built version 0.9.3 to the current 0.9.5 (in order to use the new message archiving feature) converse stopped displaying any (sent and received) messages in the chat views.
When debugging the code, I found `first_msg_date` to be `null` and therefore the code trying to determine the insert position did nothing.

The situation is fixed in my PR by handling both `undefined` and `null` correctly.
I don't know why I get `null` from `$first_msg.data('isodate')` when you were getting `undefined`, maybe it is related to me receiving archived messages or using a slightly different jQuery version (1.10.2).

Cheers!


jcbrand:
Thanks!

PR #462, opened 2015-09-03
Fix #461: MUC working with spaces etc. in room names

### LINKED ISSUE #461: MUC room names with spaces / multiple escaping
Hi,

I'm trying to resolve an issue that prevents rooms with names containing spaces (and probably other characters) from working.
Is this a known issue?

It seems, that the room jid is escaped more than once, at least when fetching the room list. ("foo bar@rooms.server.local" is escaped to "foo\20bar..." and then again to "foo\5c20bar...")
Upon room creation it is escaped correctly once.
Any ideas/hints? I'll let you know if I was able to make it work.

Cheers!

1st8:
Rooms with bullshit in name now working, yay!
Next: Submit `muc#roomconfig_roomname` on creation to persist full name on server.

bildschirmfoto 2015-09-03 um 16 20 23

### DIFF
--- a/converse.js
+++ b/converse.js
@@ -2319,15 +2319,16 @@
                 else { $nick.removeClass('error'); }
 
                 if (ev.type === 'click') {
+                    name = $(ev.target).text();
                     jid = $(ev.target).attr('data-room-jid');
                 } else {
                     $name = this.$el.find('input.new-chatroom-name');
                     $server= this.$el.find('input.new-chatroom-server');
                     server = $server.val();
-                    name = $name.val().trim().toLowerCase();
+                    name = $name.val().trim();
                     $name.val(''); // Clear the input
                     if (name && server) {
-                        jid = Strophe.escapeNode(name) + '@' + server;
+                        jid = Strophe.escapeNode(name.toLowerCase()) + '@' + server;
                         $name.removeClass('error');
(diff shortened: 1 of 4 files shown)


converse.js
```diff
@@ -3009,7 +3010,7 @@
             getRoomJIDAndNick: function (nick) {
                 nick = nick || this.model.get('nick');
                 var room = this.model.get('jid');
-                var node = Strophe.escapeNode(Strophe.getNodeFromJid(room));
+                var node = Strophe.getNodeFromJid(room);
```
jcbrand: Shouldn't the node still be escaped here? Otherwise looking good.

jcbrand:
Thanks. Please also update docs/CHANGES.rst

converse.js
1st8: That was the main problem, it caused the node to be escaped **twice**.
I don't fully understand how it was intended to be, but this fixes the problem and my testing doesn't indicate any regressions.

1st8:
CHANGES.rst updated!

converse.js
jcbrand: I think it gets escape twice because when the room is created the JID is escaped the first time.

jcbrand:
Thanks!

PR #464, opened 2015-09-04
Fix translation error in en locale

hjbolide:
Fix incorrect string mapping in en locale


jcbrand:
Please do me a favour and make a new pull request which doesn't include a changed `converse.min.js`
Whenever converse.min.js changes, the diff file is very large which unnecessarily bloats the repository.
Thanks!

hjbolide:
Sure JC, I'll be doing that as soon as I finish my morning prep.

Regards,
Charles

hjbolide:
Hi JC,

I have reverted converse.min.js on top of my current pull request, let me know if that does it for you.

Regards,
Charles

jcbrand:
Unfortunately not because the converse.min.js diff is still in the history, which is the issue.

PR #468, opened 2015-09-09
Fix [object Object] being sometimes shown as status

1st8:
Fix [object Object] being shown as status when presence stanza with custom 'status' is received from other resource.

Cheers!


jcbrand:
Nice one, thanks!

1st8:
You're welcome!

PR #500, opened 2015-10-12
Fix #467

### LINKED ISSUE #467: Messages both in local and servers archive appear twice
Hi,

when using the new MAM feature, messages stored both in local and server archive appear twice.
I remember vaguely that this is a known issue, but I couldn't find one here.

Do you have a possible solution in mind that I could implement and submit as PR?
I would otherwise try to merge the messages from the two archives by ID, if thats possible, will take a look later.
Or turning off the local archive if MAM is active would be easy too, wouldn't it?

Cheers!

1st8:
Internal usage shows no regressions.

Cheers

### DIFF
--- a/converse.js
+++ b/converse.js
@@ -932,7 +932,14 @@
             }
         });
 
-        this.Message = Backbone.Model;
+        this.Message = Backbone.Model.extend({
+            idAttribute: 'msgid',
+            defaults: function(){
+                return {
+                    msgid: converse.connection.getUniqueId()
+                };
+            }
+        });
         this.Messages = Backbone.Collection.extend({
             model: converse.Message,
             comparator: 'time'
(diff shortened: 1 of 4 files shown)


jcbrand:
Thanks. Please fix the JSHint error: https://travis-ci.org/jcbrand/converse.js/builds/84879419

1st8:
Hmm I will fix the other issues too and squash everything...
Do you have any idea why running `make check` picks up a german locale and thus fails all "Click on some english label" specs?

1st8:
Running `LANG=en_EN make check` works now... I will add this to the README

1st8:
Should be good now.

jcbrand:
@1st8 Really good work done with this, thanks.

1st8:
Thanks, you're welcome.

PR #501, opened 2015-10-12
Update test instructions

1st8:


jcbrand:
Thanks, but there's a better solution. I pushed a fix that ensures that tests always run in English.

PR #503, opened 2015-10-19
Fix roster push when new contact is pushed from server and emit rosterPush event

teseo:
Fix for the onRosterPush method and emitting a rosterPush event in to allow developers custom their applications with the new rosterPush


converse.js
```diff
@@ -4375,7 +4375,10 @@
                 $(iq).children('query').find('item').each(function (idx, item) {
                     this.updateContact(item);
                 }.bind(this));
-                return true;
+
+                converse.emit('rosterPush', iq);
+
```
jcbrand: Please document this emitted event, here: https://github.com/jcbrand/converse.js/blob/master/docs/source/development.rst#event-types

converse.js
```diff
@@ -4362,7 +4362,7 @@
                  */
                 var id = iq.getAttribute('id');
                 var from = iq.getAttribute('from');
-                if (from && from !== "" && Strophe.getNodeFromJid(from) != converse.bare_jid) {
+                if (from && from !== "" && from != converse.connection.jid) {
                     // Receiving client MUST ignore stanza unless it has no from or from = user's bare JID.
```
jcbrand: Please use "!==" instead of "!=".

teseo:
Thank you for the advise Jc. There you have the requested changes. Hope all is good.

Pleasure collaborating on converse.js

jcbrand:
Thanks @teseo!

PR #505, opened 2015-10-22
Typo caused [object Object] in room info

gromiak:
Instead of __( it was _( caused [object Object] in more room info.


jcbrand:
Thanks!

PR #509, opened 2015-10-28
Update of Polish translation to version 0.9.6.

ser:
Polish translation updated. Thanks again for your fantastic software :+1:


jcbrand:
Thanks!

Issue #524, opened 2015-12-01
auto_subscribe and chatroom invites


andreabenini:
Hi JC and Team, I'm using converse.js and I really love it.
I'm now using auto_subscribe option to get subscriptions automatically accepted when needed, it works and seems to do its job.
I'd like to have a similar feature for having the client automatically joining chat rooms because I'd like to use them in an external web portal and I don't like to get warned from an alert();
I have alread found the solution and it's pretty easy and simple, current code is:
```javascript
onInvite: function (message) {
...
if (!reason) {
    result = confirm(
                        __(___("%1$s has invited you to join a chat room: %2$s"), contact.get('fullname'), room_jid)
                    );
} else {
    result = confirm(
                         __(___('%1$s has invited you to join a chat room: %2$s, and left the following reason: "%3$s"'),
                                contact.get('fullname'), room_jid, reason)
                    );
}
```

My suggested mod is:
```javascript
onInvite: function (message) {
...
if (converse.auto_subscribe) {
    result = true;
} else {
    if (!reason) {
        result = confirm(
                            __(___("%1$s has invited you to join a chat room: %2$s"), contact.get('fullname'), room_jid)
                        );
    } else {
        result = confirm(
                             __(___('%1$s has invited you to join a chat room: %2$s, and left the following reason: "%3$s"'),
                                    contact.get('fullname'), room_jid, reason)
                        );
    }
}
```

It could be a slick 'n' easy mod, may I submit a pull request for it ? I can use a different configuration variable if you don't think auto_subscribe is your preferred one. What do you think about it ?

Ben

jcbrand:
Hi Ben, a pull request would be great thanks. Let's use a new config variable for this, I'm thinking `auto_join_on_invite`.

In your pull request, please make sure to update the [configuration](https://github.com/jcbrand/converse.js/blob/master/docs/source/configuration.rst) documentation and also the [changelog](https://github.com/jcbrand/converse.js/blob/master/docs/CHANGES.md).

andreabenini:
Just added the proposed pull, there's even a minor fix when you receive a chatroom invite from someone not in your roster list, just a one line fix

PR #526, opened 2015-12-02
Auto join on invite patch 1 (auto_join_on_invite new parameter)

andreabenini:
As planned on issue #524 I have added this new parameter.
Minor fix: fixed confirm() dialog box when a request comes from someone not in your roster list


jcbrand:
Thanks!

PR #528, opened 2015-12-04
no-otr compilation issue: Utils is not defined

JoseMariaRubioMoral:
I had problems in the no-otr compilation.  I had to change the order of the parameters in this file.


jcbrand:
Thanks!

PR #531, opened 2015-12-10
Update for more best compatibility with wordpress

camaran:


camaran:
i close, i manage it in wordpress specific plugin

PR #532, opened 2015-12-21
Add a main section to bower.json

dunglas:
Allow to use plugins such as Grunt and Gulp WireDep.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -9,6 +9,7 @@
     "requirejs-text": "~2.0.10",
     "requirejs-tpl-jcbrand": "*"
   },
+  "main": "./converse.js",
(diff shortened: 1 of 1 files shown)


dunglas:
Closing. Doesn't work well.

jcbrand:
I think `main` should be set to `main.js`, no?

dunglas:
Yes, in fact it should be an array containing all required files in the good order.

jcbrand:
The file [non_amd.html](https://github.com/jcbrand/converse.js/blob/master/non_amd.html) loads all the required files and in the correct order. Would you be willing to check whether that works?

PR #534, opened 2015-12-24
Translation for Russian language updated.

LaconicTranslator:
Still undone. Ready to continue.


jcbrand:
Thanks! Should I merge so long?

LaconicTranslator:
Yes, the proposed translation is definitely more up to date than the previous one.

PR #554, opened 2016-01-18
Fix #487

### LINKED ISSUE #487: Offline status does not update Roster
Steps to reproduce:
Assume that there are two users (user1 & user2) that are already subscribed to each other.

1.) Change the XMPP status of user2 to Offline
2.) The contact in user1's roster doesn't update and remains to be in the last known XMPP state (anything other than offline)

RosterContacts::presenceHandler has a check, 
...
else if (presence_type === 'unavailable' && contact) {
                    // Only set the user to offline if there aren't any
                    // other resources still available.
                    if (contact.removeResource(resource) === 0) {
                        contact.save({'chat_status': "offline"});
                    }
                }

The contact here has some resources (sometimes its null and contact default resource or something else).
Does bypassing the check for resources and changing the status to offline create some problem?

rlanvin:
Emptying the resources array when the resource is null

### DIFF
--- a/converse.js
+++ b/converse.js
@@ -4192,6 +4192,13 @@
                         this.save({'resources': resources});
                     }
                 }
+                else {
+                    // if there is no resource (resource is null), it probably
+                    // means that the user is now completely offline. To make sure
+                    // that there isn't any "ghost" resources left, we empty the array
+                    this.save({'resources': []});
+                    return 0;
+                }
                 return resources.length;
             },
 
--- a/docs/CHANGES.md
+++ b/docs/CHANGES.md
@@ -15,6 +15,7 @@
 - #547 By default the `offline` state is no longer choosable.
   See [include_offline_state](https://conversejs.org/docs/html/configuration.html#include_offline_state) for details. [jcbrand]
 - A chatroom invite might come from someone not in your roster list. [ben]
+- #487 Empty the resources array when the resource is null [rlanvin]
 
 ## 0.10.0 (2015-11-05)
 


converse.js
```diff
@@ -4192,6 +4192,12 @@
                         this.save({'resources': resources});
                     }
                 }
+                else {
+                    // if there is no resource (resource is null), it probably
+                    // means that the user is now completely offline. To make sure
+                    // that there isn't any "ghost" resources left, we empty the array
+                    resources.length = 0;
+                }
```
jcbrand: The change should rather be: 

```javascript
this.save({'resources': []});
```

jcbrand:
Please make the change as mentioned in the line note. Then also add an entry to the changelog: [docs/CHANGES.md](https://github.com/jcbrand/converse.js/blob/master/docs/CHANGES.md).

Then I'll merge the pull request.

rlanvin:
I need to keep the line `resources.length = 0`, otherwise the return statement  `return resources.length;` won't return zero. I added your line, the changes to CHANGES.md and squashed the commits. Hope everything is ok, I'm not really a JS developer.

jcbrand:
I don't think `resources.length = 0` is correct, because you're then overriding a computed property. `length` should be returning the length of an array, so its value changes as the array changes and should therefore not be overridden.

Instead, please just do a `return 0;` there.

rlanvin:
Alright will do, but just FYI I found the syntax here http://stackoverflow.com/questions/1232040/how-do-i-empty-an-array-in-javascript

> `A.length = 0`
> This will clear the existing array by setting its length to 0. Some have argued that this may not work in all implementations of JavaScript, but it turns out that this is not the case. It also works when using "strict mode" in ECMAScript 5 because the length property of an array is a read/write property.

irek-zaripov:
I'm from #555
I have tested by
$ git clone https://github.com/rlanvin/converse.js
The bug is still present in this pull request. I think it depends from the sessionStorage restore.
As the user is `offline` in XMPP server, the status stanza will never sent to the converse.js's roster...

rlanvin:
@irek-zaripov  Hm, well my fix happens when the XMPP server sends a new presence "unavailable" upon reconnection - which is what happen to me in your example when user A log in to Converse.js again. So are you getting a status update from the server when you reconnect? The core of the issue is that the `ressources` still contains cached resources. A cleaner fix would be to empty it when a disconnection occurs, but I don't know enough about the code to know where this is.

irek-zaripov:
`So are you getting a status update from the server when you reconnect? ` Yes I'v get.

irek-zaripov:
OK @rlanvin May be the issues #555 #487 have different nature

rlanvin:
@irek-zaripov ok just to make sure, how do you test my pull request? Are you making a new build? Because `converse.min.js` is not updated by the pull request, so you need to make new build manually. If that's what you are doing, could you please copy the presence that you are receiving from the server, or check that is corresponds to my description of the bug in #487.

irek-zaripov:
I'v made:
$  git clone https://github.com/rlanvin/converse.js
$  cd converse.js
$  make dev
$  make release

irek-zaripov:
@rlanvin By this fix, the status of user B is not get `offline` even when the user A(converse.js) is not logged off.

jcbrand:
> As the user is offline in XMPP server, the status stanza will never sent to the converse.js's roster...

I'm removing the `offline` status, partly due to the reason that `offline` users don't receive any presence stanzas anymore. So in the next release the above shouldn't be a problem.

@rlanvin I didn't know about `A.length = 0;` clearing an array. I'd prefer to just `return 0;` however, it's less "clever" and more explicit.

> A cleaner fix would be to empty it when a disconnection occurs

I think you have a point, but I'm not sure myself where to do that, I'll have to investigate a bit first. In any case, I think this change is still valid, even if we implemented that.

jcbrand:
@irek-zaripov If you are sure this pull request doesn't work in your case, the best thing you can do is to step through code in the JS debugger in your browser and see what happens when the code in this pull request executes. Then you can report back on it.

irek-zaripov:
@jcbrand , @rlanvin I set the status of user B to `dnd` and close the Pidgin application( kill -9 'Pidgin PID'). In the XMPP server(openfire)
I see that the session of user B is timed out & closed after some times.
But the status in converse.js(user A) is still `dnd`, and stands `dnd` forever till user B log in again and change the status.

rlanvin:
@jcbrand done

@irek-zaripov I tried what you described, but I cannot reproduce the bug (with Prosody XMPP server). When User B session times out, Converse.js receives the following statement, and User B is correctly shown "unavailable" (aka "offline"):

```
<presence xmlns="jabber:client" type="unavailable" to="userA@test.dev" from="userB@test.dev/757f939b-3e68-45a8-8ee1-354be4ff71df">
    <status>Disconnected: closed</status>
</presence>
```

Additionally, since in this case the resource ("757f939b-3e68-45a8-8ee1-354be4ff71df", i.e. Pidgin) is sent in the XML, this is not impacted by or related to my fix.

irek-zaripov:
OK, thanks @rlanvin, great work!
@jcbrand but the issue  #555 is not resolved yet. I'll continue to investigate in the spare time...

PR #558, opened 2016-01-22
Devel

m0cs:
Changed condition to reconnect allowing auto reconnect if option auto_reconnect is setting up.
A bug on ipad cannot find functions endsWith on min_width and min_height.
Setting up a cleaner timeout to reconnect.

### DIFF
--- a/converse.js
+++ b/converse.js
@@ -1378,8 +1378,8 @@
                 }
                 var min_width = $flyout.css('min-width');
                 var min_height = $flyout.css('min-height');
-                this.model.set('min_width', min_width.endsWith('px') ? Number(min_width.replace(/px$/, '')) :0);
-                this.model.set('min_height', min_height.endsWith('px') ? Number(min_height.replace(/px$/, '')) :0);
+                this.model.set('min_width', min_width ? Number(min_width.replace(/px$/, '')) : 0);
+                this.model.set('min_height', min_height ? Number(min_height.replace(/px$/, '')) : 0);
                 // Initialize last known mouse position
                 this.prev_pageY = 0;
                 this.prev_pageX = 0;
--- a/converse.js
+++ b/converse.js
@@ -632,7 +632,7 @@
                     converse.onConnected();
                 }
             } else if (status === Strophe.Status.DISCONNECTED) {
-                if (converse.disconnection_cause === Strophe.Status.CONNFAIL && converse.auto_reconnect) {
+                if (converse.auto_reconnect) {
                     converse.reconnect(condition);
                 } else {
                     converse.renderLoginPanel();
--- a/converse.js
+++ b/converse.js
@@ -417,6 +417,9 @@
         HEADER_WEIGHTS[HEADER_REQUESTING_CONTACTS] = 2;
         HEADER_WEIGHTS[HEADER_PENDING_CONTACTS]    = 3;
 
+        //auto_reconnect timeout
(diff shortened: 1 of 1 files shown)


converse.js
```diff
@@ -593,8 +596,10 @@
         this.reconnect = function (condition) {
             converse.log('Attempting to reconnect in 5 seconds');
             converse.giveFeedback(__('Attempting to reconnect in 5 seconds'), 'error');
-            setTimeout(function () {
+            clearTimeout(converse.reconnectTimeout);
+            converse.reconnectTimeout = setTimeout(function () {
                 if (converse.authentication !== "prebind") {
+                    this.connection.pass = this.password;
                     this.connection.connect(
```
jcbrand: This looks wrong. There's no guarantee that `this.password` has a value, it's an optional config setting.

converse.js
```diff
@@ -417,6 +417,9 @@
         HEADER_WEIGHTS[HEADER_REQUESTING_CONTACTS] = 2;
         HEADER_WEIGHTS[HEADER_PENDING_CONTACTS]    = 3;
 
+        //auto_reconnect timeout
+        this.reconnectTimeout = undefined;
+
```
jcbrand: This needs to go below the `// Module-level variables` header.

converse.js
```diff
@@ -632,7 +641,7 @@
                     converse.onConnected();
                 }
             } else if (status === Strophe.Status.DISCONNECTED) {
-                if (converse.disconnection_cause === Strophe.Status.CONNFAIL && converse.auto_reconnect) {
+                if (converse.auto_reconnect) {
```
jcbrand: Can you please explain why you removed the extra check?

converse.js
```diff
@@ -1378,8 +1387,8 @@
                 }
                 var min_width = $flyout.css('min-width');
                 var min_height = $flyout.css('min-height');
-                this.model.set('min_width', min_width.endsWith('px') ? Number(min_width.replace(/px$/, '')) :0);
-                this.model.set('min_height', min_height.endsWith('px') ? Number(min_height.replace(/px$/, '')) :0);
+                this.model.set('min_width', min_width ? Number(min_width.replace(/px$/, '')) : 0);
+                this.model.set('min_height', min_height ? Number(min_height.replace(/px$/, '')) : 0);
                 // Initialize last known mouse position
```
jcbrand: I don't think simply removing `endsWidth` is appropriate. If `endsWith` is not available on iPad, then it needs to be polyfilled in utils.js (there are already other polyfills there).

jcbrand:
Thanks for the pull request.

Since you are actually making three distinct changes, you should ideally have made three pull different pull requests.

In any case, please see my inline comments.

m0cs:
Ok ill do that! Thanks!

converse.js
m0cs: Yeap, sorry for that! When using a token as a password, maybe tokens are valid one time, if reconnects an external funcion update token and set to this.password variable, my mistake was not think about it can be without a pass.

converse.js
m0cs: Done!

converse.js
m0cs: Because some times disconnection_cause was undefined and not reconnected.

converse.js
m0cs: Yea! I'll take a look at that.

PR #582, opened 2016-02-10
Bumped version to fix mismatch

thesabbir:
Fixed:
```
bower converse#*              mismatch Version declared in the json (0.10.0) is different than the resolved one (0.10.1)
```


jcbrand:
Thanks, but I don't know if that will be enough, I might have to make a new release.

PR #587, opened 2016-02-12
logout if you are using auto_login=true

davec82:
If you use auto_login=true, when you logout, the main panel hang on loading.

### DIFF
--- a/converse.js
+++ b/converse.js
@@ -753,6 +753,7 @@
         };
 
         this.logOut = function () {
+            converse.auto_login = false;
             converse.chatboxviews.closeAllChatBoxes(false);
             converse.clearSession();
             converse.connection.disconnect();
--- a/locale/it/LC_MESSAGES/converse.json
+++ b/locale/it/LC_MESSAGES/converse.json
@@ -297,7 +297,7 @@
          ],
          "Log out": [
             null,
-            "Logo out"
+            "Log out"
          ],
          "Contact name": [
             null,


converse.js
```diff
@@ -2617,6 +2617,11 @@
                 var $feedback = this.$('.conn-feedback'); // we want to still show any existing feedback.
                 this.$el.html(converse.templates.controlbox(this.model.toJSON()));
                 var cfg = {'$parent': this.$el.find('.controlbox-panes'), 'model': this};
+
+                if (!this.model.get('connected') && converse.auto_login){
+                    converse.auto_login = false;
+                }
+
```
jcbrand: Thanks, but I'm not sure I like this check here. I think a better place would be the [logOut](https://github.com/jcbrand/converse.js/blob/875a05e0a5cceafce7befdaa64eacc357efa44dc/converse.js#L756) method. Then it's more clear that the `auto_login` gets set to `false` because the user logged out.

jcbrand:
Thanks!

mhchu:
Hi, I'm not sure if I'm supposed to comment on this closed issue or create a new one, but I think the fix that was merged on Feb 16 is missing in the current master branch. The current converse.logOut function is defined as:
```javascript
        this.logOut = function () {
            converse.chatboxviews.closeAllChatBoxes();
            converse.clearSession();
            if (typeof converse.connection !== 'undefined') {
                converse.connection.disconnect();
                converse.connection.reset();
            }
        };
```
The fixed version has this line inserted at the top of the function:
```javascript
converse.auto_login = false;
```

I can confirm in our implementation of today's master branch (April 13), that when auto_login = true and the user logs out, the main panel shows a spinning wait icon.

davec82:
that line was removed in this commit a969d010cbae36f3ecf55f787f8bff3751db8f3f (13-04).
I don't know if it is related to the problem "Don't try to reconnect on auth fail." or it's a typo. Probably, @jcbrand can help us.

mhchu:
I decided to try to patch it (for now) with my first attempt at a plugin, but unfortunately, the plug in override didn't run, so I must be doing something wrong.
```javascript
(function (root, factory) {
    if (typeof define === 'function' && define.amd) {
    	// AMD. Register as a module called "myplugin"
    	define("myplugin", ["converse"], factory);
    } else {
    	// Browser globals. If you're not using a module loader such as require.js,
    	// then this line below executes. Make sure that your plugin's <script> tag
    	// appears after the one from converse.js.
    	factory(converse);
    }
}(this, function (converse_api) {
		
    // Commonly used utilities and variables can be found under the "env"
    // namespace of converse_api
    
    // Strophe methods for building stanzas
    var Strophe = converse_api.env.Strophe,
    $iq = converse_api.env.$iq,
    $msg = converse_api.env.$msg,
    $pres = converse_api.env.$pres,
    $build = converse_api.env.$build,
    b64_sha1 = converse_api.env.b64_sha1;
    
    // Other frequently used utilities
    var $ = converse_api.env.jQuery,
    _ = converse_api.env._,
    moment = converse_api.env.moment;
    
    // The following line registers your plugin.
    converse_api.plugins.add('myplugin', {
        overrides: {
        	logOut: function() {
        		console.log('override logOut called');
        		var converse = this._super.converse;
        		converse.auto_login = false;
        		this._super.logout();
        		converse.emit('statusChanged', 'logout');
        	}
        	
        }
    });
    return converse_api;
}));
```
:(

mhchu:
I managed to figure out how to load the plugin. Not sure why, but the way the example was written didn't load the plugin but I rewrote it with require and circumvented the logout hanging bug by inserting the converse.auto_login = false right before the logOut().
```javascript
require(['converse'], function (converse_api) {
		
    // Commonly used utilities and variables can be found under the "env"
    // namespace of converse_api
    
    // Strophe methods for building stanzas
    var Strophe = converse_api.env.Strophe,
    $iq = converse_api.env.$iq,
    $msg = converse_api.env.$msg,
    $pres = converse_api.env.$pres,
    $build = converse_api.env.$build,
    b64_sha1 = converse_api.env.b64_sha1;
    
    // Other frequently used utilities
    var $ = converse_api.env.jQuery,
    _ = converse_api.env._,
    moment = converse_api.env.moment;
    
    // The following line registers your plugin.
    converse_api.plugins.add('myplugin', {
        overrides: {
        	logOut: function() {
        		console.log('plugin logOut called');
        		var converse = this._super.converse;
        		converse.auto_login = false;
        		this._super.logOut();
        		converse.emit('statusChanged', 'logout');
        	}
        	
        }
    });
});
```

Issue #566, opened 2016-02-03
How can I disable focusing latest opening chat box 


mustafaevcil:
My problem is that converse focus on latest opening chat box (when user gets new message). If the user get new message from other roster when is typing with rosters, chat box opens and focus. Therefore the user send wrong messages instead of active typing roster. I think default behaviour may be for new chat box open the left and send notification for this case or chat box minimized automatically.

jcbrand:
Fixed in master.

PR #590, opened 2016-02-14
Don't steal focus when chatbox is opened automatically

rlanvin:
This is an attempt (probably a bit ugly) to fix the problem described in #566 - when somebody talks to you, the opening chatbox steals the focus, which is extremely annoying.

Here is a quick explanation:

I noticed that the focus is happening in `ChatBoxView.show`, so I disabled it from here. However it's good to focus when the user open the chatbox manually. I noticed that when the chatbox is opened by a click from the user, the execution flows first through `ChatBoxViews.showChat`. Unfortunately from here I didn't find anyway to call `focus`... so instead I made this method trigger the `show` event with a parameter `focus` set to `true`. I'm not sure what the `callback` parameter was used for, so I left it intact. There is probably a cleaner way to pass the parameter to the event.

### DIFF
--- a/converse.js
+++ b/converse.js
@@ -2174,22 +2174,22 @@
                 return this;
             },
 
-            show: function (callback) {
+            show: function (focus) {
                 if (this.$el.is(':visible') && this.$el.css('opacity') === "1") {
                     return this.focus();
                 }
                 this.initDragResize().setDimensions();
                 this.$el.fadeIn(function () {
-                    if (typeof callback === "function") {
-                        callback.apply(this, arguments);
-                    }
                     if (converse.connection.connected) {
                         // Without a connection, we haven't yet initialized
                         // localstorage
                         this.model.save();
                     }
                     this.setChatState(ACTIVE);
-                    this.scrollDown().focus();
+                    this.scrollDown();
+                    if ( focus ) {
+                        this.focus();
+                    }
                 }.bind(this));
                 return this;
             },
@@ -3923,7 +3923,7 @@
                 if (chatbox.get('minimized')) {
                     chatbox.maximize();
                 } else {
-                    chatbox.trigger('show');
+                    chatbox.trigger('show', true);
                 }
                 return chatbox;
             }
--- a/docs/CHANGES.md
+++ b/docs/CHANGES.md
@@ -5,6 +5,7 @@
 - #261 show_controlbox_by_default config not working [diditopher]
 - #573 xgettext build error: `'javascript' unknown`
 - Save scroll position on minimize and restore it on maximize [rlanvin]
+- #566 Do not steal the focus when the chatbox opens automatically [rlanvin]
 
 ## 0.10.1 (2016-02-06)
 


converse.js
```diff
@@ -2186,7 +2186,9 @@
                         this.model.save();
                     }
                     this.setChatState(ACTIVE);
-                    this.scrollDown().focus();
+                    if (typeof callback === "object" && callback.focus === true) {
+                        this.scrollDown().focus();
+                    }
                 }.bind(this));
```
jcbrand: You've moved both `scrollDown` and `focus` in to the if statement. We still want to scroll down, right? So above the `if` should be `this.scrollDown();` and inside should be `this.focus();`.

converse.js
```diff
@@ -3920,7 +3922,7 @@
                 if (chatbox.get('minimized')) {
                     chatbox.maximize();
                 } else {
-                    chatbox.trigger('show');
+                    chatbox.trigger('show', {focus: true});
                 }
```
jcbrand: I'm unsure what I think of passing in an object like that. The only alternative I can think of right now, is to trigger a slightly different event, like `showAndFocus`. Not yet sure which is better.

rlanvin:
Ok thanks for the code review. I have two questions:

* when / what for is the `callback` argument used?
* why is this `show` method called so many times? (it's called for the chatbox, and also for every message in it)

rlanvin:
@jcbrand I pushed another version of the fix, let me know what  you think.

jcbrand:
>    when / what for is the callback argument used?

I added the code for it here: https://github.com/jcbrand/converse.js/commit/d2358a6a93a99150628908af41bf9405d903f76c

Doesn't look like it's being used for anything :-P

I think I added it because the show method for the toggle accepts a callback argument (and in that case it is used to call "show" on the control box after the toggle has been hidden).

So long story short, looks like it could be removed.

>  why is this show method called so many times? (it's called for the chatbox, and also for every message in it)

It doesn't get called for every message received, but it gets called for each old message restored from sessionStorage.

That was actually a bug, I was using debounce improperly. I fixed that now, but I did it in a branch which I'm going to merge soon, rather than master because that'll cause merging headaches.
https://github.com/jcbrand/converse.js/commit/b7fb76df2fbe0bd6f3b77412956c6c6bbb5ea3bc

jcbrand:
Your newer fix looks good. I think to wrap up, you can also remove the callback code, and then make an entry in `docs/CHANGES.md`. Then I'll merge.

rlanvin:
Done and thank your for the explanation.

The new branch looks great! I was actually starting to feel lost in the 6k+ lines of `converse.js` file, so if you're going to split it I think it would make it much easier to get into the code. :) How soon are you planning to merge it into master?

jcbrand:
I'll merge it soon, I'm basically waiting for the pull requests to be merged, then I'll merge that branch.

There are more things I want to extract into separate components/modules, like registration and perhaps message carbons. Hell, I might even move out the controlbox code entirely, there was someone recently who said he wants to use the code but without the controlbox.

PR #595, opened 2016-02-17
Fix non_amd.html

rlanvin:
- converse-muc.js was overriding the global `converse` variable
  with undefined (the factory function doesn't return anything)
- `locales` was not defined in utils.js in non AMD (not passed in
  the factory function)

Note: I'm not using any OTR related code, so I do know about that.


jcbrand:
Cheers!

PR #598, opened 2016-02-18
Add synchronize_presence option

davec82:
Add synchronize_presence option

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -679,6 +679,21 @@ Data in localStorage on the other hand is kept indefinitely.
     roster contact statuses will not become out of sync in a single session,
     only across more than one session.
 
+synchronize_availability
+--------------------
+
+* Default: ``true``
+
+Valid options: ``true``, ``false``, ``a resource name``.
+
+This option lets you synchronize your chat status (`online`, `busy`, `away`) with other chat clients. In other words,
+if you change your status to `busy` in a different chat client, your status will change to `busy` in converse.js as well.
+
+If set to ``true``, converse.js will synchronize with all other clients you are logged in with.
+
+If set to ``false``, this feature is disabled.
+
+If set to ``a resource name``, converse.js will synchronize only with a client that has that particular resource assigned to it.
 
 use_otr_by_default
 ------------------
--- a/src/converse-core.js
+++ b/src/converse-core.js
@@ -335,6 +335,7 @@
             sid: undefined,
             sounds_path: '/sounds/',
             storage: 'session',
+            synchronize_availability: true, // Set to false to not sync with other clients or with resource name of the particular client that it should synchronize with
             use_vcards: true,
             visible_toolbar_buttons: {
                 'emoticons': true,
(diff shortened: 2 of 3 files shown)


src/converse-core.js
```diff
@@ -335,6 +335,7 @@
             sid: undefined,
             sounds_path: '/sounds/',
             storage: 'session',
+            synchronize_presence: true, // Set to false to not sync with other clients or with resource name of the particular client that it should synchronize with
             use_vcards: true,
```
jcbrand: Please make this `false` by default. Ideally, existing users who upgrade their version of converse.js shouldn't have to set a new option variable to maintain existing behavior.

davec82: with default value to true, it's the current behavior

docs/source/configuration.rst
```diff
@@ -679,6 +679,17 @@ Data in localStorage on the other hand is kept indefinitely.
...
+--------------------
+
+* Default: ``true``
+
+Valid options: ``true``, ``false``, ``a resource name``.
+
+This option determines how to synchronize presence with other own client resources.
+If set to ``true``, converse.js synchronize presence with other own client resources.
+If set to ``false``, converse.js doesn't synchronize presence with other own client resources.
+If set to ``a resource name``, converse.js synchronize presence with only clients that have particular resource name.
```
jcbrand: Some language suggestions: 

> This option determines how to synchronize presence with other own client resources.

This option lets you synchronize your chat status (`online`, `busy`, `away`) with other chat clients. In other words, if you change your status to `busy` in a different chat client, your status will change to `busy` in converse.js as well.

> If set to ``true``, converse.js synchronize presence with other own client resources.

If set to ``true``, converse.js will synchronize with all other clients you are logged in with.

> If set to ``false``, converse.js doesn't synchronize presence with other own client resources.

If set to ``false``, this feature is disabled.

> If set to ``a resource name``, converse.js synchronize presence with only clients that have particular resource name.

If set to ``a resource name``, converse.js will synchronize only with a client that has that particular resource assigned to it.

docs/source/configuration.rst
```diff
@@ -679,6 +679,17 @@ Data in localStorage on the other hand is kept indefinitely.
     roster contact statuses will not become out of sync in a single session,
     only across more than one session.
 
+synchronize_presence
+--------------------
```
jcbrand: Looking at the spec again, they call it your "availability status". In converse.js I've been calling it "chat status". What makes it more confusing is that there is also a `<status>` element in some `<presence>` stanzas which is the status message that people can write, like "I'm going to town for a bit, will be back later".

In any case, I'd like to keep things as clear and distinct as possible, and if we can keep it closer to the spec that's also better.

So I think `synchronise_availability` is a better name here than `synchronize_presence`.

jcbrand:
Awesome that you added this, thanks. Please take a look at my inline comments and also add an entry to the changelog in `docs/CHANGES.md`.

src/converse-core.js
jcbrand: The current behavior is to not synchronize.

jcbrand: Actually, there might be a whole different way to fix that problem. Instances of converse.js in the same browser (in different tabs) all share the same session storage as far as I'm aware. Each instance however stores it's data under a unique key, so they don't read one another's data. In this case however, we'd like to know if another instance has changed the status. I think what could be done, would be to store the user's status in a global in sessionStorage, for example under the bare JID. Then when one instance changes it, other instances should immediately know about it, since it should change for them as well, since they share the sessionStorage. That's the theory at least.

jcbrand: That could however be left out of the scope of this pull request.

davec82: Actually I have not tried with multiple instances of converse.js in the same browser.

give a couple of  minutes and i'll update pull request with your inline suggestions.

Thanks!

davec82:
Updated pull request, let me know if it's ok.

Thanks.

jcbrand:
Looks good, cheers!

PR #600, opened 2016-02-19
change xmpp status also on icon-status click

davec82:
If you click on icon-status xmpp, you are not able to change status


jcbrand:
Thanks

PR #655, opened 2016-06-06
Fixing

dscdac:
Hi!
Some dependencies of other libraries are imported through one of the Converse.js method. Problem is that the format of the args is different; an Array instead of a String.

### DIFF
--- a/dist/converse-mobile.js
+++ b/dist/converse-mobile.js
@@ -402,6 +402,10 @@ var requirejs, require, define;
     requirejs._defined = defined;
 
     define = function (name, deps, callback) {
+        if(name instanceof Array){
+            name = name[0];
+        }
+
         if (typeof name !== 'string') {
             throw new Error('See almond README: incorrect module build, no module name');
         }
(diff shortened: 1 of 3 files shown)


jcbrand:
Your pull request only contains the generated `dist` files. What about the original source files in the `src` directory? It's also not clear to me what exactly you fixed.

dscdac:
Aw! Somehow the uploaded version is full of crap. Here it is:
https://github.com/jcbrand/converse.js/compare/master...dscdac:fix-sr
The new code only checks the type of the 'name' arg.
I didn't find where this function is generated. I have played more directly with the compiled code converse of my page.
Hope it could be useful!

jcbrand:
The code in the `dist` directory is never modified manually, but instead generated by by bundling converse.js (which is made up of various files in `src`) and its dependencies.

More info here: https://conversejs.org/docs/html/builds.html#creating-builds-and-distribution-files

The snippet you changed looks like it's not from converse.js but from one of its dependencies, specifically from [almond.js](https://github.com/requirejs/almond).

If you are sure it's a legitimate bug then you should report it to the almond.js devs. If the bug has already been fixed in a newer version of almond.js than what converse.js is using, then you can make a pull request which includes the newer version in the bower.json file.

PR #695, opened 2016-09-30
Csp compat

mathiasertl:
I am attempting to use Converse.js as a webclient for https://jabber.at. But we use [Content Security Policy](https://en.wikipedia.org/wiki/Content_Security_Policy) headers to protect us against any XSS attacks. Converse.js uses inline CSS in a few places, which is ignored on our page due to the CSP header.

This pull request replaces some minor inline CSS with CSS styles. This would improve compatibility with CSP headers at least far enough that users can log in, the roster works and chatting works. More work would need to be done, but I thought I'd start with something small. I'm willing to put further work into this if you are interested and we can get this merged.

There is a test that fails with these changes, ``spec/protocol.js:82`` and ``spec/protocol.js:86``, but manually testing reveals no display errors. [This stackoverflow ticket](http://stackoverflow.com/a/8337382) seems to imply that jQuery's `is(":visible")` check could be at fault. I do not know the system well enough to say for sure if that's true or if I actually broke anything.

### DIFF
--- a/docs/CHANGES.md
+++ b/docs/CHANGES.md
@@ -6,6 +6,7 @@
 - New event ['rosterContactsFetched'](https://conversejs.org/docs/html/development.html#rosterContactsFetched) [jcbrand]
 - New event ['rosterGroupsFetched'](https://conversejs.org/docs/html/development.html#rosterGroupsFetched) [jcbrand]
 - HTML templates are now loaded in the respective modules/plugins. [jcbrand]
+- Start improving Content-Security-Policy compatibility by removing inline CSS.
 
 ## 2.0.0 (2016-09-16)
 - #656 Online users count not shown initially [amanzur]
--- a/sass/_controlbox.scss
+++ b/sass/_controlbox.scss
@@ -201,6 +201,12 @@
                     }
                 }
             }
+
+            /* Custom addition for CSP */
+            dd.search-xmpp {
+                display: none;
+            }
+
             dd.search-xmpp ul {
                 box-shadow: 1px 4px 10px 1px rgba(0, 0, 0, 0.4);
                 li:hover {
--- a/sass/_roster.scss
+++ b/sass/_roster.scss
@@ -13,6 +13,11 @@
         height: calc(100% - #{$controlbox-dropdown-height} - 20px);
     }
 
+    /* Custom addition for CSP */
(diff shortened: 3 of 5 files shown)


jcbrand:
Hi @mathiasertl, thanks for the pull request. I'll definitely merge this, and more improvements are very welcome. Please also add your handle or github username after your changelog entry.

I see all the TravisCI tests pass, so I'm not sure why you had failing tests, I assume this was when running them in a "real" browser, right?

I'll be gone over the weekend but can come back and merge this next week.

mathiasertl:
Hi @jcbrand, great to hear this!

I'm not really sure what's happening with the tests. When running `make check`, I get the following error message somewhere in between:

```
mati@dt ~/git/mati/converse.js $ make check
...
Integration of Roster Items and Presence Subscriptions : Subscribe to contact, contact accepts and subscribes back
Error: Expected true to be falsy.
Error: Expected false to be truthy.
...

Finished
-----------------
176 specs, 1 failure in 35.948s.

ConsoleReporter finished
Makefile:188: recipe for target 'check' failed
make: *** [check] Error 1
```
When I check out the master branch, all tests pass. When I comment out the aforementioned lines, they do as well.

With the browser (running `make serve` and then opening "tests.html"), I could first reproduce the same behaviour, but now I get some 96 errors and I don't know why. But I'm not really a node developer and not familiar with the whole ecosystem. I'm sure there could just be that I've forgotten to execute some mandatory command or something.

PS: There is a real world installation with those patches at https://jabber.at/conversejs (that installation also uses local fonts, for CSP reasons, obviously).

jcbrand:
I took a look at this now. The failing tests were just because the CSS wasn't updated. After changing the Sass files, the CSS files still need to be updated by running `make css`. You can also type `make watch` to  have the CSS updated automatically when you change the Sass files.

Concerning the 96 errors when running tests in the browser, sometimes there are errors when the browser runs the tests in a different order than usual. There then seems to be improper teardown and setup between test suites, which causes failures when the tests are run in a certain order. Because this happens intermittently and difficult to reproduce I haven't really been able to fix this.
I would say this happens about 25% of the time when running tests in the browser.

However, when running the tests with PhantomJS, they always execute in the same order, which means we don't have this problem there.

PR #711, opened 2016-10-27
Use roundcube plugin fork with current converse.js

darkpixel:
The previous plugin developer is still running an ancient version of converse.js.  This updates the link to point to a fork that us running a current version of converse.js.


jcbrand:
Thanks! Would you please update the `index.html` file as well?

darkpixel:
Done.  See PR #713

Issue #755, opened 2017-01-13
Add this package in packagist.org


fabiomontefuscolo:
Hello,

Have your package listed in packagist.org would be amazing. PHP developers will be able to install and follow your updates using *composer* (the PHP dependency manager). Do you think it is possible?

Thanks in advance,

jcbrand:
I've accepted your pull request and done the steps you mentioned. Should be set up now. Thanks!

marclaporte:
Already 920 installs!
https://packagist.org/packages/jcbrand/converse.js

Well done JC and Fabio!

marclaporte:
50k installs! https://packagist.org/packages/jcbrand/converse.js

marclaporte:
As a follow-up: https://github.com/conversejs/converse.js/issues/1685

PR #756, opened 2017-01-13
create composer.json to add this project in packagist.org

fabiomontefuscolo:
Hi,

This is a start point to create `composer.json` and add your package in packagist.org. If you accept this pull request and issue #755, you should create an account in packagist.org, submit a new package and wire github webhooks with packagist.

Unfortunately I can't do this for you, but I think it tooks only 10 minutes to accomplish this. If you create an username `jcbrand` in packagist, it will reflect name I wrote in composer.json. If you choose another name the composer.json should be updated.

Thank you!!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,5 @@ css/theme.min.css
 css/mobile.min.css
 
 .sv?
+
+/vendor/
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,9 @@
+{
+    "name": "jcbrand/converse.js",
+    "description": "Browser based XMPP instant messaging client",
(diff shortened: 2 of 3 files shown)


jcbrand:
Thanks! @fabiomontefuscolo

jcbrand:
I've done the steps. Looks good, except that the packagist website says:

`There is no license information available for the latest version (v2.0.4) of this package.`

I assume that's because the `composer.json` file is not in the v2.0.4 tag, right?

fabiomontefuscolo:
Yes, because we can find this info if we select `3.0.x-dev`.

PR #758, opened 2017-01-19
Render all drag resize handles in ChatRoomView

LeoYReyes:
The ChatRoomView renderDragResizeHandles is only rendering the top resize handle (`<div class="dragresize dragresize-top"></div>`). the insert was only inserting the first child of the dragresize template, which has 3 children. Modified to insert the whole the div that the dragresize template gets inserted into.


jcbrand:
Thanks. Please also add an entry to `docs/CHANGES.md`, then I'll merge.

jcbrand:
You pushed the CHANGES.md change to your fork and not to this pull request branch.
So I just made the change manually.

PR #770, opened 2017-02-01
Allow setting contact attrs on chats.open

Ape:
@jcbrand You reverted [parts of my non roster chat commit](https://github.com/jcbrand/converse.js/commit/25d9880f9edc301ce391459d4959dbb8ae0991ff#diff-7a8f4a801d017f221fa2ccba7b7ddcb6L109). Is there a reason for this?

This commit reapplies these parts. I need to be able to set contact attrs with `chats.open`. I use this to set the `fullname` which is missing if the contact is not in the roster.

### DIFF
--- a/src/converse-api.js
+++ b/src/converse-api.js
@@ -106,20 +106,20 @@
             }
         },
         'chats': {
-            'open': function (jids) {
+            'open': function (jids, attrs) {
                 var chatbox;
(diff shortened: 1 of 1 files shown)


jcbrand:
It wasn't clear to me that it was needed. It's merged now, but will only be released with version 3.0.0

PR #776, opened 2017-02-04
french translation review and completion

Jibec:
hi, here is my contribution to your project. It would be really easier to use a translation platform like https://hosted.weblate.org The dev make configuration himself.

I did not try to produce json file because I do not wish to install development tool in this computer, please make the transformation.

### DIFF
--- a/locale/fr/LC_MESSAGES/converse.po
+++ b/locale/fr/LC_MESSAGES/converse.po
@@ -1,13 +1,14 @@
 # Spanish translations for Converse.js package.
 # Copyright (C) 2013 Jan-Carel Brand
 # This file is distributed under the same license as the Converse.js package.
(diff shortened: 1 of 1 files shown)


jcbrand:
Thanks @Jibec 

I had the translations hosted on Transifex for a while, but nobody it to translate.

It's fine for me if people only update the .po files, I can very quickly and easily generate the json files from that.

Jibec:
I'm sure it's fine ;)
But as a translator, you have dozens of projects to follow and there is no way to get a notification when converse.po is modified :(
In addition, most translator are language lovers with low technical skills. Asking to fork a project, change file then to create a pull request is a lot, you may lost some good contributions!
Please consider to bring back converse.js to a translation platform someday ;)

jcbrand:
@Jibec I've asked weblate whether they will host converse.js

Issue #785, opened 2017-02-15
Presence priority NOT handled correctly


w3host:
Hi All, 

The presence priority NOT handled correctly.

Example:
I am logged in two client programs simultaneously (this two client using the same JID).
Cellphone: Extended away status, Priority: 0, Xabber client
Computer: Online, Priority: 1, Pidgin

The conversejs show the last received presence stanza which is not the right presence state.

Received presence state from debug output:
`
  <presence xmlns="jabber:client" to="45a71af3-3597-4abd-b173-10fe25ae21df@kutyakeksz.jabbim.hu/converse.js-21770972" from="kutyakeksz_it@jabbim.hu/c71f218b-0797-4732-8a88-b42cb1d8557a">
    <priority>1</priority>
    <c xmlns="http://jabber.org/protocol/caps" hash="sha-1" ext="voice-v1 camera-v1 video-v1" ver="AcN1/PEN8nq7AHD+9jpxMV4U6YM=" node="http://pidgin.im/"/>
    <x xmlns="vcard-temp:x:update">
      <photo>ce51d94f7f22b87a21274abb93710b9eb7cc1c65</photo>
    </x>
    <delay xmlns="urn:xmpp:delay" stamp="2017-02-15T20:26:05Z" from="jabbim.hu"/>
  </presence>
  <presence xmlns="jabber:client" id="tYRdj-35" to="45a71af3-3597-4abd-b173-10fe25ae21df@kutyakeksz.jabbim.hu/converse.js-21770972" from="kutyakeksz_it@jabbim.hu/androidkhydmcKW">
    <status/>
    <priority>0</priority>
    <show>xa</show>
    <c xmlns="http://jabber.org/protocol/caps" ver="GyIX/Kpa4ScVmsZCxRBboJlLAYU=" hash="sha-1" node="http://www.igniterealtime.org/projects/smack/"/>
    <delay xmlns="urn:xmpp:delay" stamp="2017-02-15T17:02:24Z" from="jabbim.hu"/>
  </presence>
`

In this case conversejs show xa (extended away) status (which is priority 0 and last received) but should show `<priority>1</priority>` status (who is onine).

Based on the rfc3921 (section 2.2.2.1.): "If no `<show/>` element is provided, the entity is assumed to be online and available. "

Lots of time my contact doesn't send me a message because of my wrong presence state.

PR #787, opened 2017-02-19
Presence priority NOT handled correctly #785

w3host:

### DIFF
--- a/docs/CHANGES.md
+++ b/docs/CHANGES.md
@@ -16,6 +16,7 @@
     * Templates are no longer stored as attributes on the `_converse` object.
       If you need a particular template, use `require` to load it.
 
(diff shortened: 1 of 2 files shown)


jcbrand:
Refs #785

jcbrand:
Thank you.

I think however that this implementation doesn't work properly.

If the resource with the highest priority goes offline or becomes unavailable, then the chat status of the contact must fall back to that of the resource with the next highest priority.

In your example from #785, if the resource with priority `1` goes offline or becomes unavailable, then in your implementation the chat status would stay at `online`, although it must actually go to `xa`.

The solution is to update the `resources` attribute on the contact to not just be an array or strings, but instead to be a map of resources to priorities and statuses and to use that data structure.

I'm working on that currently and will commit soon.

Issue #682, opened 2016-08-28
Add "Send" button to input box in chat dialog window


MurzNN:
Will be good to see visible "Send" button to input box in dialog window, for activate send message operation via mouse, without keyboard. Sometimes on Google Chrome when I press "Enter" (without Alt, Ctrl, Shift pressed) - it goes to new line instead of sending message, so there are no way to send message.

MayankPratap:
@jcbrand  I want to work on this issue.

jcbrand:
There's already someone working on it in #796

novokrest:
@jcbrand Hi! I think this issue can be closed

PR #796, opened 2017-03-03
Adds the send button at bottom of chatbox

saganshul:
## Description
Following changes creates a send button functionality at the bottom of chatbox. This is fix regarding the issue #682.

Following changes creates a new function in converse-chatview.js which handles event triggered on clicking the send button and rests are UI changes.

## Reviewer
@jcbrand 

## Screenshots
Please find the attached screenshot of this changes.
screenshot from 2017-03-03 23-39-42

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -976,6 +976,13 @@ show_only_online_users
 If set to ``true``, only online users will be shown in the contacts roster.
 Users with any other status (e.g. away, busy etc.) will not be shown.
 
+show_send_button
+----------------------
+
+* Default:  ``false``
+
+If set to ``true``, a send button will be visible to activate send message operation via mouse, without keyboard.
+
 sounds_path
 -----------
 
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -981,7 +981,7 @@ show_send_button
 
 * Default:  ``false``
 
-If set to ``true``, a send button will be visible to activate send message operation via mouse, without keyboard.
+If set to ``true``, a button will be visible which can be clicked to send a message.
 
 sounds_path
 -----------
--- a/sass/_chatbox.scss
+++ b/sass/_chatbox.scss
@@ -219,6 +219,9 @@
             height: 206px;
             height: calc(100% - #{$toolbar-height + $chat-textarea-height +1px});
         }
+        .chat-content-sendbutton {
+            height: calc(100% - #{$toolbar-height + $chat-textarea-height + $send-button-height + 1px});
+        }
 
         .dropdown { /* status dropdown styles */
             background-color: $light-background-color;
@@ -256,12 +259,11 @@
                 resize: none;
             }
             .chat-textarea-send-button {
-                height: $chat-textarea-height - 35px;
+                height: $chat-textarea-height - $send-button-margin-bottom;
             }
             .send-button {
                 position: absolute;
-                bottom: 5px;
-                right: 10px;
+                right: $send-button-margin-right;
                 background-color: #E76F51;
                 color: #fff;
                 font-size: 80%;
--- a/sass/_chatbox.scss
+++ b/sass/_chatbox.scss
@@ -255,6 +255,17 @@
                 width: 100%;
                 resize: none;
             }
+            .chat-textarea-send-button {
+                height: $chat-textarea-height - 35px;
+            }
+            .send-button {
+                position: absolute;
+                bottom: 5px;
+                right: 10px;
+                background-color: #E76F51;
+                color: #fff;
+                font-size: 80%;
+            }
             .chat-toolbar {
                 box-sizing: border-box;
                 font-size: $font-size;
@@ -388,4 +399,3 @@
         }
     }
 }
-
--- a/sass/_variables.scss
+++ b/sass/_variables.scss
@@ -12,7 +12,7 @@ $light-text-color: #A8ABA1 !default;
 $border-color: #CCC !default;
 $icon-color: #114327 !default;
 $save-button-color: #436F64 !default;
-$chat-textarea-height: 70px !default;
+$chat-textarea-height: 100px !default;
 $message-them-color: #1A9707 !default;
 $roster-height: 194px !default;
 
--- a/sass/_variables.scss
+++ b/sass/_variables.scss
@@ -12,7 +12,10 @@ $light-text-color: #A8ABA1 !default;
 $border-color: #CCC !default;
 $icon-color: #114327 !default;
 $save-button-color: #436F64 !default;
-$chat-textarea-height: 100px !default;
+$chat-textarea-height: 70px !default;
+$send-button-height: 30px !default;
+$send-button-margin-bottom: 5px !default;
+$send-button-margin-right: 10px !default;
 $message-them-color: #1A9707 !default;
 $roster-height: 194px !default;
 
--- a/src/converse-chatview.js
+++ b/src/converse-chatview.js
@@ -605,9 +605,9 @@
                     /* Event handler for when a send button is clicked in a chat box textarea.
                      */
                     ev.preventDefault();
-                    var textarea = $(ev.target).siblings('.chat-textarea')[0],
-                        message;
-                    message = textarea.value;
+                    var textarea = this.el.querySelector('.chat-textarea'),
+                        message = textarea.value;
+
                     textarea.value = '';
                     textarea.focus();
                     if (message !== '') {
--- a/src/converse-chatview.js
+++ b/src/converse-chatview.js
@@ -91,6 +91,7 @@
                 events: {
                     'click .close-chatbox-button': 'close',
                     'keypress .chat-textarea': 'keyPressed',
+                    'click .send-button': 'onSendButtonClicked',
                     'click .toggle-smiley': 'toggleEmoticonMenu',
                     'click .toggle-smiley ul li': 'insertEmoticon',
                     'click .toggle-clear': 'clearMessages',
@@ -119,6 +120,7 @@
                                 _.extend(this.model.toJSON(), {
                                         show_toolbar: _converse.show_toolbar,
                                         show_textarea: true,
+                                        show_send_button: _converse.show_send_button,
                                         title: this.model.get('fullname'),
                                         unread_msgs: __('You have unread messages'),
                                         info_close: __('Close this chat box'),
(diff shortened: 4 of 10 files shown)


src/converse-chatview.js
```diff
@@ -91,6 +91,7 @@
                 events: {
                     'click .close-chatbox-button': 'close',
                     'keypress .chat-textarea': 'keyPressed',
+                    'click .send-button': 'clickSendButton',
```
jcbrand: `clickSendButton` makes it sound like a click action will be performed, when in reality what's happening is that the method is a response to a click action which has already happened.

So rather name this someting like `onSendButtonClicked`.

jcbrand:
Thanks.

I don't want converse.js to have send buttons by default, and since the change requires CSS, it's not so easy to have this option configurable.

I think what we can do here is to make this a separate plugin, and we can then mention it in the documentation and link to it. We can even put the actual code of the plugin into the plugin documentation as a reference implementation.

If you want to contribute to GSoC, then you'll in any case have to learn how to create a plugin, so this will be good experience.

You're welcome to put the plugin code into your own Github repo.

saganshul:
@jcbrand yeah! Plugin is better option to solve this issue.
I will start making plugin for this. It will be great if you can provide me link to some tutorial/documentation/example(implemented plugin).

jcbrand:
@saganshul See here: https://conversejs.org/docs/html/plugin_development.html

Take a look at `converse-muc-embedded.js` as an example of a plugin:
https://github.com/jcbrand/converse.js/blob/master/src/converse-muc-embedded.js

saganshul:
@jcbrand PR updated. Please review and suggest improvements.

src/converse-sendbutton.js
```diff
@@ -0,0 +1,94 @@
...
+                    }
+                    this.setChatState(_converse.ACTIVE);
+                },
+
+                render: function() {
+                    /* Render the UI container in which chat room messages will
+                     * appear.
+                     */
+                    this.__super__.render.apply(this, arguments);
+                    this.$('.sendXMPPMessage').append('<button type="button" class="pure-button send-button">Send</button>');
```
jcbrand: This HTML snippet must be made into a template.

src/converse-sendbutton.js
```diff
@@ -0,0 +1,94 @@
...
+                events: {
+                    'click .send-button': 'onSendButtonClicked'
+                },
+
+                renderChatArea: function() {
+                    /* Render the UI container in which chat room messages will
+                     * appear.
+                     */
+                    this.__super__.renderChatArea.apply(this, arguments);
+                    this.$('.sendXMPPMessage').append('<button type="button" class="pure-button send-button">Send</button>');
```
jcbrand: Here also a template must be used.

src/converse-sendbutton.js
```diff
@@ -0,0 +1,94 @@
...
+                events: {
+                    'click .send-button': 'onSendButtonClicked'
+                },
+
+                render: function() {
+                    /* Render the UI container in which chat room messages will
+                     * appear.
+                     */
+                    this.__super__.render.apply(this, arguments);
+                    this.$('.sendXMPPMessage').append('<button type="button" class="pure-button send-button">Send</button>');
```
jcbrand: Again, template.

src/converse-sendbutton.js
```diff
@@ -0,0 +1,94 @@
...
+                     * appear.
+                     */
+                    this.__super__.renderChatArea.apply(this, arguments);
+                    this.$('.sendXMPPMessage').append('<button type="button" class="pure-button send-button">Send</button>');
+                    return this;
+                }
+            },
+
+            HeadlinesBoxView: {
+                events: {
```
jcbrand: The `HeadlinesBox` doesn't have a form for sending messages, it only shows received messages of type `headline`. This whole section can be removed.

src/converse-sendbutton.js
jcbrand: This method must return what the original `__super__` method returns.

src/converse-sendbutton.js
jcbrand: This method must return what the original `__super__` method returns.

src/converse-sendbutton.js
```diff
@@ -0,0 +1,94 @@
...
+                    'click .send-button': 'onSendButtonClicked'
+                },
+
+                onSendButtonClicked: function(ev) {
+                    /* Event handler for when a send button is clicked in a chat box textarea.
+                     */
+                    var _converse = this.__super__._converse;
+                    ev.preventDefault();
+                    var textarea = $(ev.target).siblings('.chat-textarea')[0],
```
jcbrand: Please rewrite to not use jQuery. We want to remove jQuery from future versions of converse.js.

src/converse-sendbutton.js
```diff
@@ -0,0 +1,94 @@
...
+(function(root, factory) {
+    define([
+        "converse-core",
+        "converse-chatview",
+        "converse-muc",
+        "converse-headline"
+    ], factory);
+}(this, function(converse) {
+    "use strict";
+    var $ = converse.env.jQuery;
```
jcbrand: Please remove the jQuery dependency, we want to remove jQuery from future versions of converse.js

src/converse-sendbutton.js
saganshul: So which library we are using if we are removing Jquery.

src/converse-sendbutton.js
saganshul: Do you want me to do something like this?
``
this.$el.attr('id', this.model.get('box_id'))
                        .html(some_new_template(
                                _.extend(this.model.toJSON(), {
                                        show_toolbar: _converse.show_toolbar,
                                        show_textarea: true,
                                        title: this.model.get('fullname'),
                                        unread_msgs: __('You have unread messages'),
                                        info_close: __('Close this chat box'),
                                        label_personal_message: __('Personal message')
                                    }
                                )
                            )
                        );
``

src/converse-sendbutton.js
jcbrand: There's no need for a library, the browser's DOM API is good enough.

Take a look at this: http://youmightnotneedjquery.com/#siblings

saganshul: @jcbrand  I think should we should make a helper library for this type of most commonly used functions so that we do not write them again and again. what say?

jcbrand: No thanks. I've replaced lots of jQuery code with native API calls already and it's not much more verbose.

src/converse-sendbutton.js
saganshul: @jcbrand is it possible to change template without writing full render function again?

jcbrand: No, I think in this case it's best to create a totally new template containing only the HTML above.

saganshul: @jcbrand I meant to say Can I change only template `tpl_chatbox` to something `tpl_chatbox_sendbutton`(that I will create) in this code: without writing whole code again? 
```javascript
this.$el.attr('id', this.model.get('box_id'))
                        .html(tpl_chatbox(
                            _.extend(this.model.toJSON(), {
                                show_toolbar: _converse.show_toolbar,
                                show_textarea: true,
                                title: this.model.get('fullname'),
                                unread_msgs: __('You have unread messages'),
                                info_close: __('Close this chat box'),
                                label_personal_message: __('Personal message')
                            })
                        ));
                    this.$content = this.$el.find('.chat-content');
                    this.renderToolbar().renderAvatar();
                    _converse.emit('chatBoxOpened', this);
                    utils.refreshWebkit();
                    return this.showStatusMessage();
```

jcbrand: That would go against the whole idea of making this a plugin. By making it a plugin we want to keep it separate from other code.

However, I'm now starting to doubt whether it was a good idea of mine to suggest that we make it a plugin. Perhaps it's better after all to have all this in `converse-chatview.js`.

What originally made me think it would be better as a plugin was because of the fact that you changed the CSS in such a way that it would break without the button. However we can solve that by simply adding a new class to the textarea if the button must be shown and then adding the special button-specific CSS to that class.

So actually, I think it would be best to keep the code in `converse-chatview.js` where you had it originally. I'm sorry for flip-flopping and having given you extra seemingly unnecessary work.

However, the silver lining is that it was good practice. If you're going to work on a GSoC project, then you will definitely have to create a new plugin, so the knowledge you gained now wasn't for nothing.

saganshul: @jcbrand so how we gonna make user selecting one of the option to have button or not?

jcbrand: Yes. I suggest we call it `show_send_button`.

saganshul:
@jcbrand PR updated. Please review.

jcbrand:
Looking good. Almost done. A few small changes (mostly CSS) and documentation needed.

saganshul:
@jcbrand I think I updated the documentation accordingly. If something is left please let me know what else need to add.
what needs to be changed in CSS?

jcbrand:
I forgot to submit yesterdays review. Did that now.

docs/source/configuration.rst
```diff
@@ -976,6 +976,13 @@ show_only_online_users
 If set to ``true``, only online users will be shown in the contacts roster.
 Users with any other status (e.g. away, busy etc.) will not be shown.
 
+show_send_button
+----------------------
+
+* Default:  ``false``
+
+If set to ``true``, a send button will be visible to activate send message operation via mouse, without keyboard.
```
jcbrand: `to activate send message operation` sounds super technical.

Please change to:
```
If set to ``true``, a button will be visible which can be clicked to send a message.
```

sass/_chatbox.scss
```diff
@@ -255,6 +255,17 @@
                 width: 100%;
                 resize: none;
             }
+            .chat-textarea-send-button {
+                height: $chat-textarea-height - 35px;
+            }
+            .send-button {
+                position: absolute;
+                bottom: 5px;
+                right: 10px;
```
saganshul: If we change them to ``em`` then need to change others to em as well. Also these are just position values so I think keeping them in pixels will not be a problem. Please correct me if I am wrong somewhere.

sass/_variables.scss
```diff
@@ -12,7 +12,7 @@ $light-text-color: #A8ABA1 !default;
 $border-color: #CCC !default;
 $icon-color: #114327 !default;
 $save-button-color: #436F64 !default;
-$chat-textarea-height: 70px !default;
+$chat-textarea-height: 100px !default;
```
jcbrand: Ok, but you can still set a different height only for the case where the button is actually there (by having a special class on the textarea).

sass/_chatbox.scss
jcbrand: Ok, I was wondering whether that might be the case. Then leave them as pixels but please make them variables.

Then you can for example set the button height to `$chat-textarea-height - $send-button-height - $send-button-margin-bottom`

saganshul:

jcbrand:
Thanks. I'm traveling the next days and will only be able to get to it sometime next week.

saganshul:
@jcbrand Did you get time to review PR?

jcbrand:
Hi @saganshul 

I took a look at your changes. There were a few outstanding issues that I noticed.

- You didn't commit the generated CSS for your last Sass changes.
- The buttons weren't positioned correctly
- The single-user chat button had the wrong color.

Here's what it looked like after I ran `make css`.

screen shot 2017-03-30 at 12 27 30

I decided to make the last changes myself, so that this can get merged.
I also simplified the Sass/CSS a bit and removed unnecessary variables and rules.
I'll commit that after merging this.

After the changes it now looks like this:

screen shot 2017-03-30 at 12 25 35

Issue #675, opened 2016-08-03
Allow configuration of time format (24h/12h)


dl8bh:
it would be great to be able to configure converse.js to use 24h time format instead of 12h format.

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/36600616-allow-configuration-of-time-format-24h-12h?utm_campaign=plugin&utm_content=tracker%2F194169&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F194169&utm_medium=issues&utm_source=github).

t0kieu:
Or make that functionality more complex, for example time format can be defined per language.

PR #797, opened 2017-03-03
Changed to 24h time format

smitbose:
This is a fix for issue #675 

This changes the time shown in the chatbox from a 12h time format to a 24h time format.

### DIFF
--- a/docs/CHANGES.md
+++ b/docs/CHANGES.md
@@ -4,6 +4,7 @@
 
 - #585 Duplicate contact created due to JID case sensivity [saganshul]
 - #628 Fixes the bug in displaying chat status during private chat. [saganshul]
+- #797 Time format made configurable, with a choice between 12hr and 24hr format. [smitbose]
 - #806 The `_converse.listen` API event listeners aren't triggered. [jcbrand]
 - #807 Error: Plugin "converse-dragresize" tried to override HeadlinesBoxView but it's not found. [jcbrand]
 
--- a/docs/CHANGES.md
+++ b/docs/CHANGES.md
@@ -4,7 +4,7 @@
 
 - #585 Duplicate contact created due to JID case sensivity [saganshul]
 - #628 Fixes the bug in displaying chat status during private chat. [saganshul]
-- #797 Time format made configurable, with a choice between 12hr and 24hr format. [smitbose]
+- #797 Time format made configurable. [smitbose]
 - #806 The `_converse.listen` API event listeners aren't triggered. [jcbrand]
 - #807 Error: Plugin "converse-dragresize" tried to override HeadlinesBoxView but it's not found. [jcbrand]
 
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -1074,9 +1074,9 @@ time_format
 
 Examples: ``HH:MM``, ``hh:mm``, ``hh:mm a``.
 
-This option makes the time format for the time shown, for each message, configurable. Converse uses `moment.js<https://momentjs.com/>`_
+This option makes the time format for the time shown, for each message, configurable. Converse uses `moment.js <https://momentjs.com/>`_
 for showing time. This option allows the configuration of the format in which `moment` will display the time for the messages. For detailed
-description of time-format options available for `moment` visit the website.
+description of time-format options available for `moment` you can check this `link <https://momentjs.com/docs/#/parsing/string-format/>`_.
 
 use_otr_by_default
 ------------------
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -1072,7 +1072,7 @@ time_format
 
 * Default: ``HH:MM``
 
-Valid option: ``HH:MM``, ``hh:mm``, ``hh:mm a``.
+Examples: ``HH:MM``, ``hh:mm``, ``hh:mm a``.
 
 This option makes the time format for the time shown, for each message, configurable. Converse can be configured to show 
 the time in 24hr or 12hr format. The default value gives a 24hr time-format. To change it to 12hr, use `hh:mm`, and to display
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -1074,9 +1074,9 @@ time_format
 
 Examples: ``HH:MM``, ``hh:mm``, ``hh:mm a``.
 
-This option makes the time format for the time shown, for each message, configurable. Converse can be configured to show 
-the time in 24hr or 12hr format. The default value gives a 24hr time-format. To change it to 12hr, use `hh:mm`, and to display
-AM/PM, use `hh:mm a`.
+This option makes the time format for the time shown, for each message, configurable. Converse uses `moment.js<https://momentjs.com/>`_
+for showing time. This option allows the configuration of the format in which `moment` will display the time for the messages. For detailed
+description of time-format options available for `moment` visit the website.
 
 use_otr_by_default
 ------------------
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -1067,6 +1067,17 @@ If set to ``false``, this feature is disabled.
 
 If set to ``a resource name``, converse.js will synchronize only with a client that has that particular resource assigned to it.
 
+time_format
+-----------
+
+* Default: ``HH:MM``
+
+Valid option: ``HH:MM``, ``hh:mm``, ``hh:mm a``.
+
+This option makes the time format for the time shown, for each message, configurable. Converse can be configured to show 
+the time in 24hr or 12hr format. The default value gives a 24hr time-format. To change it to 12hr, use `hh:mm`, and to display
+AM/PM, use `hh:mm a`.
+
 use_otr_by_default
 ------------------
 
--- a/spec/chatbox.js
+++ b/spec/chatbox.js
@@ -1242,25 +1242,19 @@
 
                 it("will render the message time as configured", mock.initConverse(function (_converse) {
                     test_utils.createContacts(_converse, 'current');
-                    test_utils.openControlBox();
-                    test_utils.openContactsPanel(_converse);
-
+                    
                     _converse.time_format = 'hh:mm';
-                    spyOn(_converse, 'emit');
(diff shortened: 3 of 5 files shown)


smitbose:
screenshot from 2017-03-04 02-19-58

jcbrand:
The point of that ticket is to make the format configurable, not to hardcode it to a new format.

So you want to add a new configuration setting, something like `date_format`, let it have the original format as it's default and then let the user specify a different format if they want one.

smitbose:
ok, @jcbrand I get it now, I'll try to generate a new patch with the required change.

smitbose:
@jcbrand I have added a setting time_format, that takes vales 'h' and 'H' for 12 and 24 hr time formats respectively. Also, for the 12hr time format, it will display am or pm, because personally I found this am or pm not being displayed as making reading of previous messages too ambiguous.
screenshot from 2017-03-08 11-14-26

src/converse-chatview.js
```diff
@@ -344,11 +344,19 @@
...
+                        mess_time = msg_time.format('hh:mm a');
+                    else if(_converse.time_format === 'H')
+                        mess_time = msg_time.format('HH:MM');
+
                     var $msg = $(template(
                         _.extend(this.getExtraMessageTemplateAttributes(attrs), {
                             'msgid': attrs.msgid,
                             'sender': attrs.sender,
-                            'time': msg_time.format('hh:mm'),
+                            'time': mess_time,
```
smitbose: @jcbrand in one of the above comments, you have told to keep the original one(which according to current release is hh:mm) as the default. Can you confirm once again, which one will I finally keep .

Moreover, doing the above check helps in printing am or pm beside the time in 12hr format. Because for that moment needs hh:mm a. One way is of-course to let the user specify 'hh:mm a' as the setting. And the other way would be to simply leave the printing of 'am' or 'pm' in messages, which I feel makes a message sent at 6am and 6pm confusing.

jcbrand: My original point had less to do with what the default was and more to do with the fact that you hardcoded the change.

I think I prefert 24-hour time as well, so I'd prefer HH:MM.

Concerning showing AM and PM, since we want to make the user (really the integrator) responsible for choosing the format, they can decide whether to show AM/PM or not.

smitbose: ok, so I'm making a change with, keeping three settings values then, 'HH:MM', 'hh:mm' and 'hh:mm a', which will give a 24hr time format, 12hr time format, and a 12hr time format with am/pm displayed.

jcbrand: No. That's not what I meant. We want the user to specify the format. Whatever format that is. Don't try to restrict the format that the user can provide.

src/converse-core.js
```diff
@@ -280,6 +281,10 @@
             }
         }
 
+        if(this.time_format !== 'h' && this.time_format !== 'H') {
+            throw new Error("Config Error: wrong time format configuration");
+        }
+
```
jcbrand: This must be removed since we don't want to restrict the values the user can enter as the format.

src/converse-core.js
```diff
@@ -267,7 +267,7 @@
             whitelisted_plugins: [],
             xhr_custom_status: false,
             xhr_custom_status_url: '',
-            time_format: 'h', // H stands for 24hr time format and 'h' stands for 12hr time format
+            time_format: 'HH:MM', // H stands for 24hr time format and 'h' stands for 12hr time format
```
jcbrand: Please remove the comment.

src/converse-core.js
```diff
@@ -281,7 +281,7 @@
             }
         }
 
-        if(this.time_format !== 'h' && this.time_format !== 'H') {
+        if(this.time_format !== 'hh:mm' && this.time_format !== 'HH:MM' && this.time_format !== 'hh:mm a') {
```
jcbrand: Remove this whole `if` statement. We don't want to restrict what format the user can provide.

smitbose: okay.

src/converse-chatview.js
```diff
@@ -344,11 +344,13 @@
                                "Output has been shortened."),
                             true, true);
                     }
+
+
```
jcbrand: Unnecessary whitespace.

src/converse-core.js
```diff
@@ -267,6 +267,7 @@
             whitelisted_plugins: [],
             xhr_custom_status: false,
             xhr_custom_status_url: '',
+            time_format: 'HH:MM', // H stands for 24hr time format and 'h' stands for 12hr time format
```
jcbrand: Please remove the comment.

jcbrand:
@smitbose Almost ready for merge. What still needs to happen is the documentation needs to be updated.

Please add an entry for `time_format` in the configuration settings page:
https://github.com/jcbrand/converse.js/blob/master/docs/source/configuration.rst

Some info on writing documentation: https://conversejs.org/docs/html/documentation.html

Then also update the changelog in `docs/CHANGES.md`

docs/source/configuration.rst
```diff
@@ -1067,6 +1067,17 @@ If set to ``false``, this feature is disabled.
 
 If set to ``a resource name``, converse.js will synchronize only with a client that has that particular resource assigned to it.
 
+time_format
+-----------
+
+* Default: ``HH:MM``
+
+Valid option: ``HH:MM``, ``hh:mm``, ``hh:mm a``.
+
```
smitbose: okay

docs/source/configuration.rst
```diff
@@ -1067,6 +1067,17 @@ If set to ``false``, this feature is disabled.
...
+This option makes the time format for the time shown, for each message, configurable. Converse can be configured to show 
+the time in 24hr or 12hr format. The default value gives a 24hr time-format. To change it to 12hr, use `hh:mm`, and to display
+AM/PM, use `hh:mm a`.
+
```
jcbrand: Also the text here is too specific, it makes me think that you're still missing the point somewhat, I think because of the way the original ticket was specified.

What we are doing is letting the user specify **any** format for message time, they can even include the date if they want to. So don't say it's either 24hr or 12hr, just say that it can show any format that moment.js supports.

smitbose: okay, I'll try to make it more clear

jcbrand:
One more thing. Please add a test to show that different formats can be used. You can add this test in [spec/chatbox.js](https://github.com/jcbrand/converse.js/blob/master/spec/chatbox.js).

Let me know if you need help.

smitbose:
@jcbrand as far as the documentation is concerned, adding a link to moment.js will give the user a better idea of the different formats they can use, along with an explanation with an instance. 
As for the tests, would just checking for the value of `_converse.time_format` suffice, or should I check the time of a dummy message sent?

jcbrand:
+1 for adding a link (if you find a place in the moment.js docs where the different formats are shown).

Concerning the test: set a non-default value for the `time_format` and the check that a sent messages renders correctly.

smitbose:
https://momentjs.com/docs/#/parsing/string-format/

I think this link makes a place in the doc. What say?

spec/chatbox.js
```diff
@@ -1240,6 +1240,33 @@
                     });
                 }));
 
+                it("will render the message time as configured", mock.initConverse(function (_converse) {
+                    test_utils.createContacts(_converse, 'current');
+                    test_utils.openControlBox();
+                    test_utils.openContactsPanel(_converse);
+
```
jcbrand: I think lines 1245 and 1246 can be removed.

spec/chatbox.js
```diff
@@ -1240,6 +1240,33 @@
...
+                    _converse.time_format = 'hh:mm';
+                    spyOn(_converse, 'emit');
+                    var contact_jid = mock.cur_names[0].replace(/ /g,'.').toLowerCase() + '@localhost';
+                    test_utils.openChatBoxFor(_converse, contact_jid);
+                    expect(_converse.emit).toHaveBeenCalledWith('chatBoxFocused', jasmine.any(Object));
```
jcbrand: Irrelevant check which can be removed. You can then also remove the `spyOn(_converse, 'emit');` line.

spec/chatbox.js
```diff
@@ -1240,6 +1240,33 @@
...
+                    var view = _converse.chatboxviews.get(contact_jid);
+                    var message = 'This message is sent from this chatbox';
+                    spyOn(view, 'sendMessage').andCallThrough();
+                    test_utils.sendMessage(view, message);
+                    expect(view.sendMessage).toHaveBeenCalled();
+                    expect(view.model.messages.length, 2);
```
jcbrand: The two `expect` lines above are irrelevant for this test and should be removed. Also the `spyOn` above that.

spec/chatbox.js
```diff
@@ -1240,6 +1240,33 @@
...
+
+                    var chatbox = _converse.chatboxes.get(contact_jid);
+                    expect(chatbox.messages.models.length, 1);
+                    var msg_object = chatbox.messages.models[0];
+                    var msg_time_author = view.$el.find('.chat-content').find('.chat-message').last().find('.chat-msg-author.chat-msg-me').text();
```
jcbrand: Wrap this line please, so that it's not so long.

docs/CHANGES.md
```diff
@@ -4,6 +4,7 @@
 
 - #585 Duplicate contact created due to JID case sensivity [saganshul]
 - #628 Fixes the bug in displaying chat status during private chat. [saganshul]
+- #797 Time format made configurable, with a choice between 12hr and 24hr format. [smitbose]
```
jcbrand: Update this entry please. Still falsely creates the impression that there's 2 choices.

jcbrand:
@smitbose I'm not sure whether you were still busy with this or already waiting for review, but I went ahead and did a quick review anyway.

docs/CHANGES.md
smitbose: yeah @jcbrand I think this 12hr/24hr thing has got embedded in my head! :P I keep repeating the same damn thing every single time! :D

smitbose:
@jcbrand please review

jcbrand:
Looking good thanks!

Issue #585, opened 2016-02-10
Duplicate contacts - case sensivity


Zerebokep:
It's possible to add the same contact multiple times by writing the JID different.

e.g. Test@test.com and test@test.com.

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/30734802-duplicate-contacts-case-sensivity?utm_campaign=plugin&utm_content=tracker%2F194169&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F194169&utm_medium=issues&utm_source=github).

<bountysource-plugin>

saganshul:
@jcbrand You can close this issue now. :)

PR #799, opened 2017-03-04
Fixes the duplication bug in adding contact

saganshul:
## Description

This PR fixes the bug mentioned in #585. Bug is on adding contact say **Singhal@conversejs.org** after **singhal@conversejs.org**(which are both same) gets add to contact list. This behaviour is bug because singhal@conversejs.org and Singhal@conversejs.org both are technically same JID.  Hence converting the JID to lowercase anytime user add a contact will solve this issue.

## Reviewer
@jcbrand 
 
## Screenshots
11f3a944938e934847b04b94a4308510
0b8c6f5d2cd5159347545903da151e48
584f6587390942d79f9dee6c59ac5a0d

### DIFF
--- a/docs/CHANGES.md
+++ b/docs/CHANGES.md
@@ -64,6 +64,7 @@
   New configuration setting:
   [muc_show_join_leave](https://conversejs.org/docs/html/configuration.html#muc-show-join-leave)
 - #366 Show the chat room occupant's JID in the tooltip (if you're allowed to see it). [jcbrand]
+- #585 Fixes the duplication bug due to case sensivity in adding contacts [saganshul]
 - #610, #785 Add presence priority handling [w3host, jcbrand]
 - #620 `auto_away` shouldn't change the user's status if it's set to `dnd`. [jcbrand]
 - #694 The `notification_option` wasn't being used consistently. [jcbrand]
--- a/spec/controlbox.js
+++ b/spec/controlbox.js
@@ -1087,6 +1087,34 @@
             // XXX: Awaiting more tests, close it again for now...
             panel.$el.find('a.toggle-xmpp-contact-form').click();
         }));
+
+        it("can be used to add contact and it checks for case-sensivity", mock.initConverse(function (_converse) {
+            spyOn(_converse, 'emit');
+            spyOn(_converse.rosterview, 'update').andCallThrough();
+            runs(function () {
+                test_utils.openControlBox();
+                // Adding two contacts one with Capital initials and one with small initials of same JID (Case sensitive check)
+                _converse.roster.create({
+                    jid: mock.pend_names[0].replace(/ /g,'.').toLowerCase() + '@localhost',
+                    subscription: 'none',
+                    ask: 'subscribe',
+                    fullname: mock.pend_names[0]
+                });
+                _converse.roster.create({
+                    jid: mock.pend_names[0].replace(/ /g,'.') + '@localhost',
+                    subscription: 'none',
+                    ask: 'subscribe',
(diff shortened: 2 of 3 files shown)


jcbrand:
Thank you @saganshul 

Please also add a test for this. You can add it here: https://github.com/jcbrand/converse.js/blob/master/spec/controlbox.js#L1079

To run tests you can run `make check` but for debugging it's easier to run them in the browser. To do this run `make serve` and then open `localhost:8000/tests.html`

You can click on the individual test title to run only that test.

Also please update the changelog.

saganshul:

docs/CHANGES.md
```diff
@@ -61,6 +61,7 @@
   an instant room is created. [jcbrand]
 - Ensure consistent behavior from `show_controlbox_by_default` [jcbrand]
 - #365 Show join/leave messages for chat rooms.
+- #585 Fixes the duplication bug due to case sensivity in adding contacts.
```
jcbrand: You've spilt up a single changelog entry and put your new entry in the middle of it.

spec/controlbox.js
```diff
@@ -1087,6 +1087,33 @@
             // XXX: Awaiting more tests, close it again for now...
             panel.$el.find('a.toggle-xmpp-contact-form').click();
         }));
+
+        it("can be added to roster and check for case-sensivity", mock.initConverse(function (_converse) {
```
jcbrand: The `describe` section and the `it` section should together read like a sentence.

For example: "The 'Add Contact' widget opens up an add form when you click on it.

What you have done is to create the following sentence:
"The 'Add Contact widget" can be added to roster and check for case-sensitivity"
which doesn't make sense.

docs/CHANGES.md
saganshul: I did that because I thought the logs below that belongs to **New configuration setting:** section. Is that the case or should I add log at the end of this or somewhere else?

spec/controlbox.js
```diff
@@ -1087,6 +1087,33 @@
...
+                _converse.roster.create({
+                    jid: mock.pend_names[0].replace(/ /g,'.') + '@localhost',
+                    subscription: 'none',
+                    ask: 'subscribe',
+                    fullname: mock.pend_names[0]
+                });
+            });
+            waits(300);
+            runs(function () {
+                expect(_converse.rosterview.$el.find('dd:visible').length).toBe(1); // Checking that only one entry is created because both JID is same (Case sensitive check)
```
jcbrand: Please put the comment on a new line.

jcbrand:
@saganshul Looking good, just a few more small changes please.

spec/controlbox.js
saganshul: Oh! I didn't know that. Should I write something like
""The 'Add Contact widget" can be used to add contacts to roster and it check for case-sensitivity"
?

saganshul:
@jcbrand PR updated as mentioned. Please review.

docs/CHANGES.md
```diff
@@ -70,6 +70,7 @@
 - #745 New config option [priority](https://conversejs.org/docs/html/configuration.html#priority) [jcbrand]
 - #770 Allow setting contact attrs on chats.open [Ape]
 - #790 MAM retrieval broken [jcbrand]
+- #585 Fixes the duplication bug due to case sensivity in adding contacts [saganshul]
```
jcbrand: Sorry for being pedantic here, but I prefer it if the issue numbers are sorted.

jcbrand:
Thank you @saganshul :)

Issue #628, opened 2016-04-06
Typing chat status


themaverik:
I'm logged in with message carbons enabled. When logged in (say A) from multiple resources R1 and R2 and chatting with B, when A starts typing from R1, R2 shows as B is typing although B is idle which means that the carbon message should not show that the other person is typing when i'm typing. This is a minor bug, but it can be misleading when i'm on different devices at the same time with the wrong chat status.

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/32654552-typing-chat-status?utm_campaign=plugin&utm_content=tracker%2F194169&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F194169&utm_medium=issues&utm_source=github).

<bountysource-plugin>

PR #801, opened 2017-03-04
Fixes the bug in displaying chat status during private chat (Message carbons)

saganshul:
## Description
This PR fixes the bug mentioned in #628 
The bug was actually in assignment of `fullname` variable. It should always be set to `from`.
So what was happening `fullname` variable in private chat(message carbons enabled) was getting set to `this.get('fullname)'` which is actually not the correct `fullname` in this case and due to this variable is set, `_.isEmpty(fullname)` results into `false` and this operation `fullname = from;` fail to carry out.

Hence the perfect fix should be to just to initialise it first and then set it to `from` .

## Reviewer
@jcbrand 

## Explaination of screenshot(for testing):
So to test I registered with two accounts on conversejs.org -
* singhal@conversejs.org
* anshul@conversejs.org.
In screenshots - First block is of singhal@conversejs.org and rest two are of anshul@conversejs.org(to test message carbons - two end points). I used in-cognitive mode of google chrome
In first image: singhal@conversejs.org is typing and the same information is reflected in other two.
In second image: first(middle block) anshul@conversejs.org is typing and the same info is reflected which was previously a bug because block 3 was shows singhal@conversejs.org which was a bug. Hence after this fix this is not a case.

I checked this fix with group chat also and it is working correctly.

### DIFF
--- a/docs/CHANGES.md
+++ b/docs/CHANGES.md
@@ -72,6 +72,7 @@
 - #585 Fixes the duplication bug due to case sensivity in adding contacts [saganshul]
 - #610, #785 Add presence priority handling [w3host, jcbrand]
 - #620 `auto_away` shouldn't change the user's status if it's set to `dnd`. [jcbrand]
+- #628 Fixes the bug in displaying chat status during private chat (Message carbons) [saganshul]
 - #694 The `notification_option` wasn't being used consistently. [jcbrand]
 - #745 New config option [priority](https://conversejs.org/docs/html/configuration.html#priority) [jcbrand]
 - #770 Allow setting contact attrs on chats.open [Ape]
--- a/docs/CHANGES.md
+++ b/docs/CHANGES.md
@@ -4,6 +4,7 @@
 
 - #806 The `_converse.listen` API event listeners aren't triggered. [jcbrand]
 - #807 Error: Plugin "converse-dragresize" tried to override HeadlinesBoxView but it's not found. [jcbrand]
+- #628 Fixes the bug in displaying chat status during private chat (Message carbons) [saganshul]
 
 ## 3.0.0 (2017-03-05)
 
@@ -72,7 +73,6 @@
 - #585 Fixes the duplication bug due to case sensivity in adding contacts [saganshul]
 - #610, #785 Add presence priority handling [w3host, jcbrand]
 - #620 `auto_away` shouldn't change the user's status if it's set to `dnd`. [jcbrand]
-- #628 Fixes the bug in displaying chat status during private chat (Message carbons) [saganshul]
 - #694 The `notification_option` wasn't being used consistently. [jcbrand]
 - #745 New config option [priority](https://conversejs.org/docs/html/configuration.html#priority) [jcbrand]
 - #770 Allow setting contact attrs on chats.open [Ape]
--- a/docs/CHANGES.md
+++ b/docs/CHANGES.md
@@ -4,7 +4,7 @@
 
 - #806 The `_converse.listen` API event listeners aren't triggered. [jcbrand]
 - #807 Error: Plugin "converse-dragresize" tried to override HeadlinesBoxView but it's not found. [jcbrand]
-- #628 Fixes the bug in displaying chat status during private chat (Message carbons) [saganshul]
+- #628 Fixes the bug in displaying chat status during private chat [saganshul]
 
 ## 3.0.0 (2017-03-05)
 
--- a/docs/CHANGES.md
+++ b/docs/CHANGES.md
@@ -2,9 +2,9 @@
 
 ## 3.0.1 (Unreleased)
 
+- #628 Fixes the bug in displaying chat status during private chat. [saganshul]
 - #806 The `_converse.listen` API event listeners aren't triggered. [jcbrand]
 - #807 Error: Plugin "converse-dragresize" tried to override HeadlinesBoxView but it's not found. [jcbrand]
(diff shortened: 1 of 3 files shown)


jcbrand:
We don't want to always show `from` since we rather want to show a user's full name if we have it, instead of their JID.

Also, it doesn't make sense to declare `fullname` as `undefined` and then to do an `isEmpty` check further down. We know it's always empty if we have defined it as such, so that check is unnecessary.

I think a better fix is to set the value of fullname where the code determines who sent the message (`me` or `them`).
Here: https://github.com/jcbrand/converse.js/blob/06668d8733a463ff621ae82a7c05270fb98d7d2e/src/converse-core.js#L1402

If the message is from me, then `fullname  = converse.xmppstatus.get('fullname');` can be used. If the message is from then, then you set `fullname  = isEmpty(this.get('fullname')) ? from : this.get('fullname');`

saganshul:
@jcbrand `_converse.xmppstatus.get('fullname');` is giving me `undefined`.
Upon priting `_converse.xmppstatus` object in console I found no `fullname` attribute. Can you please help me out here?

```javascript
if ((is_groupchat && from === this.get('nick')) || (!is_groupchat && from === _converse.bare_jid)) {
           sender = 'me';
           fullname = _converse.xmppstatus.get('fullname');
      } else {
          sender = 'them';
          fullname = _.isEmpty(this.get('fullname')) ? from : this.get('fullname');
}
```

jcbrand:
Try this:

```
if ((is_groupchat && from === this.get('nick')) || (!is_groupchat && from === _converse.bare_jid)) {
           sender = 'me';
           fullname = _converse.xmppstatus.get('fullname') || from;
      } else {
          sender = 'them';
          fullname = this.get('fullname') || from;
}
```

saganshul:

src/converse-core.js
```diff
@@ -1377,7 +1377,7 @@
                     body = _.propertyOf(message.querySelector('body'))('textContent');
                 }
                 var delayed = !_.isNull(delay),
-                    fullname = this.get('fullname'),
+                    fullname,
```
jcbrand: This variable declaration is unnecessary, the variable is never used.

jcbrand:
Looks good @saganshul, but to be sure that it actually works, I'd like to see a test which exposes the bug mentioned in #628.

This test must fail without your changes and must pass with your changes.

You can put your test in `spec/chatbox.js` under the `A composing notification` "describe" section.

Something like this:

```javascript
describe("A composing notification", function () {
    it("shows the fullname or JID of the user", mock.initConverse(function (_converse) 
     });
}));
```

What you want to do is to recreate the conditions mentioned in #628 

You can emulate the reception of a specific stanza (for example a carbons message) by calling `test_utils.createRequest(stanza)`. There are lots of examples of me doing that in the tests.

That way you can create the exact stanza that causes the bug to appear.

saganshul:
@jcbrand I think the test thing which you mentioned is already present. https://github.com/saganshul/converse.js/blob/85c0c69a4ad0a157626288b7a97db68f654f0a11/spec/chatbox.js#L861

Please correct me if I am wrong.

jcbrand:
Yes, that looks like a relevant test, but if it actually tested for the bug in #628 then it would have failed, so it's not actually testing that particular bug.

I think you can just update it to expose the bug. In other words, update the test so that without your changes it fails because of the bug and then with your changes applied it passes.

saganshul:
@jcbrand Should I submit two seperate PRs for test and fix then?

jcbrand:
No you can update the test in this pull request. 

When I say you must test it without your changes, I mean that you can do this on your own machine, without having to make a separate pull request without your changes.

This is generally how you do test-driven-development. When you find a bug, you first write a test that assumes correct working of the code without the bug. Because the bug however exists, the test will fail.

Then you fix the bug and the test will pass.

Because you have already fixed the bug, if you now write the test, it will pass. However, just to make sure that your test is working correctly, it's good to disable your fix and to then check that the test fails.

That's what I'm referring to.

saganshul:
@jcbrand PR updated. Please review. 
As you suggested I first created a commit to update the present test to expose the bug. Then following commit fixes the bug.
I hope I get you right. Please let me know your thoughts on the same.

jcbrand:
Looks good thanks! Please rebase your commits onto master, then update the changelog then I'll merge.

saganshul:
@jcbrand Rebased the PR onto master and changes added to changelog. 
Please review :)

jcbrand:
You're adding it to the changelog of the previous release, it should go under `3.0.1`.

docs/CHANGES.md
```diff
@@ -4,6 +4,7 @@
 
 - #806 The `_converse.listen` API event listeners aren't triggered. [jcbrand]
 - #807 Error: Plugin "converse-dragresize" tried to override HeadlinesBoxView but it's not found. [jcbrand]
+- #628 Fixes the bug in displaying chat status during private chat (Message carbons) [saganshul]
```
jcbrand: Alphanumerical please.

saganshul:
@jcbrand change log added to 3.0.1 . Please review.

jcbrand:
No, sorry, that's not what I meant with alphanumeric. That was probably too cryptic and the wrong word that I used.

Generally we order the ticket numbers from lowest to highest. So 628 must go above 806.

saganshul:
@jcbrand PR updated.

jcbrand:
Merged, thanks!

Issue #803, opened 2017-03-05
Wrong path to fonticons


jkufner:
converse.js/css/converse.css contains following references:

```
@font-face {
  font-family: 'Converse-js';
  src: url("/fonticons/fonts/icomoon.eot?wvi0ht");
  src: url("/fonticons/fonts/icomoon.eot?wvi0ht#iefix") format("embedded-opentype"), ...
```

These paths should be relative, so the Converse.js can be installed in subdirectory.

Once the path is prepended with `../`, it works:
```
@font-face {
  font-family: 'Converse-js';
  src: url("../fonticons/fonts/icomoon.eot?wvi0ht");
  ...
```

PR #804, opened 2017-03-05
Makes the path relative instead of absolute for fonts

saganshul:
This PR is regarding the issue #803 .

I can see that you also had same code in comment. 
Please let me know your thoughts on the same.

### DIFF
--- a/sass/_variables.scss
+++ b/sass/_variables.scss
@@ -1,4 +1,4 @@
-/* 
+/*
     Color scheme helpers:
     https://coolors.co/app/264653-2a9d8f-e9c46a-f4a261-e76f51
     http://paletton.com/#uid=53f0u0knsvIdILAj5Cftgu3uBmZ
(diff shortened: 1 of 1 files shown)


jcbrand:
Actually, the value used to be `https://cdn.conversejs.org/fonticons/fonts/`, so please change it back to that.

I changed it recently because I updated the fonts and they weren't yet available in the CDN.

Also please put an entry in the changelog in `docs/CHANGES.md` to document that you made this change.

saganshul:
@jcbrand URL changed to CDN.

jkufner:
> Actually, the value used to be https://cdn.conversejs.org/fonticons/fonts/, so please change it back to that.

Please don't use CDN by default. Relative paths allow to use CDN too, but they also allow to not use it. CDN introduces hidden dependency on 3rd party service and also leaks private data to CDN provider.

jcbrand:
Ok, I merged only the first commit from @saganshul which makes it relative.

PR #808, opened 2017-03-08
Discussion

theSouL:
Adding support for https://xmpp.org/extensions/xep-0382.html

Spoiler messages.

### DIFF
--- a/src/converse-chatview.js
+++ b/src/converse-chatview.js
@@ -354,6 +354,9 @@
                             'extra_classes': this.getExtraMessageClasses(attrs)
                         })
                     ));
+                    //Where the argument "attrs" come from? Could "attrs" have information from onMessage?
+                    //Could be possible (and would it be ok) to place a button instead of .text(text)?
+                    //Making the button to show the text set here onclick?
                     $msg.find('.chat-msg-content').first()
                         .text(text)
                         .addHyperlinks()
--- a/src/converse-core.js
+++ b/src/converse-core.js
@@ -1479,7 +1479,11 @@
             onMessage: function (message) {
                 /* Handler method for all incoming single-user chat "message"
                  * stanzas.
-                 */
+                 */          
+                //How can I propagate an element found in message.childNodes until the message is displayed on the chat?
+                //I've seen in some functions, message's attributes are used. Should I setAttribute() to the message
+                //to be able to check the element's information later?
+                
                 var original_stanza = message,
                     contact_jid, forwarded, delay, from_bare_jid,
                     from_resource, is_me, msgid,


src/converse-chatview.js
```diff
@@ -354,6 +354,9 @@
                             'extra_classes': this.getExtraMessageClasses(attrs)
                         })
                     ));
+                    //Where the argument "attrs" come from? Could "attrs" have information from onMessage?
+                    //Could be possible (and would it be ok) to place a button instead of .text(text)?
+                    //Making the button to show the text set here onclick?
```
jcbrand: Use `grep` to look at all the places from where `renderMessage` called. You'll see it's called by `showMessage` and `insertMessage` and they both also accept `attrs`.

So then you can check where and how they are called.

`showMessage` is for example called here: https://github.com/jcbrand/converse.js/blob/master/src/converse-chatview.js#L414

But also in other places.

Eventually you'll see that `attrs` are the attributes on the `Message` Backbone.Model.
And that message model is created here:
https://github.com/jcbrand/converse.js/blob/master/src/converse-core.js#L1420

And from that we can see that attributes of a message are determined by `getMessageAttributes`
Here: https://github.com/jcbrand/converse.js/blob/master/src/converse-core.js#L1369

src/converse-core.js
```diff
@@ -1479,7 +1479,11 @@
             onMessage: function (message) {
                 /* Handler method for all incoming single-user chat "message"
                  * stanzas.
-                 */
+                 */          
+                //How can I propagate an element found in message.childNodes until the message is displayed on the chat?
+                //I've seen in some functions, message's attributes are used. Should I setAttribute() to the message
+                //to be able to check the element's information later?
+
```
jcbrand: You probably want to look at `getMessageAttributes` mentioned above.

src/converse-chatview.js
theSouL: `                    if(typeof attrs.spoiler !== 'undefined' && attrs.spoiler != 'no' ){
                        var button = document.createElement("button");
                        button.type = "button";
                        button.value = attrs.spoiler;
                        button.textContent = attrs.spoiler;

                        button.onclick = function (){
                            alert("Button clicked");
                        };


                        $msg.find('.chat-msg-content').first()
                        //.text(attrs.spoiler)
                        .addHyperlinks()
                        .append(button)
                        .addEmoticons(_converse.visible_toolbar_buttons.emoticons);
                        
                    }else{
                        
                    $msg.find('.chat-msg-content').first()
                        .text(text)
                        .addHyperlinks()
                        .addEmoticons(_converse.visible_toolbar_buttons.emoticons);
                    }
                    return $msg;
                },
`

This is a bit redundant at this point but I'm interested in the
`                        button.onclick = function (){
                            alert("Button clicked");
                        };`

part.

How can I set this function on the button element? Before thinking about creating a specific function for this matter, I wanted to try to create an anonymous function as you can see in the code.
The problem is, in the HTML, there's no onclick attribute. I tried other methods like .on(click.., .bind(, setAttribute, etc. But none of them work.

I guess it is a pretty basic issue, but I'm not able to find the solution.

Thank you :)

jcbrand: You have to register the event on the relevant Backbone.View.

For example here some event handlers are registered:
https://github.com/jcbrand/converse.js/blob/9df3715ac881b608a88bc9cf3ffd1e7bc6a0f815/src/converse-chatview.js#L91

Issue #704, opened 2016-10-20
Default domain for registration


dnut:
Just like we can set up default domains, we should be able to setup a default registration form, especially if we are locking the client to a single domain. That way, the user does not have to type in our domain name to register. The register tab should open straight to our registration form.

smitbose:
@jcbrand should I add a config setting to open up the registration form to a default domain specified by the user?

jcbrand:
Yes please @smitbose

PR #812, opened 2017-03-13
Implemented auto fetching of registration form for default XMPP domain

smitbose:
This is an implementation of the feature requested in #704 .

A new configuration variable `enable_default_domain` which when set to true, the registration form on initiation reads the domain from `default_domain_name` and fetches the registration form for the same. It doesn't allow the user to input the domain or cancel the fetched registration form by any means.

@jcbrand please review the codes, while I'm moving on to adding documentation for the added configurations, as well as updating changelog. I don't think any specific tests can be written for this because the behaviour is same as the mainline sequence of taking user input and fetching registraion form, except with forced event triggers.

### DIFF
--- a/docs/CHANGES.md
+++ b/docs/CHANGES.md
@@ -6,6 +6,8 @@
 - #628 Fixes the bug in displaying chat status during private chat. [saganshul]
 - #628 Changes the message displayed while typing from a different resource of the same user. [smitbose]
 - #675 Time format made configurable. [smitbose]
+- #704 Automatic fetching of registration form when [registration_domain](https://conversejs.org/
+docs/html/configurations.html#registration-domain) is set. [smitbose]
 - #806 The `_converse.listen` API event listeners aren't triggered. [jcbrand]
 - #807 Error: Plugin "converse-dragresize" tried to override HeadlinesBoxView but it's not found. [jcbrand]
 - #820 Inconsistency in displaying room features. [jcbrand]
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -483,6 +483,14 @@ JIDs with other domains are still allowed but need to be provided in full.
 To specify only one domain and disallow other domains, see the `locked_domain`_
 option.
 
+registration_domain
+-------------------
+
+* Default: ``''``
+
+Specify a domain name for which the registration form will be fetched automatically,
+without the user having to enter any XMPP server domain name.
+
 default_state
 -------------
 
--- a/src/converse-core.js
+++ b/src/converse-core.js
@@ -256,6 +256,7 @@
             password: undefined,
             prebind_url: null,
             priority: 0,
+            registration_domain: '',
             rid: undefined,
             roster_groups: true,
             show_only_online_users: false,
--- a/src/converse-register.js
+++ b/src/converse-register.js
@@ -72,6 +72,11 @@
                             'model': this
                         });
                         this.registerpanel.render().$el.addClass('hidden');
+                        
+                        if (_converse.registration_domain) {
+                            this.registerpanel.renderRegistrationRequest('');
+                            this.registerpanel.fetchRegistrationForm(_converse.registration_domain);   
+                        }
                     }
                     return this;
                 }
@@ -116,6 +121,7 @@
                 render: function () {
                     this.$parent.append(this.$el.html(
                         tpl_register_panel({
+                            'default_domain': _converse.registration_domain,
                             'label_domain': __("Your XMPP provider's domain name:"),
                             'label_register': __('Fetch registration form'),
                             'help_providers': __('Tip: A list of public XMPP providers is available'),
@@ -227,18 +233,36 @@
                         $domain_input.addClass('error');
                         return;
                     }
-                    $form.find('input[type=submit]').hide()
-                        .after(tpl_registration_request({
-                            cancel: __('Cancel'),
-                            info_message: __('Requesting a registration form from the XMPP server')
-                        }));
-                    $form.find('button.button-cancel').on('click', this.cancelRegistration.bind(this));
+                    $form.find('input[type=submit]').hide();
+                    this.renderRegistrationRequest(__('Cancel'));
+                    this.fetchRegistrationForm(domain);
+                },
+
+                fetchRegistrationForm: function(domain_name) {
+                    /* This is called with a domain name based on which, it fetches a
+                     * registration form from the requested domain.
+                     *
+                     * Parameters:
+                     *      (Domain name) domain_name - XMPP server domain
+                     */ 
                     this.reset({
-                        domain: Strophe.getDomainFromJid(domain),
+                        domain: Strophe.getDomainFromJid(domain_name),
                         _registering: true
                     });
                     _converse.connection.connect(this.domain, "", this.onRegistering.bind(this));
-                    return false;
+                    return false; 
+                },
+
+                renderRegistrationRequest: function(cancel_label) {
+                    var form_help = document.querySelector('.form-help');
+                    $(form_help).after(tpl_registration_request({
+                        cancel: cancel_label,
+                        info_message: _converse.__('Requesting a registration form from the XMPP server')
+                    }));
+                    if (!_converse.registration_domain) {
+                        var cancel_button = document.querySelector('button.button-cancel');
+                        cancel_button.addEventListener('click', this.cancelRegistration.bind(this));
+                    }
                 },
 
                 giveFeedback: function (message, klass) {
@@ -349,6 +373,9 @@
                         $form.append('<input type="button" class="submit" value="'+__('Return')+'"/>');
                         $form.find('input[type=button]').on('click', this.cancelRegistration.bind(this));
                     }
+                    if (_converse.registration_domain) {
+                        $form.find('input[type=button]').hide();
+                    }
                 },
 
                 reportErrors: function (stanza) {
@@ -390,6 +417,11 @@
                     if (ev && ev.preventDefault) { ev.preventDefault(); }
                     _converse.connection.reset();
                     this.render();
+                    if (_converse.registration_domain) {
+                        this.renderRegistrationRequest(__('Retry'));
+                        document.querySelector('button.button-cancel').onclick = 
+                            _.bind(this.fetchRegistrationForm, this, _converse.registration_domain); 
+                    }
                 },
 
                 submitRegistrationForm : function (ev) {
--- a/src/templates/register_panel.html
+++ b/src/templates/register_panel.html
@@ -1,7 +1,14 @@
 <form id="converse-register" class="pure-form converse-form">
     <span class="reg-feedback"></span>
     <label>{{{label_domain}}}</label>
-    <input type="text" name="domain" placeholder="{{{domain_placeholder}}}">
+    {[ if (default_domain) { ]}
+    	<label>{{{default_domain}}}</label>
+    {[ } ]}
+    {[ if (!default_domain) { ]}
+    	<input type="text" name="domain" placeholder="{{{domain_placeholder}}}">
+    {[ } ]}
     <p class="form-help">{{{help_providers}}} <a href="{{{href_providers}}}" class="url" target="_blank" rel="noopener">{{{help_providers_link}}}</a>.</p>
-    <input class="pure-button button-primary" type="submit" value="{{{label_register}}}">
+    {[ if (!default_domain) { ]}
+    	<input class="pure-button button-primary" type="submit" value="{{{label_register}}}">
+    {[ } ]}
 </form>
--- a/src/templates/registration_request.html
+++ b/src/templates/registration_request.html
@@ -1,3 +1,5 @@
 <span class="spinner login-submit"/>
 <p class="info">{{{info_message}}}</p>
-<button class="pure-button button-cancel hor_centered">{{{cancel}}}</button>
+{[ if (cancel) { ]}
+	<button class="pure-button button-cancel hor_centered">{{{cancel}}}</button>
+{[ } ]}


src/templates/registration_request_auto.html
```diff
@@ -0,0 +1,3 @@
+<span class="spinner login-submit"/>
+<p class="info">{{{info_message}}}</p>
```
jcbrand: It doesn't look to me like you need a separate template. Instead you can simply use the existing `registration_request.html` template and add an "if" statement around the button which you don't want to show.

src/converse-core.js
```diff
@@ -242,7 +242,9 @@
             credentials_url: null, // URL from where login credentials can be fetched
             csi_waiting_time: 0, // Support for XEP-0352. Seconds before client is considered idle and CSI is sent out.
             debug: false,
+            default_domain_name: 'chatme.im',
```
jcbrand: Also, rather rename it to `registration_domain`.

src/converse-register.js
```diff
@@ -72,6 +74,16 @@
...
                         this.registerpanel.render().$el.addClass('hidden');
+                        if(_converse.enable_default_domain === true) {
+                            console.log("Registering directly to domain");
+                            var $form = $('form#converse-register');
+                            $form.find('input[name=domain]').attr('disabled', 'disabled');
+                            console.log(_converse.default_domain_name);
+                            $form.find('input[name=domain]').val(_converse.default_domain_name);
+                            console.log($form.find('input[name=domain]').val());
+                            $form.trigger("submit_auto");
+                            console.log("Rendered registration form");
```
jcbrand: I don't like this way of doing things.

Instead of rendering a form and then programmatically setting values and submitting it, you can just avoid all that and fetch the registration form yourself in the code.

Something like this:

```
if(!._.isEmpty(_converse.registration_domain)) {
    chooseProvider();
}
```

Then `chooseProvider` looks basically like the last 5 lines of `onProviderChosen`. In fact, you can then also let `onProviderChosen` call `chooseProvider` and remove those 5 lines from it.

smitbose:
@jcbrand please review the modifications as asked.

docs/CHANGES.md
```diff
@@ -4,6 +4,7 @@
 
 - #585 Duplicate contact created due to JID case sensivity [saganshul]
 - #628 Fixes the bug in displaying chat status during private chat. [saganshul]
+- #704 Automatic fetching of registration form when registration_domain is set. [smitbose]
```
jcbrand: `registration_domain` should be a hyperlink like I showed in my previous comment.

docs/source/configuration.rst
```diff
@@ -483,6 +483,12 @@ JIDs with other domains are still allowed but need to be provided in full.
 To specify only one domain and disallow other domains, see the `locked_domain`_
 option.
 
+registration_domain
+-------------------
+
+Specify a domain name for which the registration form will be fetched automatically,
+without the user having to enter any XMPP server domain name.
```
jcbrand: Specify also the default value, which is an empty string.

src/converse-register.js
```diff
@@ -390,6 +416,10 @@
                     if (ev && ev.preventDefault) { ev.preventDefault(); }
                     _converse.connection.reset();
                     this.render();
+                    if(_converse.registration_domain !== '')
+                    {
```
jcbrand: Why do you start the curly bracket on a new line? Please stick to the coding convention used throughout. Also, add spaces between your `if` statements and the first left bracket `(`

src/converse-register.js
```diff
@@ -241,6 +245,25 @@
...
+                     * which loads the registration form directly
+                     */
+                    var $form = $('form#converse-register');
+                            $form.find('input[name=domain]').hide();
+                            $form.find('input[type=submit]').hide()
+                                .after(tpl_registration_request({
+                                    cancel: _converse.__('Cancel'),
+                                    info_message: _converse.__('Requesting a registration form from the XMPP server')
+                                }));
+                            $form.find('button.button-cancel').hide();
```
jcbrand: Instead of manually hiding the cancel button here, you can hide it in the template itself with an `if` statement. You pass in another boolean value to `tpl_registration_request` and then check that value in the template to decide whether the cancel button should be shown or not.

src/converse-register.js
```diff
@@ -241,6 +245,25 @@
                     return false;
                 },
 
+                chooseProvider: function (ev) {
+                    /* Method called when registration_domain is provided
+                     * which loads the registration form directly
+                     */
+                    var $form = $('form#converse-register');
+                            $form.find('input[name=domain]').hide();
```
jcbrand: 1. Fix the indentation please.

2. Instead of manually hiding the `input` elements, you can rather just replace the contents of the entire form with the template.

3. I'm in the process of removing jQuery from converse.js. I know there's still lots of jQuery code, but in we should at least not add any more jQuery stuff. Instead of using the jQuery API here, please use instead the native DOM API. This might help: http://youmightnotneedjquery.com/
Instead of using `.hide()`, you can use `el.classList.add('hidden')`.

src/converse-register.js
```diff
@@ -241,6 +245,25 @@
                     return false;
                 },
 
+                chooseProvider: function (ev) {
```
jcbrand: `chooseProvider` and `onProviderChosen` are both very similar. I think they can be combined into one method instead (keep the name `onProviderChosen`).

src/converse-register.js
```diff
@@ -349,6 +372,9 @@
                         $form.append('<input type="button" class="submit" value="'+__('Return')+'"/>');
                         $form.find('input[type=button]').on('click', this.cancelRegistration.bind(this));
                     }
+                    if(_converse.registration_domain !== '') {
+                        $form.find('input[type=button]').hide();
+                    }
```
jcbrand: Why are you hiding buttons here?

src/converse-register.js
smitbose: @jcbrand I will need another method, because in case registration_domain is mentioned by user, I'll have to invoke that function. onProviderChoosen() has been kept specifically for binding it to the `submit` event. How specifically do you want the onProviderChoosen() to be called? Firstly it requires an event as a parameter. Secondly, if I do call it separately, should I wrap codes specific to the `registration_domain` config in an if, and do the necessary? Waiting for your inputs on this...

src/converse-register.js
smitbose: This is the registration form rendered. The cancel button basically allows the end user to go back to the form where they can specify the registration domain. Since, there is no requirement to go back to that form, we don't need this button right?

smitbose:
@jcbrand made most of the changes. Only one thing, I was trying to remove most of the jQuery, except the part where one .after() function call was required. That for some reason was not working properly for javascript .insertAdjacentHTML(). I've written the alternate code in my comment. Please review. Also, please consider my explanation for writing a chooseProvider() function.

src/converse-register.js
```diff
@@ -241,6 +246,24 @@
                     return false;
                 },
 
+                chooseProvider: function () {
+                    /* Method called when registration_domain is provided
+                     * which loads the registration form directly
+                     */
+                    var form_help = document.getElementsByClassName('form-help');
```
jcbrand: Use `querySelector('.form-help')` instead. Then it will return only the element, or `null` if no element is found.

src/converse-register.js
```diff
@@ -241,6 +246,24 @@
...
+                    var reg_req = tpl_registration_request({
+                        cancel: '',
+                        info_message: _converse.__('Requesting a registration form from the XMPP server')
+                    });
+                    //form_help[0].insertAdjacentHTML('afterend', reg_req); @jcbrand not working for some reason
```
jcbrand: What exactly happens?

src/converse-register.js
jcbrand: Javascript is very forgiving when it comes to parameters passed to methods or functions. So `onProviderChosen` doesn't *have* to be called with an event as first parameter, it just currently assumes that it will. This is something that can be fixed with a bit of refactoring.

Alternatively what you can do is to have one method for the `submit` event, which then simply calls another method which then contains the common code from `chooseProvider` and `onProviderChosen`.

src/templates/registration_request.html
```diff
@@ -1,3 +1,5 @@
 <span class="spinner login-submit"/>
 <p class="info">{{{info_message}}}</p>
-<button class="pure-button button-cancel hor_centered">{{{cancel}}}</button>
+{[ if (cancel != '') { ]}
```
jcbrand: Rather do `!cancel` please, instead of checking for an empty string.

src/converse-register.js
smitbose: As far as the rendering is concerned, this should render a spinner and then a <p> tag. When using jQuery, it does exactly that, but the javascript insertAdjacentHTML counterpart renders the spinner text together, spinning the entire thing. More technically, on doing some inspect element, it shows that the javascript code renders the <p> element inside the span element. I think this is mainly because of parsing problem for the `/>` in the html string at the end of the `<span>`.

smitbose:
@jcbrand I have merged the code for chooseProvider() and onProviderChoosen(). However, the problem of replacing .after() jquery still remains. I've still kept that as a comment. Waiting for review on this one, after which I'll deal with it.

src/converse-register.js
```diff
@@ -349,6 +371,9 @@
                         $form.append('<input type="button" class="submit" value="'+__('Return')+'"/>');
                         $form.find('input[type=button]').on('click', this.cancelRegistration.bind(this));
                     }
+                    if (_converse.registration_domain !== '') {
+                        $form.find('input[type=button]').hide();
+                    }
```
jcbrand: Isn't this button now being omitted via the template? In which case this call is no longer necessary, right?

src/converse-register.js
```diff
@@ -390,6 +415,9 @@
                     if (ev && ev.preventDefault) { ev.preventDefault(); }
                     _converse.connection.reset();
                     this.render();
+                    if (_converse.registration_domain !== '') {
```
jcbrand: I think this should rather check something like `!!_converse.registration_domain` so that other falsy values also work. For example, the above code won't work properly if the user passes in `null` or `undefined` or `false`.

The same applies for the other areas where you check specifically against an empty string.

src/converse-register.js
```diff
@@ -219,20 +224,37 @@
...
+                        domain = _converse.registration_domain;
+                    }
+
+                    var reg_req = tpl_registration_request({
+                        cancel: cancel_label,
+                        info_message: _converse.__('Requesting a registration form from the XMPP server')
+                    });
+
+                    if (!_converse.registration_domain) {
+                        $form.find('input[type=submit]').hide()
```
jcbrand: Isn't this `submit` button now no longer rendered in the template if the default domain is specified?

src/templates/register_panel.html
```diff
@@ -1,7 +1,14 @@
...
+    	<label>{{{default_domain}}}</label>
+    {[ } ]}
+    {[ if (!default_domain) { ]}
+    	<input type="text" name="domain" placeholder="{{{domain_placeholder}}}">
+    {[ } ]}
     <p class="form-help">{{{help_providers}}} <a href="{{{href_providers}}}" class="url" target="_blank" rel="noopener">{{{help_providers_link}}}</a>.</p>
-    <input class="pure-button button-primary" type="submit" value="{{{label_register}}}">
+    {[ if (!default_domain) { ]}
+    	<input class="pure-button button-primary" type="submit" value="{{{label_register}}}">
+    {[ } ]}
```
jcbrand: Looks good!

src/converse-register.js
```diff
@@ -219,20 +224,37 @@
...
+                            domain = $domain_input.val();
+                        if (!domain) {
+                            $domain_input.addClass('error');
+                            return;
+                        }
+                        cancel_label = __('Cancel');
+                    }
+                    else {
+                        domain = _converse.registration_domain;
+                    }
```
jcbrand: There's a bit more refactoring that can be done here. Lines 228 to 238 can go into a separate method (we can again call it `onProviderChosen`) which is the event handler for the `submit` event. That method can then assume that `ev` is always passed in.

Then, when you call `chooseProvider`, call it always with the domain. So in `onProviderChosen` you call `chooseProvider` with the domain that you get from the form element.

 And when you call `chooseProvider` in the case where a default domain is specified, you call `chooseProvider(_converse.registration_domain)`

src/converse-register.js
smitbose: But the check here is `!_converse.registration_domain` which ensures the `submit` button is always rendered in this scope.

src/converse-register.js
smitbose: Right, that will be a better organisation for code.

src/converse-register.js
smitbose: No this cannot be done from the template. There is no template seperate template file for this. If you see the code for the function above this line, the button is rendered in the function itself, by passing the entire html string. So this is the best method to hide it here.

src/converse-register.js
smitbose: ok, I'll change it.

src/converse-register.js
jcbrand: It's hiding the submit button if there's no default domain. Isn't that wrong? In any case, I prefer these buttons being shown/hidden via the template.

I know that I'm responsible for the original code that does the above, but please don't follow my bad habits :)

src/converse-register.js
jcbrand: ok.

src/converse-register.js
smitbose: Yeah, there was an error in logic here, I agree, didn't notice that. I'll change the if condition. But for this one, I simply have no option of hiding it from a template, unless I move the code that is above this to a new template.

jcbrand: ok.

smitbose: ok, my bad, I got confused with another review. This one has a template definitely. But the hiding in onProviderChoosen() is based on the event that the submit button is clicked. How can I make such event based changes from template?

jcbrand:
Concerning the `after()` issue, would you be willing/interested in removing jQuery as a dependency for this entire plugin? We can then handle that as a separate pull request and not do any of that stuff (including replacing `after()`) here in this PR.

smitbose:
@jcbrand okay, that seems to be a nice idea.

src/converse-register.js
smitbose: No @jcbrand, I was reviewing my own changes. I just realised(probably after the much needed sleep) that there is nothing wrong in the logic. The submit button is not just hidden in this case. Its replaced with the spinner template and the cancel button.

Anyways, I'm moving on to changing the function calls to put the common code into a function.

jcbrand: ok

smitbose:
@jcbrand the build integration test doesn't seem to play well with the `!!` introduced!

smitbose:
screenshot from 2017-03-16 20-28-50
screenshot from 2017-03-16 20-29-27
screenshot from 2017-03-17 19-53-27
These are the made UI changes.

smitbose:
@jcbrand still waiting for review on this one

src/converse-register.js
```diff
@@ -227,18 +233,36 @@
...
+
+                renderRegistrationRequest: function(cancel_label) {
+                    var form_help = document.querySelector('.form-help');
+                    $(form_help).after(tpl_registration_request({
+                        cancel: cancel_label,
+                        info_message: _converse.__('Requesting a registration form from the XMPP server')
+                    }));
+                    if (!_converse.registration_domain) {
+                        var cancel_button = document.querySelector('button.button-cancel');
+                        $(cancel_button).on('click', this.cancelRegistration.bind(this));
```
jcbrand: I know you struggled with `insertAdjacentHTML` so it's ok to use `after` above until we remove jQuery completely from this plugin, but here you can already stop using jQuery.

Use `cancel_button.addEventListener('click', this.cancelRegistration.bind(this));`

src/converse-register.js
```diff
@@ -390,6 +417,14 @@
                     if (ev && ev.preventDefault) { ev.preventDefault(); }
                     _converse.connection.reset();
                     this.render();
+                    if (_converse.registration_domain) {
+                        this.renderRegistrationRequest(__('Retry'));
+                        var _register_panel = this;
+                        document.querySelector('button.button-cancel').onclick =  
+                            function () { 
+                                _register_panel.fetchRegistrationForm(_converse.registration_domain); 
+                            };
```
jcbrand: We can use partial application here instead of writing out a function.

Try this:
```
document.querySelector('button.button-cancel').onclick =  
      _.bind(this, this.fetchRegistrationForm, _converse.registration_domain);
```

jcbrand:
@smitbose two small changes still.

smitbose:
@jcbrand made the final changes, please check

jcbrand:
Merged, thanks!

PR #832, opened 2017-04-02
Fix #814: Images from URLs with query strings aren't being rendered.

### LINKED ISSUE #814: Images from URLs with query strings aren't being rendered.
Converse.js renders images from pasted URLs.

However, if the URL contains a query string, then the image doesn't get rendered. This must be because the code that checks a URL to determine whether it points to an image doesn't recognize that this is an image URL.

The solution would be to update that code to take query strings into account.

An example of such an URL which should work, but doesn't:

https://cloud.opkode.com/index.php/apps/files_sharing/ajax/publicpreview.php?x=2872&y=850&a=true&file=screen_shot_2017-03-14_at_09.06.39_720.png&t=9szwpg220warFeY&scalingup=0

novokrest:

- [x] Add a changelog entry for your change in `docs/CHANGES.md`

### DIFF
--- a/spec/chatbox.js
+++ b/spec/chatbox.js
@@ -1224,6 +1224,32 @@
                         // the page is loaded via file:///
                         return;
                     }
+                    var message = document.URL.split(window.location.pathname)[0] + "/logo/conversejs.svg";
+                    var contact_jid = mock.cur_names[0].replace(/ /g,'.').toLowerCase() + '@localhost';
+                    test_utils.openChatBoxFor(_converse, contact_jid);
+                    var view = _converse.chatboxviews.get(contact_jid);
+                    spyOn(view, 'sendMessage').andCallThrough();
+                    runs(function () {
+                        test_utils.sendMessage(view, message);
+                    });
+                    waits(500);
+                    runs(function () {
+                        expect(view.sendMessage).toHaveBeenCalled();
+                        var msg = view.$el.find('.chat-content').find('.chat-message').last().find('.chat-msg-content');
+                        expect(msg.html()).toEqual('<img src="'+message+'" class="chat-image">');
+                    });
+                }));
+
+                it("will render images from their URLs with query strings containing HTML-escaping characters", mock.initConverse(function (_converse) {
+                    test_utils.createContacts(_converse, 'current');
+                    test_utils.openControlBox();
+                    test_utils.openContactsPanel(_converse);
+
+                    if (/PhantomJS/.test(window.navigator.userAgent)) {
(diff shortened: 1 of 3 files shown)


spec/chatbox.js
```diff
@@ -1224,7 +1224,8 @@
                         // the page is loaded via file:///
                         return;
                     }
-                    var message = document.URL.split(window.location.pathname)[0] + "/logo/conversejs.svg";
+                    var message = document.URL.split(window.location.pathname)[0] + "/logo/conversejs.svg?param1=val1&param2=val2",
+                        htmlEscapedMessage = message.replace(/&/g, '&amp;');
```
jcbrand: Instead of modifying the existing check, please add an additional check for the above URL, so that both the original and the new URL are checked.

src/utils.js
```diff
@@ -44,6 +44,13 @@
         }
     };
 
+    // Unescape HTML-escaped characters (e.g. transform occurrences of '&amp;' to '&')
+    var unescapeHTML = function(htmlEscapedText) {
```
jcbrand: Please put the comment below the function declaration and put a space between `function` and the `parameter list`, I.e. `function (htmlExcapedText) {` as is the convention elsewhere in the code.

jcbrand:
Hi @novokrest, thanks for this pull request. Looking good! I just have two small change requests. please expand the existing test instead of overwriting it and also make sure that your code changes follow the existing conventions.

When in doubt, you can consult the [coding style-guide](https://conversejs.org/docs/html/style_guide.html).

novokrest:
Hi @jcbrand! 
* I change comment on added unescapeHTML() function. 
* Also I create separate test which checks rendering images from URL with query string but "my inner perfectionist" tell me that it results in some DRY violation. May be it is worth to move common part from "will render images from their URLs" and "will render images from their URLs with query strings containing HTML-escaping characters" tests? If 'yes', could you suggest the most appropriate place in code for this common part?

jcbrand:
@novokrest I agree that the extra test now violates DRY. I on purpose used the word "check" instead of "test" to try to avoid creating the impression that you should create a new `it` test.

Instead, I think the existing test should just be expanded/extended, to check the original image URL as well as the new one which contains a query string. That way you keep things DRY.

jcbrand:
@novokrest I did the merge now and decided to make the last change myself, since I want to make a new release soon.

You can see the change here:
https://github.com/jcbrand/converse.js/commit/ecdb3a12be202f0434e0922767ab123bb9caab9d

Thanks for contributing!

PR #836, opened 2017-04-04
Improve spelling

marclaporte:


jcbrand:
Apparently `publically` is gaining traction ;) http://english.stackexchange.com/questions/45136/difference-between-publicly-and-publically

marclaporte:
hehe, yes I has seen that page, but it felt better to change it anyway ;-)

Issue #847, opened 2017-04-17
Some CSS is broken in Firefox but fine in Chrome. Fix provided inside.


Semperverus:
I have been making some modifications using "Inspect Element" to fix these, and dropping them into a file called theme-fixes.css, then loading that after the converse.min.css and theme.min.css to correct for it. Basically, Firefox is either not inheriting styles properly, or chrome is doing weird stuff that @jcbrand is developing against. I've tested these fixes in both chrome and firefox, and chrome looks identical before and after.

Here are my fixes so far:
```
#conversejs #controlbox #fancy-xmpp-status-select a.change-xmpp-status-message {
    margin-right: 5px;
    margin-top: 5px;
}

#conversejs #controlbox .dropdown dd.search-xmpp {
    width: 100%;
}

.current-xmpp-contact {
    display: block !important;
}
```

This fixes the following:

- The pencil icon in the status change box floats far too much into the upper right corner

- The "Add Contact" box has a width of 0% and makes adding new contacts difficult

- Contact entries are set to "display: inline;" instead of "display: block;" sometimes, and break the list when this happens. Names will either overlap names below them or all names will be displayed side by side instead of in a vertical list.

There's one more I'm trying to fix right now with the ( i ) button next to room names not being aligned. One thing that works for Firefox but not Chrome is to set
```
#conversejs #controlbox #chatrooms dl.rooms-list dd.available-chatroom a.room-info::before {
    position: relative;
    top: 1px;
}
```

I hope this helps.

jcbrand:
Thank you. Can you please make a pull request with the first two fixes?

Concerning the 3rd one, `!important` in CSS rules is harmful and shouldn't be used. Also, it works around the problem instead of finding the underlying cause. The proper solution will be to figure out why they get `display: inline` sometimes, which is admittedly more difficult to do.

Semperverus:
I'd love to, but I unfortunately don't know all the necessary steps in Git
that I'd need to take. I know how to clone and pull, but that's about it.
If I "Git Push" will it clone it to my account here?

jcbrand:
I disabled all non-converse CSS and then tested in Firefox and couldn't reproduce the errors you mentioned. Are you sure it's not your own CSS that's interfering here?

Semperverus:
Correct.

So basically, what I've got set up is a really basic webpage. All I did was
clone your git repo, and then go to the default page and the CSS issues
start up in firefox. Everything is self-hosted, so I'm not linking to
[...]

PR #852, opened 2017-04-20
CSS fixes for firefox

Semperverus:
@jcbrand I figured out how to do pull requests (A lot easier than I thought it was gonna be). Here is what you requested.

In relation to #847

### DIFF
--- a/css/converse.css
+++ b/css/converse.css
@@ -1956,6 +1956,7 @@
       #conversejs #controlbox .dropdown dd ul li:hover {
         background-color: #DCF9F6; }
     #conversejs #controlbox .dropdown dd.search-xmpp {
+      width: 100%
       display: none; }
     #conversejs #controlbox .dropdown dd.search-xmpp ul {
       box-shadow: 1px 4px 10px 1px rgba(0, 0, 0, 0.4); }
@@ -2044,7 +2045,8 @@
       float: right;
       clear: right;
       width: 12px;
-      margin-right: 0.3em;
+      margin-right: 5px;
+      margin-top: 5px;
       color: #2A9D8F; }
   #conversejs #controlbox .controlbox-pane {
     background-color: white;


Semperverus:
As a side note, I managed to make this as a pull request in my own fork on accident, and thought I was committing to yours. Ignore the extra junk below :P

jcbrand:
Thanks. The CSS files however are generated from SASS files in the `sass` folder and not edited by hand. So the fixes need to go there.

Semperverus:
I can change it to em if you like, that's not an issue. I set it to px though because the box it's in was defined by px and not by em.

Semperverus:
Ok, I looked at the sass folder and I really don't understand what is going on in it...

I'm sorry, I want to help but this keeps getting weirder and weirder for me :(

jcbrand:
> I can change it to em if you like, that's not an issue. I set it to px though because the box it's in was defined by px and not by em.

Yes, the CSS is definitely not consistent with these things, which is unfortunate. In general however I want to move away from `px` as much as possible.

> I'm sorry, I want to help but this keeps getting weirder and weirder for me :(

I understand. I appreciate the effort you have done so far.

At the very least, I will add the 100% fix you made to the relevant Sass file.
I'm not yet convinced of the margin-top fix, because when I add that it pushes the pencil down too much (and that's without any other CSS in the page).
And then concerning the !important fix, I've already mentioned my problem with that.

CSS is hard and flaky across browsers unfortunately.

Semperverus:
A friend of mine helped me figure it out a bit. He's gonna be probably teaching me about all the git commands later (pretty sure i bothered him lol, but he does this sort of thing for a living). I'm closing this and re-submitting.

The SASS looked a lot weirder, but there were some context clues I was missing. the #conversejs and #controlbox tags correlated to specific .scss files. I found where stuff needs to go.

Semperverus:
So I just tested this issue on Firefox for Android as well, and all the same issues are there. Even the 5px by 5px fix works (and somehow manages to nail it right in the center. Perhaps 5pt by 5pt would be more technically apropriate?)

PR #853, opened 2017-04-20
Update _controlbox.scss

Semperverus:
Alright! I figured it out this time (thanks to a close friend of mine who I was talking to).

Here are the changes for [Issue 847](https://github.com/jcbrand/converse.js/issues/847).

Let me know if I did it right this time.

### DIFF
--- a/sass/_controlbox.scss
+++ b/sass/_controlbox.scss
@@ -203,6 +203,7 @@
 
             /* Custom addition for CSP */
             dd.search-xmpp {
+                width: 100%;
                 display: none;
             }
 
@@ -317,6 +318,7 @@
                 float: right;
                 clear: right;
                 width: 12px;
+                margin-top: 0.3em;
                 margin-right: 0.3em;
                 color: $link-color;
             }


jcbrand:
On a page with zero other CSS besides converse.css, when I apply your patch, then the result looks like this (chromium and Firefox on Mac):

screen shot 2017-04-21 at 16 57 18

I can only conclude that you have other CSS in your page that interferes and is causing trouble.

Semperverus:
So, it doesn't look any different on my end at all. Still renders the same. I suppose what I'm going to do is just keep my fixes locally, unless someone else chimes in here with the same issue.

jcbrand:
I haven't made the release yet. ;)

Semperverus:
Oh, oops

jcbrand:
I made release 3.0.2 now, so you can check.

Semperverus:
Ok, sorry about the wait. I got a little busy. I checked again, and it got a little worse. I've also made a dedicated page without my fixes (it's at chat.mysite.com/converse/index-withoutfix.php, useful for mobile testing).

Desktop:
untitled

Android:
android

jcbrand:
There was an issue with the icons not loading (due to having upgraded to a newer version of cleancss-cli which now minifies CSS in a different way).

I've now fixed that (although the old files might still be cached in the CDN).

Besides that, I still haven't been able to reproduce the rendering glitches you see.

Semperverus:
Like I said, I'll keep my fixes local for now. Also thank you for all the
time you've put into this.

This is why people don't use firefox anymore (I'm on it because I trust it
doesn't phone home, unlike chrome and chromium).

jcbrand:
@Semperverus I was actually able to reproduce the one effect while running tests, although it's still unclear to me under what circumstances it actually manifests.

In any case, I've now updated the HTML/CSS related to the "choose status" dropdown a bit to make it simpler (I removed some HTML) and to solve the problems you mentioned.

I also added the `100%` width to the `dd.search-xmpp` element as you suggested. Thanks for the pull request, even though it didn't get merged as is, it helped me to get to the bottom of the problem.

Semperverus:
That's awesome! Glad I could be of some help then.

PR #855, opened 2017-04-21
support opening chat in minimized mode initially

novokrest:
**Converse.js** already supports opening chat in minimized (trimmed) mode by specifying **minimized** attribute: **_converse.api.chats.open(jid, { minimized: true })**. But current implementation has incorrect behavior: if you pass **minimized: true** to **chats.open()**, then **both** normal and trimmed chatboxes will be opened. 
Proposed PR fixes this bug.

### DIFF
--- a/spec/chatbox.js
+++ b/spec/chatbox.js
@@ -122,6 +122,22 @@
                 });
             }));
 
+            it("can be opened in minimized mode initially", mock.initConverse(function(_converse) {
+                test_utils.createContacts(_converse, 'current');
+
+                var sender_jid = mock.cur_names[0].replace(/ /g,'.').toLowerCase() + '@localhost';
+                var chat = _converse.api.chats.open(sender_jid, {
+                    minimized: true
+                });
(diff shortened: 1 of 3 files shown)


spec/chatbox.js
```diff
@@ -122,6 +122,24 @@
                 });
             }));
 
+            it("can be opened in minimized mode initially", mock.initConverse(function(_converse) {
+                test_utils.createContacts(_converse, 'current');
+                test_utils.openControlBox();
+                test_utils.openContactsPanel(_converse);
```
jcbrand: Lines 127 and 128 don't appear necessary.

jcbrand:
Thanks @novokrest!

Looking good. Please just remove the `/dist`/ files from the pull request. To keep diffs smaller and readable I only generate them before making a release.

novokrest:
@jcbrand I've adjusted test in `chatbox.js` and removed `/dist` files from the pull request

PR #857, opened 2017-04-23
Discussion about Spoilers 

Delape:
Hi! :)

As it can be seen in this screenshot:

The icon is duplicated. By default is placed on the left, but for me is more natural to display it on the right.

I never used templates, but for what I have seen in the code, should this be used as templates?
https://github.com/Delape/converse.js/blob/master/src/converse-chatview.js#L899

There's one thing to be discussed regarding the implementation. Spoilers can have "hints" to indicate what is the content about.

Like this:

They are totally optional but can be really handy. 
Currently, we have:

And I thought about doing something like:

But, personally, it seems to take too much space and could be «troublesome» when writing long messages.

Thank you for your time :)


jcbrand:
Hi @Delape 

To Some initial thoughts.

1. To keep the diff as small as possible, the `dist` files shouldn't be included in the pull request. These files are only generated before a release.

The easiest way you can revert them to before is to simply copy the dist files from the master branch into your branch and then `commit` and `push`.

2. You're also not supposed to modify or include `fonticons/style.css`. Instead, to add a new font, the Sass file `sass/_fonts.scss` needs to be updated to include the new font.

3. Concerning the toolbar icon, it should be put in a template, and yes, you should be using templates ;)

It should go into the [chatroom_toolbar.html](https://github.com/jcbrand/converse.js/blob/913eeec3515a290a1e64eca4eeafd36332092e89/src/templates/chatroom_toolbar.html) template.

Toolbar icons go to the left, because the right is reserved for the encryption menu.

4. I like the idea of being able to provide a hint. Either by using the separate input field, or by requiring a second iteration, after writing the message, where (in the same textarea input) you then are asked to write the hint.

PR #864, opened 2017-04-28
Csp compat 2

mathiasertl:
As a follow-up to #695, this PR removes the last inline CSS styles so converse.js complies with strict [Content Security Policy](https://en.wikipedia.org/wiki/Content_Security_Policy) headers.

There still is one issue left with CSP, and that is one inline image, for which I will write a separate PR.

Executing `make clean dist check` seems to say that everything works fine:

```
...
Jasmine 2.5.2 finished in 32.581s
215 specs, 0 failures, 1 pending spec
```

There is a installation running this build, with CSP headers, [on jabber.at](https://jabber.at/conversejs/).

Please let me know if there is anything missing or want other improvements to the PR.

### DIFF
--- a/css/converse.css
+++ b/css/converse.css
@@ -1966,6 +1966,7 @@
       display: block;
       padding: 4px 7px 0 5px; }
   #conversejs #controlbox #select-xmpp-status {
+    display: none;
     float: right;
     margin-right: 0.5em; }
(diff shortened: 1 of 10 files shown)


jcbrand:
Looking good, thanks!

Issue #591, opened 2016-02-14
Unread message counter is reset when the chatbox is closed


rlanvin:
When a chatbox is closed, and then re-opened upon reception of a new message, the notification (visible in the browser tab's title) will show "Messages (XX)", with XX = total number of messages in the conversation, instead of "Messages (1)" as expected.

To reproduce:
1. Start chatting with another user
2. Close the chatbox completely
3. Have the other user send you a message
4. The chatbox re-opens automatically, and the browser tab's title will show the total number of messages

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/30842993-unread-message-counter-is-reset-when-the-chatbox-is-closed?utm_campaign=plugin&utm_content=tracker%2F194169&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F194169&utm_medium=issues&utm_source=github).
</bountysource-plugin>

novokrest:
@jcbrand Can I work on this issue?

novokrest:
I found that `_converse.msg_counter` variable seems to be used for showing count of unread incoming messages in the title of browser page. But there is also `num_unread` property defined on each chatbox. At the moment, `msg_counter` and `num_unread` aren't related to each other. Because _unread_|_read_ message's status is determined by visibility of chatboxes (_open_|_minimized_|_closed_), I think, it would be better to 'bind' `msg_counter` with `num_unread` properties of chats by aggregating theirs unread message counters (pseudocode): 
`msg_counter = _.reduce(chats, (sum, chat) => sum += chat.num_unread)`

Of course, it is necessary to add updating of the `num_unread` property to other places in the code, and, I think (after looking usages of `msg_counter` and `num_unread`), this refactoring can be performed quite painlessly.

As a result, this solution will fix current bug and also introduce following changes in behavior (comparing with current): if chatboxes are minimized then **total sum** of all unread messages are **always** showed even if user has already opened page (not chatboxes!) between arriving of messages.

For example, consider following use case:

1. User opens `converse.js` page
2. User opens `chatbox1` and minimize it
3. User opens `chatbox2` and minimize it
4. User leaves `converse.js` page by opening any other browser page
5. Message is received in `chatbox1`: `converse.js`-page's title is appended by `Messages(1)`
6. Message is received in `chatbox2`: `converse.js`-page's title is appended by `Messages(2)`
[...]

Issue #866, opened 2017-04-29
Add babel.js in order to support ES2015 syntax


jcbrand:
The goal would be to start using some ES2015 features such as arrow functions, `let` and `const` and perhaps [default parameters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters).

I did some exploratory work to get the infrastructure in place, which is in this commit: https://github.com/jcbrand/converse.js/commit/07b4355770bf17597dccdf3f162801ce7d818a03

What still needs to be done is to update the `Makefile` so that the builds for the `/dist` directory are made from the babel-compiled files that are placed in the `/build` directory.

We might also want to update the require.js config to refer to the `/build` directory instead of the `/src` directory.

Additionally the babel configuration needs to be updated to use [babel-preset-env](http://2ality.com/2017/02/babel-preset-env.html) and sane defaults for browser support should be chosen.

I'm thinking something like `"browsers": ["last 2 versions", "ie >= 11"]`.

The the fun starts (out of scope for this ticket), which would be to start using ES2015 features.

Any help with this would be appreciated.

PR #867, opened 2017-05-01
#591 Unread message counter is reset when the chatbox is closed 

novokrest:
@jcbrand Hi! Solution turned out to be quite simple and I decide to make separate PR for fixing just this issue.

P.S. And, as I've already mentioned, I've found some inconsistent handling of new incoming messages that cause information about unread messages is out of sync in different parts of GUI. 

Problem scenario:
1. User opens chat
2. User scrolls up so that new message will be not visible
3. User receives message: indicator of 'new unread messages' are appeared at the bottom of chatbox
4. Roster view does not show unread messages count (**bug**)
5. User minimizes chatbox: there is no indicator of unread messages count (**bug**)

Now I continue working on it and (as soon as possible) will create separate PR.

### DIFF
--- a/spec/chatbox.js
+++ b/spec/chatbox.js
@@ -1825,6 +1825,48 @@
                 expect(_converse.incrementMsgCounter).not.toHaveBeenCalled();
                 expect(_converse.msg_counter).toBe(0);
             }));
+
+            it("is incremented from zero when chatbox was closed after viewing previously received messages and the window is not focused now", mock.initConverse(function (_converse) {
+                test_utils.createContacts(_converse, 'current');
+
+                // initial state
+                expect(_converse.msg_counter).toBe(0);
+
+                var message = 'This message will always increment the message counter from zero',
+                    sender_jid = mock.cur_names[0].replace(/ /g,'.').toLowerCase() + '@localhost',
+                    msgFactory = function () { 
+                        return $msg({ 
+                            from: sender_jid,
+                            to: _converse.connection.jid,
+                            type: 'chat',
+                            id: (new Date()).getTime()
+                        })
+                        .c('body').t(message).up()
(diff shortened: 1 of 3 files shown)


spec/chatbox.js
```diff
@@ -1825,6 +1825,46 @@
...
+
+                var message = 'This message will always increment the message counter from zero',
+                    sender_jid = mock.cur_names[0].replace(/ /g,'.').toLowerCase() + '@localhost',
+                    msgFactory = () => $msg({
+                        from: sender_jid,
+                        to: _converse.connection.jid,
+                        type: 'chat',
+                        id: (new Date()).getTime()
+                    }).c('body').t(message).up()
+                      .c('active', {'xmlns': 'http://jabber.org/protocol/chatstates'}).tree();
```
jcbrand: We don't yet support ES2015 (see #866), so this needs to be rewritten as ES5.

spec/chatbox.js
```diff
@@ -1825,6 +1825,46 @@
                 expect(_converse.incrementMsgCounter).not.toHaveBeenCalled();
                 expect(_converse.msg_counter).toBe(0);
             }));
+
+            it("is incremented from zero when chatbox was closed after viewing previously received messages and the window is not focused now", mock.initConverse(function (_converse) {
+                test_utils.createContacts(_converse, 'current');
+                test_utils.openControlBox();
+                test_utils.openContactsPanel(_converse);
+
```
jcbrand: Lines 1831 and 1832 aren't needed, right?

novokrest: @jcbrand Yes, you are right, these lines can be removed

novokrest:
@jcbrand Yes, you understood correctly

spec/chatbox.js
novokrest: Done

Issue #894, opened 2017-06-14
MUC configuration icon lost with prebind


Rayzen:
Hi, I'm trying to initialize converse.js 3.0.2 with openfire 4.1.3 and everything is working quite fine except for the following error: if I set the authentication method to 'prebind' I got everything working but when the user logged enter a room he owns, the wrench icon is missing. I tried to debug and found that when converse start rendering the HTML heading of the MUC  "affiliation" attribute is null and that's why the icon is missing. Enable debug mode to logs all XMPP messages exchanged with Openfire and turns out that the the affiliation attribute is present and set to owner. Of course if set the auth method to 'login' everything works and the icon is correctly displayed.

Here my configuration:
```
(function () {
    converse.initialize({
        authentication :"prebind",
        keepalive:true,
        auto_reconnect: true,
        prebind_url:"http://bar.org/api/prebind",
        bosh_service_url: "http://bar.org:7070/http-bind",
        hide_muc_server:true,
        auto_join_on_invite:true,
        allow_contact_requests:false,
        auto_list_rooms:true,
        allow_otr:false,
        allow_registration:false,
        notify_all_room_messages:true,
        show_only_online_users:false,
        locked_domain:"foo.org",
        muc_domain:"conference.foo.org",
        muc_nickname_from_jid:true,
        show_controlbox_by_default: false,
        roster_groups: true,
        message_carbons: true,
        play_sounds: true,
        jid:"FOO@bar.org/1",
        allow_logout: false,
        storage:"session",
        debug:true
    });



})
()
;

```

the http://bar.org/api/prebind returns the following json
```
{
	jid: FOO@bar.org/1,
	sid: $openfire_sid_value,
	rid: $openfire_rid_value
}
```
and the XMPP message exchanged with brebind, while entering in room:
```

`_throttledRequestHandler called with 1 requests
_processRequest: first request has readyState of 1
_throttledRequestHandler called with 2 requests
_processRequest: first request has readyState of 1
request id 15.0 posting
request id 15.0 state changed to 1
<body rid="9243053" xmlns="http://jabber.org/protocol/httpbind" sid="7cc5voko00"><iq from="FOO@bar.org/1" to="room_test@conference.bar.org" type="get" xmlns="jabber:client" id="dba1d0b9-381a-4aea-82b2-b6d2d7c44c29:sendIQ"><query xmlns="http://jabber.org/protocol/disco#info"/></iq></body>
request id 14.1 state changed to 2
[...]

PR #895, opened 2017-06-14
Converse ryzen 894

Rayzen:
As discussed in Issue #894 the methosSaveAffiliationAndRole compared strictly the bare jid bound to connection and the bare jid retrieved from XMPP presence message.
Fixed by converting both bare jid to lowe case.


jcbrand:
Thanks!

PR #911, opened 2017-08-25
Fix regarding ChatRoomView.getDefaultNickName() and muc_nickname_from_jid

lduburas:
Allows plugins to override nickname when muc_nickname_from_jid is true.


jcbrand:
Thanks!

PR #948, opened 2017-11-08
IM is not a french word / « ouvrir » means « to open »

Reventl0v:
Simple locale fix, I guess I do not have to provide test and such.

Replacing « IM » with what is used in the rest of the file for « chat », and fixing a typo on « open » (which has to be translated by « ouvert », not « ouvrir », which means « to open ».


jcbrand:
Thanks!

PR #955, opened 2017-11-16
Changed connection manager URL

taylorsabell:
Changed the documentation to have the correct connection manager URL. If you try to use the old one, you'll get runtime JavaScript errors.


jcbrand:
Thanks!

PR #960, opened 2017-12-04
Fixed error building DOM toggle_chats.html span.unread-message-count class attribute

bolste:
toggle_chats.html template was broken in the construction of span.unread-message-count class attribute.
A missing `"` was causing the making of the DOM like so: 
`<span class="unread-message-count href=" #">...</span>`
Please note the `href=` class and the absence of the `href` attribute (see attachment).

image


jcbrand:
Thank you!

PR #961, opened 2017-12-05
Option to use RVM for ruby

jjrh:


Makefile
```diff
@@ -1,3 +1,4 @@
+SHELL := /bin/bash --login
```
jcbrand: Looks like this variable isn't being used. Why did you add it?

jjrh: I believe that is used when commands are run - otherwise, at least on my machine sh is used which causes some issues mostly with RVM.

jjrh: And yes, RVM is like virtualenv. On my machine the system version of ruby is pretty ancient so some things break.

PR #963, opened 2017-12-05
added ignore files for css and dist

jjrh:


jcbrand:
Thanks, but these files are not ignored on purpose (e.g. to see whether they have been built).

If you want them ignored, you can add them to a `.gitignore_global` file in your home directory.

jjrh:
Ah fair enough.

PR #977, opened 2017-12-29
Send emojis as unicode

ChaosKid42:
I would like converse.js to send emojis in unicode instead of shortnames. This way emojis are rendered nicely in other applications such as Conversations app.


jcbrand:
Thanks!

PR #1001, opened 2018-01-26
LibreJS support

gnunicode:
Makes changes for converse.js to be usable with the LibreJS plugin. Link to the LibreJS Web Labels HTML file was added to index.html but not yet to other HTML files.


docs/CHANGES.md
```diff
@@ -0,0 +1 @@
+- Add LibreJS support
```
linkmauve: This should be added to the CHANGES.md file at the root of the repository.

demo/anonymous.html
```diff
@@ -50,6 +50,9 @@ <h1 class="brand-heading"><i class="icon-conversejs"></i>Converse.js</h1>
 </body>
 
 <script>
+        /* This Source Code Form is subject to the terms of the Mozilla Public
+         * License, v. 2.0. If a copy of the MPL was not distributed with this
+         * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
```
linkmauve: I think this should be public domain or CC-0 or something similar, I wouldn’t be comfortable with putting that kind of license on a “configuration file”.  Same for other occurences of `converse.initialize()`.

jslicenses-converse.html
```diff
@@ -0,0 +1,63 @@
...
+    <body>
+        <table id="jslicense-labels1">
+            <tr>
+                <td>
+                    <a href="https://cdn.conversejs.org/3.3.1/dist/converse.min.js">converse.min.js</a>
+                </td>
+                <td>
+                    <a href="https://www.mozilla.org/en-US/MPL/2.0/">MPL-2.0</a>
+                    <br />
+                    <a href="magnet:?xt=urn:btih:e95b018ef3580986a04669f1b5879592219e2a7a&dn=public-domain.txt">Public-Domain</a>
```
linkmauve: `&amp;` in @href, instead of `&`.

jslicenses-inverse.html
```diff
@@ -0,0 +1,63 @@
...
+    <body>
+        <table id="jslicense-labels1">
+            <tr>
+                <td>
+                    <a href="https://cdn.conversejs.org/3.3.1/dist/converse.min.js">converse.min.js</a>
+                </td>
+                <td>
+                    <a href="https://www.mozilla.org/en-US/MPL/2.0/">MPL-2.0</a>
+                    <br />
+                    <a href="magnet:?xt=urn:btih:e95b018ef3580986a04669f1b5879592219e2a7a&dn=public-domain.txt">Public-Domain</a>
```
linkmauve: Same here, `&amp;`.

src/analytics.js
```diff
@@ -0,0 +1,11 @@
...
+_paq.push(['trackPageView']);
+_paq.push(['enableLinkTracking']);
+(function() {
+  var u="//stats.opkode.com/";
+  _paq.push(['setTrackerUrl', u+'piwik.php']);
+  _paq.push(['setSiteId', '1']);
+  var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+  g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'piwik.js'; s.parentNode.insertBefore(g,s);
+})();
+
```
linkmauve: Why are you adding such a script?

src/inverse-analytics.js
```diff
@@ -0,0 +1,11 @@
...
+  /* tracker methods like "setCustomDimension" should be called before "trackPageView" */
+  _paq.push(['trackPageView']);
+  _paq.push(['enableLinkTracking']);
+  (function() {
+    var u="//stats.opkode.com/";
+    _paq.push(['setTrackerUrl', u+'piwik.php']);
+    _paq.push(['setSiteId', '5']);
+    var d=document, g=d.createElement('script'), s=d.getElementsByTagName('script')[0];
+    g.type='text/javascript'; g.async=true; g.defer=true; g.src=u+'piwik.js'; s.parentNode.insertBefore(g,s);
+  })();
```
linkmauve: Same question.

demo/anonymous.html
gnunicode: Good point. I'll put the Unlicense on these as it's short enough.

docs/CHANGES.md
gnunicode: Done

src/analytics.js
gnunicode: These aren't mine - they're from here: https://conversejs.org/analytics.js and https://inverse.chat/inverse-analytics.js. Adding them to the source tree means it's clear they aren't proprietary.

jslicenses-converse.html
gnunicode: Changed those. It appears to be a mistake in the LibreJS documentation: https://www.gnu.org/software/librejs/manual/librejs.html#Detected-Free-Licenses

gnunicode: It seems to be the format for magnet links to have &dn instead of &amp;dn - I'm changing it back just in case LibreJS doesn't detect it as a valid license.

gnunicode: No you're right, my bad

src/analytics.js
jcbrand: @gnunicode I don't think I want to add these to the repo. What then happens is that people have these scripts included in their self-hosted setups and then my analytics server gets (incorrectly) called by them.

Is there another way to do this?

jcbrand: Other than that, thanks for the effort!

gnunicode: I've removed the scripts. What could be useful is to add a comment to the analytics scripts stating the license (in the HTML file it says the analytics files are MPL-2.0 so a comment that states they are MPL'ed could be helpful). It's not needed for LibreJS to work as I've already stated the license of those files in the HTML files, but an analytics script with no license information on it can look like it's proprietary even if LibreJS lets it through.

jcbrand:
Hi @gnunicode, doesn't look like the licenses-inverse.html file is actually linked to from anywhere, is it really needed? Alternatively (if it's needed), could we perhaps have only one licenses file for both the overlayed and fullscreen pages?

gnunicode:
Yeah it's possible to have just one license file - I'll combine the two files. I haven't linked anything on the fullscreen page yet as I wasn't sure where it would be best to put it on the page, as the link would have to be there constantly, though I guess it doesn't have to be visible

jcbrand:
Thanks for your work on this @gnunicode

jcbrand:
@gnunicode After merging, I took a closer look and noticed that the licenses page is not complete.

There are lots of other dependencies in converse.min.js which you don't list.

Here's a list of the files that you missed:

```
node_modules/otr/build/dep/crypto.js
node_modules/otr/build/dep/salsa20.js
node_modules/otr/build/dep/eventemitter.js
node_modules/otr/build/otr.js
node_modules/strophe.js/strophe.js
node_modules/strophejs-plugin-vcard/strophe.vcard.js
node_modules/strophejs-plugin-disco/strophe.disco.js
node_modules/strophejs-plugin-rsm/strophe.rsm.js
node_modules/strophejs-plugin-ping/strophe.ping.js
node_modules/snabbdom/dist/snabbdom.js
node_modules/snabbdom/dist/snabbdom-attributes.js
node_modules/snabbdom/dist/snabbdom-class.js
node_modules/snabbdom/dist/snabbdom-dataset.js
node_modules/snabbdom/dist/snabbdom-eventlisteners.js
node_modules/snabbdom/dist/snabbdom-props.js
node_modules/snabbdom/dist/snabbdom-style.js
node_modules/lodash/lodash.js
node_modules/backbone/backbone.js
node_modules/backbone.nativeview/backbone.nativeview.js
node_modules/backbone.browserStorage/backbone.browserStorage.js
node_modules/backbone.overview/dist/backbone.overview.js
node_modules/backbone.overview/dist/backbone.orderedlistview.js
node_modules/backbone.vdomview/dist/backbone.vdomview.js
node_modules/awesomplete-avoid-xss/awesomplete.js
node_modules/moment/min/moment-with-locales.js
```

jcbrand:
In order to get these files, you need to run `make dev` inside the repo, then they'll be fetched via NPM.

PR #1013, opened 2018-02-06
improve @title on roster item

linkmauve:


src/converse-rosterview.js
```diff
@@ -358,9 +358,8 @@
                     this.el.innerHTML = tpl_roster_item(
                         _.extend(item.toJSON(), {
                             'desc_status': STATUSES[item.get('chat_status')||'offline'],
-                            'desc_chat': __('Click to chat with this contact'),
+                            'desc_chat': __('Click to chat with %1$s (JID: %2$s)', item.get('fullname'), item.get('jid')),
```
jcbrand: I have to double check but I think`this.get('fullname')` is the same as `this.get('jid')` when we don't have a fullname, in which case this tooltip will have the JID twice.

jcbrand: Although I see the original text could also have had it twice. So I'll merge.

PR #1029, opened 2018-03-02
send emojis as unicode also in MUCs

ChaosKid42:
I recognized that https://github.com/jcbrand/converse.js/pull/977 does not work in MUCs. This PR fixes it.


jcbrand:
Thanks!

PR #1032, opened 2018-03-07
Move some MUC logic from ChatRoomView to ChatRoom

cmrd-senya:
In this commit I moved functions which doesn't belong to UI from `src/converse-muc-views.js` to `src/converse-muc.js`, from `ChatRoomView` class to `ChatRoom` class.

Some of the functions which contain mixed behavior (both UI and non-UI) I copied over and removed UI references (and left original functions in-place). This was enough for me to use MUC in a headless mode. However to make it in a clean way, ChatRoomView functions should be changed to reuse the behavior from ChatRoom functions. This requires some efforts to redesign the code which I can't make right now.

Here is a list of duplicate functions:
- saveConfiguration
- onOwnChatRoomPresence
- onChatRoomPresence
- onChatRoomMessage
- handleMUCMessage

### DIFF
--- a/src/converse-muc-views.js
+++ b/src/converse-muc-views.js
@@ -11,7 +11,6 @@
 (function (root, factory) {
     define([
         "converse-core",
-        "emojione",
         "tpl!add_chatroom_modal",
         "tpl!chatarea",
         "tpl!chatroom",
@@ -35,11 +34,12 @@
         "tpl!rooms_results",
         "tpl!spinner",
         "awesomplete",
-        "converse-chatview"
+        "converse-chatview",
+        "backbone.orderedlistview",
+        "backbone.vdomview"
     ], factory);
 }(this, function (
     converse,
-    emojione,
     tpl_add_chatroom_modal,
     tpl_chatarea,
     tpl_chatroom,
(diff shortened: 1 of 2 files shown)


jcbrand:
@cmrd-senya  I've moved some methods out yesterday, not the ones you're particularly interested in, but it's a work in progress :)

cmrd-senya:
Cool, thanks :)

jcbrand:
@cmrd-senya I've now moved the rest of the methods from `converse-muc-views` to `converse-muc`.

You're invited to test things out and to drop your fork :)

Some refactoring was required, although I don't expect them to affect your views much.

I moved registration of the presence and message stanza handlers to `converse-muc`. I don't know whether you registered these in your view or not.

I've come to the conclusion that excessive stanza processing in a view is a code smell. This should be done in the model/collection, and I've tried to do so where possible and time allowed.
For example, join/leave messages are now shown based on whether occupants are created/destroyed, and not by manually parsing the presence stanzas in the view.

There's still some stanza processing being done in ChatRoomView however. Eventually this will be refactored out I think. The stanza processing is generally done in order to parse message and presence stanzas to see whether we should show various notification messages (like nickname changes). Instead of parsing the stanzas, I think we should be creating special notification message objects in `converse-muc` which are then rendered in the views. By doing this we retain these notifications upon page reload (because these objects persist via sessionStorage) and we don't have to do the stanza processing in the view. However, I've run out of time to do this for now.

PR #1050, opened 2018-03-30
Make wikisuite.org link point to HTTPS

linkmauve:
They don’t have an automated redirection, nor HSTS preload, so it is better to directly point to the HTTPS version of their website.


jcbrand:
Thanks, I had to do some history rewriting due to weblate merging issues. I'll make the change manually.

PR #1051, opened 2018-03-30
Make two versions of the logo, one transparent, and one white

linkmauve:
This series adds back the transparent version of the logo, in addition to the one filled white.

In order to avoid polluting the logo directory with many pre-rendered versions of it, I added a Makefile rule to generate the PNG files automatically from the SVGs, using [Inkscape](https://inkscape.org/) and [oxipng](https://github.com/shssoichiro/oxipng), which can be called with `make logo`.

There is also an unrelated fix for `make help`.


jcbrand:
My history rewriting so that I could cleanly merge in the translations has now caused problems here.

Could you perhaps create a new PR with only your changes?

I guess there are multiple ways to do this.

For example, I would do something like this:

```
git checkout -b master-backup
git checkout master
git reset --hard HEAD~50
git fetch --all
git merge remotes/conversejs/master // Or whatever the remote is called for you
// then cherry pick the particular commits from master-backup branch
```

I didn't test the above, but something like that would work I believe.

linkmauve:
There, fixed.

jcbrand:
Thanks!

Issue #1017, opened 2018-02-12
Minified builds do not work in some circumstances


richard-fairthorne:
Minified builds found on CDN's do not work. They produce errors related to emojione.

## example

```
<html>
<head>
  <link rel="stylesheet" type="text/css" media="screen" href="https://cdn.conversejs.org/3.3.2/css/converse.min.css">
  <script src="https://cdn.conversejs.org/3.3.2/dist/converse.min.js"></script>
</head>
<body>
  <script>
      converse.initialize({
          bosh_service_url: 'https://conversejs.org/http-bind/', // Please use this connection manager only for testing purposes
          show_controlbox_by_default: true
      });
  </script>
</body>
</html>
```

## result in chrome

```
[...]

PR #1053, opened 2018-03-30
Specify utf-8 charset in documentation so minified builds will work across platforms

richard-fairthorne:
Documentation fix makes minified versions work across platforms - Addresses #1017


jcbrand:
Super, thanks!

PR #1054, opened 2018-03-31
replace geoURIs (e.g. from Convesations) by links to openstreetmap

ChaosKid42:
This will replace geoURIs (e.g. from Convesations) by links to openstreetmap. Will also work the other way round.

If you like to use other websites instead of openstreetmap, this can be configured.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,10 @@
 
 The UI is now based on Bootstrap4 and Flexbox is used extensively.
 
+## New Features
+
+- geo-URIs (e.g. from Conversations) are now replaced by links to openstreetmap (works in reverse also)
+
 ## Configuration changes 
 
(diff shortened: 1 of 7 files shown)


jcbrand:
This is great, thanks!

Please also add a test case for this.

Your test will probably look very similar to the one that checks that images URLs are rendered inline.

See here:
https://github.com/conversejs/converse.js/blob/44a2f7807c9f85d5b5b2f85b81c892908b882c09/spec/chatbox.js#L1556

To run the tests you can type `make check` or you can run them in the browser (better for debugging) by typing `make serve` and then going to localhost:8000/tests

ChaosKid42:
Ok. I will try :-)

ChaosKid42:
Is this test-case sufficient? Never done something like this before :-)

jcbrand:
Looks good, thanks!

PR #1060, opened 2018-04-04
XEP-0363 HttpFileUpload Implementation

worlword:
This adds httpfileUpload (XEP-0363) to converse.js. It uses xep-0030 Discovery to discover the xmpp-server-upload-sertvice. 
The File-Upload only works if CORS is enabled on the xmpp-server.
Furthermore it sends XEP-0066 "out-of-band-data" messages to the chat-partners when a file is send. This allows other clients (like conversations) to interpret it as a file and display it correctly. 
There is also support for mp3- and mp4-Files now. Those files will be displayed with the appropriate-player in the chat instead of a download-link. For the moment I only added those two file-categories because IE11 only supports those. But it should be easy to support more files.


jcbrand:
Thank you!

I'll only be able to look at this more closely in a week and a half once I'm back to work.

src/converse-httpFileUpload.js
```diff
@@ -0,0 +1,269 @@
...
+                },
+                
+                /**
+                * Discover upload service for http upload.
+                *
+                */
+                discoverUploadService () {
+                    var self = this;
+                    console.log('discover http upload service');
+                    connection.disco.items(domain, null, function(items) {
```
jcbrand: Converse.js has it's own, higher-level and Promise-based, service discovery API, exposed in `src/converse-disco.js`, and we should rather use that than the one from Strophe.

By using Converse's API, we don't need this `discoverUploadService` method at all, since that happens automatically.

Instead, we can simply call 

```
_converse.api.disco.supports(Strophe.NS.HTTPUPLOAD, _converse.domain)
    .then((result) => {
        // Your code here
    });
```

whenever we want to know whether the service is available.

See this example in the `http-file-upload` branch I've been working on:
https://github.com/conversejs/converse.js/blob/720ae390eeb1c743308c94f0ef5e1e476f76abcf/src/converse-http-file-upload.js#L40

src/converse-httpFileUpload.js
```diff
@@ -0,0 +1,269 @@
...
+                */
+                initFiletransfer () {
+                    connection = _converse.connection;
+                    domain = _converse.connection.domain; 
+
+                    if (httpUploadOption && requestSlotUrl != undefined) {
+                        ready = true;
+                        return;
+                    }
+                    this.discoverUploadService();
```
jcbrand: I see this method gets called when the file upload button is clicked, but only in this method is the check made to see whether the server at all supports file upload.

What we instead want to do is we want to know whether file upload is supported before we at all render the file upload button in the toolbar.

I therefore think that with proper refactoring the `initFileTransfer` method can be removed entirely.

src/converse-httpFileUpload.js
```diff
@@ -0,0 +1,269 @@
...
+                    });
+                },
+                
+                /**
+                 * Query item for upload service.
+                 *
+                 * @param {String} jid of the logged-in user
+                 * @param {Function} cb Callback on success
+                */
+                queryItemForUploadService (jid) {
```
jcbrand: In light of the above two comments, this method also looks unnecessary.

src/converse-httpFileUpload.js
```diff
@@ -0,0 +1,269 @@
...
+
+                /**
+                 * Saves the file the user has picked.
+                 * 
+                 * @param {*} file the name of the file the user has picked.
+                 * @param {*} chatBox the chatbox from which the user initiated the file-upload
+                 */
+                setFile (file1, chatBox1){
+                    file = file1;
+                    chatBox = chatBox1;
```
jcbrand: The setting of two global variables (global to the plugin) looks like a definite [code smell](https://en.wikipedia.org/wiki/Code_smell) to me.

src/converse-httpFileUpload.js
```diff
@@ -0,0 +1,269 @@
...
+                    this.sendFile();
+                },
+                
+                /**
+                * Upload file.
+                * Waits till the Upload-Service is discovered and till the user has picked a file.
+                *
+                */
+                sendFile () {
+                    var self = this;
```
jcbrand: Inasmuch this method is needed, it should go onto the `_converse.ChatBox` model.

That said, given what I've seen above, I suspect that a rewrite of the rest of the methods in this file might be necessary.

src/converse-muc-views.js
```diff
@@ -82,6 +82,8 @@
         'unmoderated': 'moderated'
     };
 
+    Strophe.addNamespace('OUTOFBAND', 'jabber:x:oob');
```
jcbrand: This namespace is declared twice, we might as well rather declare it once in `converse-core.js` or perhaps in `converse-chatboxes.js`.

src/converse-muc-views.js
```diff
@@ -693,6 +697,17 @@
...
+                    const msgid = _converse.connection.getUniqueId();
+                    const stanza = $msg({
+                        'from': _converse.connection.jid,
+                        'to': this.model.get('jid'),
+                        'type': 'groupchat',
+                        'id': msgid
+                    }).c("body").t(text).up()
+                      .c("x", {'xmlns': Strophe.NS.OUTOFBAND}).c('url').t(text).up();
+                     _converse.connection.send(stanza);
+                },
```
jcbrand: This should to onto the `_converse.ChatRoom` model in `converse-muc.js` for the same reasons as mentioned above.

src/converse-httpFileUpload.js
```diff
@@ -0,0 +1,269 @@
...
+                        }
+                    });
+                },
+
+                /**
+                * Request upload slot from xmpp-server
+                *
+                * @param  {File} file the file the user picked
+                * @param  {Function} cb Callback after finished request
+                */
```
jcbrand: These docstrings don't conform to how docstrings are done in this project. In general the styleguide in this file doesn't conform to the [Converse.js Code Style Guide](https://conversejs.org/docs/html/style_guide.html).

worlword:
Hey JCBrand,

I updated the pull request. Sadly i could not figure out how to use the converse-discovery service to dynamically add the button for file upload. The tests still failed for me.
Other than that I hope you can use the changes that I made. All your points from the previous review should have been implemented.

PR #1086, opened 2018-04-26
add another server which supports websockets

midzer:
Hi,

just a minor extension to the docs.

Cheers and have a nice week
midzer


jcbrand:
Thanks!

PR #1099, opened 2018-05-10
fix comparator in case nick is undefined

ChaosKid42:
I found that 'nick' is not always defined (e.g. if a member is currently _absent_ from a room he has no nick). To prevent an exception in that case I suggest to use the user's jid for comparing instead.

### DIFF
--- a/src/converse-muc.js
+++ b/src/converse-muc.js
@@ -1006,8 +1006,18 @@
                     const role1 = occupant1.get('role') || 'none';
                     const role2 = occupant2.get('role') || 'none';
                     if (MUC_ROLE_WEIGHTS[role1] === MUC_ROLE_WEIGHTS[role2]) {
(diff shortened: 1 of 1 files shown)


jcbrand:
> if a member is currently from a room he has no nick

What do you mean by this? If a user is  currently in a room, then he should by definition have a nick. 

Perhaps this nick is not returned in the member list, but we should have received it with the initial presence. Perhaps the issue is that the initial presence is received *after* the member-list, which causes this problem. Is that what you're referring to?

ChaosKid42:
Sorry the word "absent" was missing.

jcbrand:
Thanks for the explanation and the PR @ChaosKid42 

I've made a fix based on your code, but which relies on the `getDisplayName` convention and which is a bit more concise.

Issue #1039, opened 2018-03-15
Multi-option data form elements not saved correctly


jcbrand:
When configuring an OpenFire room, there is a multi-option widget for setting who's presences should be broadcasted.

Converse.js for some reason don't show the right values selected (should be all, but converse.js only shows "moderators") and doesn't properly save the right values if multiple values are selected.

PR #1101, opened 2018-05-12
fix selected values in list-multi form fields

ChaosKid42:
Fixed a bug which caused the list-multi form field e.g. for muc#roomconfig_presencebroadcast not being read and written correctly.


jcbrand:
Thanks! AFAIK this fixes #1039

PR #1102, opened 2018-05-13
Fix #1101 fo IE11

ChaosKid42:
This PR changes my previous PR https://github.com/conversejs/converse.js/pull/1101. Otherwise it won't work in IE11.

### DIFF
--- a/src/utils/form.js
+++ b/src/utils/form.js
@@ -98,7 +98,7 @@
                         return tpl_select_option({
                             'value': value,
                             'label': option.getAttribute('label'),
-                            'selected': values.includes(value),
+                            'selected': _.includes(values, value),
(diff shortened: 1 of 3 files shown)


src/utils/form.js
```diff
@@ -98,7 +98,7 @@
                         return tpl_select_option({
                             'value': value,
                             'label': option.getAttribute('label'),
-                            'selected': values.includes(value),
```
jcbrand: You can use `_.includes` here instead. 

Also, I forgot to ask in the previous PR, please add en entry for the original fix in `CHANGES.md`.

Thanks

ChaosKid42:
Right. Thanks for pointing me to Lodash. Fixed now.

jcbrand:
Thanks!

PR #1103, opened 2018-05-16
remove "?" from user-search-query

ChaosKid42:
I'd like to remove the trailing "?" from the call to `xhr.open` in src/converse-rosterview.js when using `xhr_user_search_url` because of increased flexibility.

You might for example add more parameters to the API call (e.g. for supplying an API_KEY or similar). In this case, the already included "?" needs to be removed from the code.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -1587,3 +1587,18 @@ This is the URL to which an XHR GET request will be made to fetch user data from
 The query string will be included in the request with ``q`` as its key.
 
 The data returned must be a JSON encoded list of user JIDs.
+
+.. note::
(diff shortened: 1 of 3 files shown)


docs/source/configuration.rst
```diff
@@ -1587,3 +1587,9 @@ This is the URL to which an XHR GET request will be made to fetch user data from
 The query string will be included in the request with ``q`` as its key.
 
 The data returned must be a JSON encoded list of user JIDs.
+
```
jcbrand: Please mention here that the `?` is required. You can use the `.. note::` directive.

jcbrand:
Thanks!

PR #1112, opened 2018-06-11
Fix build targets in Makefile

Ppjet6:
The 'dist/converse-no-dependencies-es5.js' rule is called
'dist/converse-no-dependencies-es2015.js'.

---

Do you still want a commit in `CHANGES.md`? This is just a quick fix in between a release :)


jcbrand:
Thanks!

PR #1122, opened 2018-07-06
Remove position:relative in embedded CSS

linkmauve:
This was rendering the content outside of the viewport, rendering Converse.js unusable.


jcbrand:
Sorry, but this breaks the `demo/embedded.html` example. I'll have to revert.

With `position: relative`:

screen shot 2018-07-16 at 01 19 10

Without `position: relative`:
screen shot 2018-07-16 at 01 19 34

PR #1132, opened 2018-07-14
Remove @locked_domain if it has been provided in the JID

### LINKED ISSUE #917: locked_domain should not disallow domain-part in username.
If `locked_domain` is set, the users cannot authenticate against another XMPP domain than the one that's configured in that setting.

Currently (inVerse 3.2.1), authentication will fail when `locked_domain` is set, and the username is formatted as `user@locked_domain`. That seems overly cautious (and might confuse people).

When `locked_domain` is set, inVerse should allow usernames both in the format `username` and `username@locked_domain`, just like when this option is not defined. Obviously, if the provided domain part does not match the `locked_domain` value, users must not be allowed to log in.

linkmauve:
Fixes #917.

### DIFF
--- a/src/converse-controlbox.js
+++ b/src/converse-controlbox.js
@@ -471,7 +471,11 @@
 
                     let jid = form_data.get('jid');
                     if (_converse.locked_domain) {
-                        jid = Strophe.escapeNode(jid) + '@' + _converse.locked_domain;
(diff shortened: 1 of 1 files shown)


linkmauve:
This is now fixed.

jcbrand:
Thanks!

PR #1136, opened 2018-07-20
Fix some more links to use HTTPS.

linkmauve:
All of them have been tested manually to see they still point to the same resource.


jcbrand:
Thanks!

PR #1163, opened 2018-08-14
Improve the moderated and not moderated explanations

linkmauve:
This now actually explains what it does.


jcbrand:
Thanks @linkmauve, please also fix the failing test:

https://github.com/conversejs/converse.js/blob/master/spec/roomslist.js#L165

jcbrand:
Look at the test I linked to, it expects the output to match hardcoded text in the test itself.

That text needs to be updated as well.

Issue #1131, opened 2018-07-14
Converse does not work on xhtml pages


Lukc:
Steps to reproduce:
- take any converse.js example
- change its extension from .html to .xhtml
- enjoy XML parsing errors

jcbrand:
Why should this project care about supporting xhtml?

Lukc:
Well… I wanted to use converse.js and my website is xhtml. Having to switch to html for converse.js alone would be annoying.

jcbrand:
If you're willing to make the necessary changes for xhtml support, and to make a pull request, I'll happily review and merge. Until then, I'll close this ticket, since this is not something that I'll work on.

PR #1166, opened 2018-08-17
Fix a wrongly closing <p/> into </p>.

linkmauve:
It was creating an empty element in the DOM.


linkmauve:
#1131 is also related, it would help to find these issues immediately rather than while diving into the DOM.

jcbrand:
Thanks @linkmauve, I fixed it locally due to the conflict.

PR #1167, opened 2018-08-17
Make Converse look more familiar to some users.

linkmauve:
This is totally not inspired by https://discordapp.com/ :-° Hopefully
copyright doesn’t apply to a few CSS values, but this should be checked
before merging this PR.

Also I didn’t take much advantage of SCSS variables, so there is a lot
of duplication for colour values and such, this should be fixed before
this can be merged.

There is also no theme support in Converse yet, we may want to fix that
for people who prefer the old look.

Here is how it currently looks:
converse

### DIFF
--- a/css/converse.css
+++ b/css/converse.css
@@ -9306,6 +9306,7 @@ readers do not read off random characters that represent icons */
   --message-receipt-color: #3AA569;
   --save-button-color: #3AA569;
   --chat-textarea-color: #666;
+  --chat-textarea-background-color: white;
   --chat-textarea-height: 60px;
   --send-button-height: 27px;
   --send-button-margin: 3px;
@@ -9315,6 +9316,9 @@ readers do not read off random characters that represent icons */
   --chat-head-color: #3AA569;
   --chat-head-color-lighten-50-percent: #e7f7ee;
   --chat-head-text-color: white;
+  --chat-correcting-color: var(--chat-head-color-lighten-50-percent);
+  --chat-topic-display: block;
+  --chat-info-display: block;
   --highlight-color: #DCF9F6;
   --primary-color: #578EA9;
   --primary-color-dark:  #397491;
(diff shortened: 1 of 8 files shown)


Nyco:
Wow, thanks a lot for this @linkmauve ! ;-)

Regarding IP/copyrights, if there is a real risk, can we change what we estimate might hurt?

I have the feeling (hopefully neutral) that it looks better! ;-) My opinion is: it definitely looks better!

Another opinion: we have to be careful with "too grey-ish" colors, we might want to present a more colourful look.

Another personal observation, it seems to me that the world separates in two, light vs dark modes... And be careful, lovers/hater seem very opinioniated ;-)

Can you please post a screenshot in mobile mode?

jcbrand:
Yes, thanks @linkmauve 

Quite a number of people have spoken positively about this change.

> There is also no theme support in Converse yet, we may want to fix that for people who prefer the old look.

Inasmuch as you can edit the variables in `_converse.scss` and generate a new CSS file, there is theme support. The challenge is to make this simpler, more user-friendly and exposing this to end-users.

One way would be to use [CSS variables](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_variables) instead of Sass. That would mean dropping support for IE11.

The user could then choose a theme themselves, and have Converse automatically update. 
With Sass this is not possible because there's a compilation step necessary.

> Also I didn’t take much advantage of SCSS variables, so there is a lot of duplication for colour values and such, this should be fixed before this can be merged.

I consider this a blocker to merging this PR. The current Sass variables might not be enough, or used properly, but the solution is to fix them, not to ignore them.

Ideally this pull request would just be another variables file, which specifies a different theme.
That might not be possible in this case (i.e. other files might also have to be modified), but that's the goal we should be working towards IMO.


Concerning what @Nyco wrote:

> Another opinion: we have to be careful with "too grey-ish" colors, we might want to present a more colourful look.

Depends on the audience. I too prefer more colour, but some people like things darker.
That's another reason why better theme support would be helpful, so that people can choose for themselves how it should look.

---

Lastly, I have asked a designer to come up with a new logo for Converse and in the process a new colour scheme.

I expect that a new theme might come from his work, and that this would become the default theme.

However, with support for multiple themes, we can still accommodate this PR and any other themes people wish to submit.

Nyco:
> That would mean dropping support for IE11.

Good riddance, it is at 2.64% market share a the time of this writing.
https://caniuse.com/usage-table

linkmauve:
This work was mostly a proof of concept (hence the WIP flag) because it was the most common criticism I heard about Converse after deploying it here and there, it is definitely not intended to be merged as is, but I’m already using it and my users generally prefer it to the current one.

Ignoring SCSS variables was only a way for me to get this done easily, if I do put more work into this the first thing to do will be to extract every place where I used the same colour and make that a variable. I could also try to copy sass/_variables.scss into another file and call it a theme.

As for the theme itself, I don’t especially like it either, I just chose Discord as a model because I’ve heard a lot of gamers use it, one of the communities I helped migrate to XMPP was about game development, and I thought some familiarity would help. Note Nÿco that I didn’t go to the very end of having a clone, the original has support for dark/light themes, a separator between each message and not just different days, etc.

I don’t own a mobile device, so at best I could screenshot Firefox’s devtools in responsive mode, instead you can try JabberFR’s deployment at https://chat.jabberfr.org/ (click on « Rejoindre », “join” in French, for a room you want to test).

I probably won’t work on reworking both themes to use the same SCSS code and only change variables in the few next weeks, but if there is interest in getting this merged before your designer comes up with something I can definitely priorise it more.

linkmauve:
This has been rebased on top of #1268, using custom properties instead of hardcoding any value. The two themes can now be toggled just by adding or removing the `theme-discord` class on `#conversejs`. I’m not too sure if this selector is correct because it made me add an `!important` on every CSS rule, but as a WIP this should work much better than the existing version.

I also have a sizing issue between the `chat-area` and the `occupants` `div`s, which I’ve been unable to fix.

sass/discord.css
```diff
@@ -0,0 +1,56 @@
+#conversejs.theme-discord {
```
jcbrand: I think we should not use the name `discord` anywhere. You can come up with a new name, or we simply call it the "dark" theme.

jcbrand:
Thanks!

licaon-kter:
@stevenroose As said in the MUC: https://conversejs.org/docs/html/configuration.html#theme

licaon-kter:
@linkmauve Only MUCs were styled? 

1:1 still (ugly) green
cxontact

PR #1168, opened 2018-08-18
Use the roster name or vCard nickname before falling back to vCard fullname or JID

linkmauve:
The order of vCard nickname and fullname is subject to debate, but I just copied what was done elsewhere in Converse. The roster override though should imo always be first.


jcbrand:
Thanks

Issue #1109, opened 2018-06-04
XEP-0156: Discovering Alternative XMPP Connection Methods


jcbrand:
Via @linkmauve  from discuss@conference.conversejs.org:

> It’s basically a matter of doing a GET XHR to https://&lt;user-domain&gt;/.well-known/host-meta
And parsing the results.

Some initial code here:
https://linkmauve.fr/files/0156.js

linkmauve:
This is a duplicate of #129, fyi.

A new WIP version is available here: https://github.com/linkmauve/converse.js/tree/xep-0156

PR #1170, opened 2018-08-19
WIP: Add support for XEP-0156

### LINKED ISSUE #129: XEP-0156 (Discovering Alternative XMPP Connection Methods)
If I have properly set up a BOSH server and configured my DNS to serve up a URL per XEP-0156 (e.g., `koehn.com`), there's no need to use a third-party connection manager. The system could check the DNS for a `TXT` record at the XMPP host with text starting with `_xmpp-client-xbosh` and use that as the connection manager. 

Trouble is I'm not sure how you do a `TXT` DNS resolution from the browser; you certainly could have a server-side component to do the discovery though, you'd just have to have a standard API to send the user's JID and get back the connection manager to use. The server-side component could get the domain of the JID, check the XEP-0156 DNS entries, and return either a default connection manager or the one used by the domain. 

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/17882044-xep-0156-discovering-alternative-xmpp-connection-methods?utm_campaign=plugin&utm_content=tracker%2F194169&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F194169&utm_medium=issues&utm_source=github).
</bountysource-plugin>

linkmauve:
It is disabled by default until more testing has been done, set
use_xep_0156 to true for now.

There is also the very big problem that the `Strophe.Connection` is
created before the user has had the time to enter their JID, which is
why we currently hardcode the domain to be resolved (this MUST be fixed
before this can be merged).  I haven’t been able to figure out where to
move this Connection creation, which must be after the JID is known but
before the actual connection is attempted.

There is also a bit of code duplication in it, this will be fixed in a
future version of this commit.

Fixes #129 and #1109.


jcbrand:
@linkmauve The login code is fairly complex because there are multiple different ways a user could log in.

They could log in manually with the login form or automatically via XHR-fetched credentials, config credentials passed in to `converse.initialize` or via BOSH prebind.

Also, the different login types happen at different times within startup/setup of the client.
Manual login happens after `converse.initialize` has completed, automatic login happens before it has completed.

Due to this there is no one place where login can happen. It can happen in `converse-core` (automatic) or in `converse-controlbox` (manual).

Manuel login happens in the [authenticate](https://github.com/conversejs/converse.js/blob/master/src/converse-controlbox.js#L461) method of the LoginPanel view in converse-controlbox.js

Automatic login happens in various places in converse-core.js, such as (but not only) the [autoLogin](https://github.com/conversejs/converse.js/blob/master/src/converse-core.js#L1034) method.

In all these methods, it's assumed that `converse.connection` exists.

Modern JavaScript has a [getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get) syntax which allows you to call a method as if it was an attribute, similar to Python's `@property` decorator.

I think we should rewrite the `initConnection` method to be a `getter` named `connection` instead.

Something like:

```
const converse = {
    get connection () {
        // code comes here
    }
}
````

Then, whenever `_converse.connection` is called, the connection will be instantiated if it doesn't exist already.

That might not be enough yet, because the connection might still be instantiated in code at times when we don't have a JID, but I think it's a good first step and those problematic cases will then have to be refactored somehow.

PR #1171, opened 2018-08-21
Keep multiple whitespace characters as is

linkmauve:
Currently multiple consecutive spaces or tabs were being transformed into a single space, rendering some ASCII art unreadable.  This patch fixes it by giving each message text the CSS behaviour of <pre/>.


jcbrand:
Thanks!

PR #1232, opened 2018-10-06
Fix failing build because of website.scss name mismatch

cmrd-senya:
`make dist` failed because the Makefile expects `sass/_website.scss` to be
present, but the file was called `sass/website.scss`


jcbrand:
Thanks but I renamed that file on purpose yesterday. I made the fix in the Makefile.

Issue #1209, opened 2018-09-20
Bring color codes of users' status in line with other clients


alexara:
Hi! I got a little confused by the current color codings of the users' status. I was really concerned that my contacts are all offline, until I looked very very closely to see that there is a tiny difference between "away" and "offline": 

* Offline: Gray with a X
* Away:  Gray with a O
* DND: Red with -
* Online: Green

Do you mind to bring it in line with other chat clients e.g. Conversations, i.e.

* Offline: Gray
* Away:  Orange
* DND: Red with -
* Online: Green

Regards

laurenmayers:
Happy to work on this if I can!

jcbrand:
Go for it!

laurenmayers:
Hey @jcbrand, where's the best place to contact you/the right people for help?
I'm having trouble locating the file/code that sets the background color of the users' status.

jcbrand:
@laurenmayers Asking here is fine, alternatively you can [ask in the our chatroom](https://inverse.chat/#converse/room?jid=discuss@conference.conversejs.org). You'll need an XMPP account for that, you can register one at `conversejs.org`. Pretty soon it'll also be possible to log in via Github.

For this ticket, we need to modify the CSS to make sure that the away status renders orange.
However, we use [Sass](http://sass-lang.com/) to generate our CSS, so the source file where we need to make a change is in the [sass](https://github.com/conversejs/converse.js/tree/master/sass) folder.

I don't know the relevant CSS class that determines the style, so I used the browser's DOM inspector to look at the DOM element in question. I saw that it has the class `.fa-dot-circle`.

I then used `grep` to search for files containing `fa-dot-circle` in the `sass` directory and I found the file file [_controlbox.scss](https://github.com/conversejs/converse.js/blob/d30efcb3abd70b9b187fa8c2d9f3f801f4accc82/sass/_controlbox.scss#L32).

By looking at the place where the string `fa-dot-circle` appears, I immediately saw the bug (in part because I have context, due to having worked on this before).

In the Sass file, the class is `fa-dot-circle-o`, not `fa-dot-circle`.  Recently some FontAwesome classes were renamed, but the Sass file didn't update accordingly. The away status was actually shown in orange originally.

So by renaming `.fa-dot-circle-o` to `.fa-dot-circle` in that file, the bug is fixed. Now it's just a matter of running `make watch` to regenerate the CSS file.

laurenmayers:
[...]

PR #1236, opened 2018-10-07
#1209 update user status colors

laurenmayers:
Fixed [issue #1209: bring color codes of users' status in line with other clients](https://github.com/conversejs/converse.js/issues/1209#issuecomment-427617211).

Previous status colors:
status-colors-previous

Updated status colors:
status-colors-updated

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@
 - #1199 Can't get back from to login screen from registration screen
 - #1214 Setting `allow_contact_requests` to `false` has no effect
 - #1222 Adding a bookmark should prefill the room name
+- #1209 Bring color codes of users' status in line with other clients
 
 ## 4.0.2 (2018-10-02)
 
--- a/css/converse.css
+++ b/css/converse.css
@@ -1160,7 +1160,7 @@ readers do not read off random characters that represent icons */
   content: "\f52b"; }
 
 .fa-dot-circle:before {
-  content: "\f192"; }
+  content: "\f111"; }
 
 .fa-dove:before {
   content: "\f4ba"; }
@@ -9714,9 +9714,9 @@ body.reset {
 #conversejs .xmpp-status .fa-minus-circle,
 #conversejs .roster-contacts .fa-minus-circle {
   color: #E77051; }
-#conversejs .set-xmpp-status .fa-dot-circle-o,
-#conversejs .xmpp-status .fa-dot-circle-o,
-#conversejs .roster-contacts .fa-dot-circle-o {
+#conversejs .set-xmpp-status .fa-dot-circle,
+#conversejs .xmpp-status .fa-dot-circle,
+#conversejs .roster-contacts .fa-dot-circle {
   color: #E7A151; }
 #conversejs .set-xmpp-status .far.fa-circle,
 #conversejs .set-xmpp-status .fa-times-circle,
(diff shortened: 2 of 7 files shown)


css/converse.css
```diff
@@ -1160,7 +1160,7 @@ readers do not read off random characters that represent icons */
   content: "\f52b"; }
 
 .fa-dot-circle:before {
-  content: "\f192"; }
+  content: "\f111"; }
```
jcbrand: Did you modify this file directly? I don't see where you made this change in any scss (Sass) file.

In any case, this is not the right approach. These font-awesome classes should be considered as constants, they never change.

Instead, if you want to show a solid circle, then you need to use a different class, not modify `.fa-dot-circle`.

Looks like the classes you need to use is `fas fa-circle`, that's a solid circle.
See here: https://fontawesome.com/icons/circle?style=solid

CHANGES.md
```diff
@@ -8,6 +8,7 @@
 - #1199 Can't get back from to login screen from registration screen
 - #1214 Setting `allow_contact_requests` to `false` has no effect
 - #1222 Adding a bookmark should prefill the room name
+- #1209 Bring color codes of users' status in line with other clients
```
jcbrand: Please make sure these are sorted according to issue number.

css/font-awesome.css
```diff
@@ -0,0 +1,3795 @@
+@font-face {
```
jcbrand: I know this file gets built when you run `make watch` or `make css`, but I don't think it should be included... I should probably update the Makefile to avoid building it.

Please remove it from this pull request.

css/converse.css
laurenmayers: Ah this should have been obvious, I'll fix this up.

css/font-awesome.css
laurenmayers: Ok, will do.

CHANGES.md
laurenmayers: Oh I didn't realise it was sorted that way. Thanks, will change.

css/converse.css
laurenmayers: I will need to add a new class name so that the away icon does not override the online icon, since both use the fa-circle class name. Do you have a system or preference for amending the class name for the away fa-circle?
For example, I notice the 'Away for long' status uses fa-circle, but it has the preceding class 'far' rather than 'fa'.

jcbrand: We've been using the fontawesome class names to style the icons, but this isn't really a good idea, we need our own class names, as you've figured out.

So please add the class `chat-status` to all 4 status icons, and then another class `chat-status--${show}` as well.

So for the away icon the classes are `fa fa-circle chat-status chat-status--away` and for the online icon it will be `fa fa-circle chat-status chat-status--online`.

jcbrand:
Hi @laurenmayers 

Do you think you'll be able to finish this ticket before the end of the week? I'd like to make a new release soon and would like to include this in the next release.

If not, that's fine and I'll see if I can fix it before releasing.

laurenmayers:
@jcbrand Hey, yes for sure 👍🏼I'll be finishing it off tonight :)

laurenmayers:
@jcbrand Hey JC, having some trouble trying to bring about changes with the additional class names. I'm not sure with work if I'll be able to get this done by the end of the week - it might be best if I surrender now. Thanks for the opportunity though.

laurenmayers:
@jcbrand No worries! If I could still work on it, I would love to! I could get it to you early this week coming, if that works.

laurenmayers:
Hey I have the code finished, just working out how to rebase/revert my previous commits so that I can PR the new stuff only, if that makes sense. Will be PR'd today! :)

jcbrand:
Hi @laurenmayers, if you're struggling with rebasing or reverting, don't worry about that. Just make a PR with all the commits.

I can very easily squash them into one commit and if necessary rebase them onto the latest master branch.

I'd like to make a release soon.

laurenmayers:
Thanks @jcbrand! Appreciate your patience on this one!

PR #1246, opened 2018-10-12
Create a lerna managed package "@converse/headless"

cmrd-senya:
This PR introduces lerna support for this repository.

Lerna is configured to search for packages in `./packages` dir and the root directory.

Currently there is only one new package `@converse/headless` introduced here. Another package is the `converse.js` package in the root directory. Since the `converse.js` package now depends on `@converse/headless` the root package must therefore be under lerna management too.

The main thing that lerna does is simlinking internal dependencies inside the repository.

The root converse NPM package depends on `@converse/headless` and lerna makes it available for it by adding a symlink to `./node_modules`. To use that feature instead of running `npm install` one should run `lerna bootstrap --hoist` which is a wrapping around `npm install`. I replaced the `npm install` in the `Makefile` for that reason.

With this PR, `make dist` passes well and generates `converse.js` builds using the new dependency structure.

I noticed that some of your aliases in your webpack configuration doesn't make too much sense. For instance you resolve backbone as:
```
"backbone":                 path.resolve(__dirname, "node_modules/backbone/backbone"),
```
Which actually doesn't make difference with not resolving it. If you just point `backbone` as a package dependency, it will pick the file from the `package.json#main` value, which points to the same `node_modules/backbone/backbone.js` file anyway. So everything will work the same way without this line.

Since `@converse/headless` package is intended to use outside of `converse.js` build system, it can't actually rely on webpack configuration. There is no way you can built webpack configuration to a NPM package, so I fixed dependencies inside `@converse/headless` so that they are loaded properly without aliasing. Files outside of `@converse/headless` in their turn now depend on files from `@converse/headless` NPM package. Lerna is responsible for "installing" `@converse/headless` dependency for the root package here.

Another important note: it looks like `lerna bootstrap --hoist` actually fails if there is no referenced package in the repository. When you run `lerna bootstrap --hoist` it links the NPM package from `./packages/headless`, but it fails, if `@converse/headless` of the required version is not available in the NPM repository. I don't know why it is so. In my case I'm using a private NPM repository and I just pushed `@converse/headless` v4.0.2 there, and it worked fine. This doesn't seem like a terrible issue (though it is somewhat inconvenient), it just means you have to bump package versions right before you push them to the NPM repo and keep versions while you do development.

You can do version bumping manually, but lerna actually includes a command which automates it. [`lerna version`](https://github.com/lerna/lerna/blob/master/commands/version#readme) allows you to bump version of all the managed packages in the repository.

Also, if you'd like to add dependencies to package.json of a managed package,  you can use [`lerna add`](https://github.com/lerna/lerna/tree/master/commands/add#readme) for that.

Also my text editor fixed some occasional EOL whitespaces, I hope it is fine.


cmrd-senya:
So here is what I was talking about: https://travis-ci.org/conversejs/converse.js/builds/440681827#L502

```
npm ERR! code E404
npm ERR! 404 Not Found: @converse/headless@4.0.2
npm ERR! A complete log of this run can be found in:
npm ERR!     /home/travis/.npm/_logs/2018-10-12T15_15_52_286Z-debug.log
make: *** [node_modules/.bin/lerna] Error 1
```

Lerna is not going to make a build unless you have at least something at `@converse/headless`. Maybe you can push some kind of stub there?

cmrd-senya:
Alright, I modified travis configuration to make it work with lerna. I also pushed a stub package `@converse/headless@4.0.2` to override the issue I described earlier.

Now travis is green!

jcbrand:
Excellent work, thank you.

I think I'd like to make the 4.0.3 release before merging this PR, since these changes seem too invasive for a patch release. I don't want people to be blindsided by files no longer being where they were in 4.0.2.

I'll then merge this PR for a 4.1.0 release.

Besides that, I think there are a few more things to do related to this PR (but they don't necessarily have to be done within this PR).

These should be done (if possible) before we make a 4.1.0 release:

* [ ] Package @converse/headless without moment locales
* [ ] Remove the html templates currently in the @converse/headless package 
* [ ] Add converse-muc to @converse/headless
* [ ] Remove backbone.nativeview from @converse/headless

Note, I'm not implying you should do all this. I'll likely also work on these.

cmrd-senya:
Alright then, I'll do a rebase after you release 4.0.3.

> Besides that, I think there are a few more things to do related to this PR (but they don't necessarily have to be done within this PR).

Yeah, I think it is fine to work on the things you listed after you merge this PR. What I can do in the scope of this PR is to add `converse-muc.js` to the `@converse/headless` package. Adding it is easy.

It is harder to me to work on the removal of dependencies because I don't know what is gonna break and how to fix it, so I'll need your help with that.

jcbrand:
Sounds good. I'll likely make the 4.0.3 some time next week.

cmrd-senya:
I updated the PR and moved `converse-muc.js` to `@converse/headless`

jcbrand:
Will be merged via #1256 where I rebased it and made some further changes.

PR #1258, opened 2018-10-22
#1257: Prefer 'probably' over 'maybe' when evaluating Audio support.

### LINKED ISSUE #1257: Notification audio doesn't (always) play
When enabling notifications, the corresponding audio doesn't always play. From todays chatlog:

> Guus 17:41
> jcbrand: I'm not getting the sound notifications to work. Do I explicitly install the plugin (even though the webapp already downloads the .ogg file?)
>  
> Dele Olajide 17:45
> I had an issue with sound in Pade as well. I assumed it was chrome extension issue
> ```
> if (audio.canPlayType('audio/ogg') == "probably") {   // BAO
> ```
>
> Dele Olajide 17:47
> chrome returns "maybe" and does not play it
>
> Link Mauve 17:49
> You most likely need to use audio/ogg;codec=vorbis instead.
> Or codec=opus if that’s what you are using.
>
> Dele Olajide 17:56
> The msg_received.ogg file in converse is not playable by my chrome browser. I needed it to fall-through an play the msg_received.mp3 instead

guusdk:
This aims to resolve #1257 but is untested by me!


deleolajide:
Typo
```
        let audioOgg, audioMp3, canPlayOgg, canPlayMp3;
```
Once I fixed that, it worked

guusdk:
Thanks for checking, @deleolajide! I've fixed the typo.

Link Mauve commented on this change via chat:
> A much better solution is to use the <source/> element, instead of two <audio/> elements.
That way the browser will select the best one to play.
Guus, see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio#<audio>_with_multiple_<source>_elements

I have no reason to question this, but such a change is above my pay grade. As the change in this PR is an improvement over the existing code, I'd suggest merging this as-is. The additional improvement suggested by Link Mauve can be added in due time.

jcbrand:
Thanks for the fix @guusdk!

PR #1265, opened 2018-10-24
Convert a few functions to async/await

linkmauve:
This seems to have started a few commits ago, so let’s continue. :)


jcbrand:
Continuing in #1269

PR #1267, opened 2018-10-25
Switch from SASS variables to CSS custom properties

linkmauve:
This will eventually allow client-side theming support.


linkmauve:
Superseded by #1268.

PR #1268, opened 2018-10-25
Switch from SASS variables to CSS custom properties

linkmauve:
This will eventually allow client-side theming support.

Supersedes #1267.

### DIFF
--- a/css/converse.css
+++ b/css/converse.css
@@ -3804,31 +3804,114 @@ readers do not read off random characters that represent icons */
   position: static;
   width: auto; }
 
-/*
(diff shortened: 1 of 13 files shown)


jcbrand:
Thanks a lot for this! I'll look into merging after the next release.

jcbrand:
Great work on this, thanks.

Unfortunately looks like the Sass variables also applied to the bootstrap sass files, but now with the CSS properties the bootstrap styles get their original colours (look for example at the badges).

We'll have to add some new css rules to make up for that.

PR #1269, opened 2018-10-25
Async await

jcbrand:
Continuing #1265

### DIFF
--- a/css/converse.css
+++ b/css/converse.css
@@ -9569,12 +9569,15 @@ body.reset {
   max-height: 200px; }
 #conversejs.converse-overlayed .emoji-picker {
   height: 100px; }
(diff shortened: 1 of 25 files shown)


jcbrand:
@linkmauve Thanks very much for the async/await changes.

In future, please make sure to also include an updated build file (`dist/converse.js`). Otherwise your changes don't actually get tested.

There were lots of failing tests (mostly due to timing issues because async/await appears to be a bit slower) that had to be fixed and which took multiple hours.

I also had to revert some of your changes due to breaking tests.
See here: https://github.com/conversejs/converse.js/pull/1269/commits/6519083414732d2c5dfb5aa2c01b21845bd28f42

PR #1271, opened 2018-10-25
Fix #1252, incorrect bookmark icon title

### LINKED ISSUE #1252: Incorrect text in tooltipo of top-right bookmark button
When joined a bookmarked multi-user chat room, the tooltip of the 'bookmark' button in Converse 4.0.2 reads "Bookmark this groupchat". As the groupchat is already bookmarked, the tooltip should probably read something like "Unbookmark this groupchat."

Of interest: after applying the bookmark, the icon does change (from 'unbookmarked' to 'bookmarked' state). Als, the behavior is correct (clicking the button will unbookmark the room). The problem seems to be isolated to the tool tip text.

Ppjet6:
Please tell me if there's a better way to do that.
Still figuring out how tests work.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 - Use ES2015 modules instead of UMD.
 - #1257 Prefer 'probably' over 'maybe' when evaluating audio play support.
 - #1261 File upload not working
+- #1252 Correctly reflect the state in bookmark icon title.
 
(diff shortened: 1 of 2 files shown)


jcbrand:
Thanks, looks good to me. :)

jcbrand:
BTW, next time, please make sure CHANGELOG entries are sorted by issue number.

PR #1277, opened 2018-10-28
Switch avatar rendering from canvas to SVG.

### LINKED ISSUE #1156: Make sure SVG avatars work on Firefox too
When joining a MUC where my regular JID is present, I see most of the other people’s avatars except for mine, which is transparent rather than unset.

In Firefox’s dev console, tab Storage, I correctly see the complete base64’d SVG file in the `image` key, as well as the image/svg+xml in `image_type`.

wayland-screenshot-2018-08-10_18-19-20

linkmauve:
This delegates the calculation of the aspect ratio to the browser, makes the rendered image independent of the device resolution, and generally simplifies the code.

Fixes #1156.


jcbrand:
Merged manually.

PR #1278, opened 2018-10-28
Replace the default avatar with a SVG version

linkmauve:
This is both lighter, looks better when rendered at arbitrary resolutions (as there is no scaling being done), and actually editable for different deployments.

This image has been hand-drawn using the PNG as a background.


jcbrand:
Thanks!

PR #1295, opened 2018-11-02
Upload files on drag-and-drop to the textarea

### LINKED ISSUE #1188: Feature request - drag and drop file to HTTP Upload
I dragged and dropped a file into inverse.chat - Firefox opened the file instead, and when I pressed back I was logged out :)

Firefox 62.0 (32-bit), Debian GNU/Linux 9.5 (stretch)

linkmauve:
When one or more files are dropped on the textarea, they get uploaded or an error is printed if this isn’t possible.

There is also a fix to correctly displays an error when XEP-0363 isn’t supported by the server.

Fixes #1188.

### DIFF
--- a/dist/converse.js
+++ b/dist/converse.js
@@ -59022,7 +59022,9 @@ _converse_headless_converse_core__WEBPACK_IMPORTED_MODULE_5__["default"].plugins
         'click .toggle-smiley': 'toggleEmojiMenu',
         'click .upload-file': 'toggleFileUpload',
         'input .chat-textarea': 'inputChanged',
-        'keydown .chat-textarea': 'keyPressed'
+        'keydown .chat-textarea': 'keyPressed',
+        'dragover .chat-textarea': 'onDragOver',
+        'drop .chat-textarea': 'onDrop'
       },
 
       initialize() {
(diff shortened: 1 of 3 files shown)


dist/converse.js
```diff
@@ -59128,6 +59130,19 @@ _converse_headless_converse_core__WEBPACK_IMPORTED_MODULE_5__["default"].plugins
...
       },
 
+      onDragOver(evt) {
+        evt.preventDefault();
+      },
+
+      onDrop(evt) {
+        /* There are no files to be dropped, this isn’t a file transfer
+         * operation.
+         */
```
jcbrand: This comment confuses me. Generally I use the `/* */` form as docstrings explaining the purpose of the method/function.

jcbrand:
Thanks! Glad to have this feature.

dist/converse.js
linkmauve: Oh, usually I see `/** … */` in the role of docstrings; do you prefer `//` for plain comments then?

linkmauve: This is now fixed in 9a12afc70230a45afed180756ff1c8d24a0edb37.

PR #1302, opened 2018-11-03
Improve XEP-0245 support

linkmauve:
Namely:
- Only match messages which start with exactly "/me ".
- Simplify the match, checking a string instead of a complex regex.
- Always remove the first four characters in the case of a match, to not
leave an extra leading space.


jcbrand:
Thank you!

PR #1305, opened 2018-11-04
new value 'all' for 'show_desktop_notifications'

ChaosKid42:
I propose that we add an option `notify_in_view` which defaults to `false`. Enabling it would make converse.js show notifications even if the tab of converse.js is shown to the user.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,7 @@
 - #1268 Switch from SASS variables to CSS custom properties
 - #1278 Replace the default avatar with a SVG version
 - #1306 added option `notification_delay`
+- #1305 added value 'all' for 'show_desktop_notifications' to notifiy even if converse.js is open
 
 ## 4.0.4 (2018-10-29)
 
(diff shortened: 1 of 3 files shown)


jcbrand:
I'd like us to reuse existing config settings whenever possible.

So instead of having a new option, I suggest we re-use [show-desktop-notifications](https://conversejs.org/docs/html/configuration.html#show-desktop-notifications) for this.

It can then have the following values:

* `true` or `'hidden'` which is the current functionality.
* `false` or `null` or `''`, which disables desktop notifications.
* `all` which would be like setting `notify_in_view` to `true`.

This would be a backwards compatible way of adding your desired feature without adding a new config option.

ChaosKid42:
@jcbrand Another great idea. PR updated.

ChaosKid42:
@jcbrand: I rebased the commit. Should be bossible to merge it now.

jcbrand:
Thanks!

PR #1306, opened 2018-11-04
added option "notification_requires_interaction"

ChaosKid42:
This PR introduces an option `notification_requires_interaction` which if enabled makes the notifications stay until explicitly dismissed by the user.


jcbrand:
Thanks @ChaosKid42!

I have a different suggestion however.

We can make the 5000ms value configurable with `5000` the default and with `0` meaning that the message doesn't disappear.

That way your feature becomes available and its possible to change the timeout to something else. Two birds with one stone :)

What do you think?

ChaosKid42:
@jcbrand Great idea. Changed the PR accordingly.

jcbrand:
Thanks!

PR #1319, opened 2018-11-12
Send presences according to XEP-0319: Last User Interaction in Presence

ChaosKid42:
This implements sending of idle presences according to XEP-0319: Last User Interaction in Presence. Using this, other clients can show at what time the user has been last active:

screenshot_20181112-194846


jcbrand:
Thanks!

PR #1324, opened 2018-11-14
Add XHTML compatibility

### LINKED ISSUE #1131: Converse does not work on xhtml pages

linkmauve:
This is mainly a testing tool, which makes a mode where all HTML bugs
which could slightly mess up the rendering will trigger proper errors in
the browser.

Fixes #1131.


jcbrand:
Thanks

PR #1331, opened 2018-11-16
Fix missing checkmarks in old muc messages

ChaosKid42:
I see missing checkmarks in older muc messages. This tries to fix it.


jcbrand:
Thanks!

Issue #1333, opened 2018-11-19
Dont request Receipts in MUCs


lovetox:
I question the usefulness of these, in big MUCs it would generate a lot of traffic in theory.

Server Admin would need to block them from going to MAM, because otherwise the Archive would grow exponentially

Altough this showed a bug in Gajim that it answers these requests, which i will fix.

jcbrand:
CC @ChaosKid42 

@lovetox Have you seen receipt requests in the stanza traffic? If so, that's a bug.

lovetox:
yes, i use inverse.chat, so dont know on what version they are

```
<message xmlns="jabber:client" xml:lang="en" to="me" from="gajim@conference.gajim.org/lovetox__" type="groupchat" id="d3141fdd-05b0-49f6-926d-0a47e0171594">
<active xmlns="http://jabber.org/protocol/chatstates" />
<request xmlns="urn:xmpp:receipts" />
<body>a</body>
</message>
```

ChaosKid42:
@lovetox @jcbrand I believe it's a bug. https://github.com/conversejs/converse.js/pull/1339 is supposed to fix it.

jcbrand:
Thanks for the quick fix @ChaosKid42

PR #1339, opened 2018-11-20
don't send receipt requests in MUCs

ChaosKid42:
This fixes https://github.com/conversejs/converse.js/issues/1333.


jcbrand:
Thanks for the quick fix!

Issue #1322, opened 2018-11-13
Display occupants’ avatars in the occupants list


linkmauve:
This will make the list a lot more user-friendly, and help identify people right away without having to read their name.

Nyco:
This can apply to both occupants in a MUC on the right side, as well as the left bar showing both "Groupchats" and "Contacts".

PR #1354, opened 2018-11-27
Display occupants avatar

raphaelSeguin:
Hi,
this is a first contribution. There are some 
I copied the AvatarMixin from chatbox-views, would have it been better to import the whole file, or maybe abstract this mixin in separate file ?

### DIFF
--- a/sass/_roster.scss
+++ b/sass/_roster.scss
@@ -118,7 +118,7 @@
                     }
 
                     .avatar {
-                        float: left;
+                        //float: left;
                         margin-right: 0.5em;
                     }
                 }
--- a/sass/_roster.scss
+++ b/sass/_roster.scss
@@ -119,7 +119,9 @@
 
                     .avatar {
                         float: left;
-                        display: inline-block;
+                        //display: inline-block;
+                        margin-right: 0.5em;
+                        //margin-top: -4px;
                     }
                 }
                 &.current-xmpp-contact span {
--- a/sass/_roster.scss
+++ b/sass/_roster.scss
@@ -119,9 +119,7 @@
 
                     .avatar {
                         float: left;
-                        //display: inline-block;
                         margin-right: 0.5em;
-                        //margin-top: -4px;
                     }
                 }
                 &.current-xmpp-contact span {
--- a/src/converse-rosterview.js
+++ b/src/converse-rosterview.js
@@ -23,7 +23,7 @@ import tpl_search_contact from "templates/search_contact.html";
 const { Backbone, Strophe, $iq, b64_sha1, sizzle, _ } = converse.env;
 const u = converse.env.utils;
 
-// this is copied from converse-chatbowviews.js
+// // this is copied from converse-chatbowviews.js
 const AvatarMixin = {
 
     renderAvatar (el) {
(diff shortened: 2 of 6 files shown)


src/converse-rosterview.js
```diff
@@ -480,7 +506,11 @@ converse.plugins.add('converse-rosterview', {
                         'desc_chat': __('Click to chat with %1$s (JID: %2$s)', display_name, item.get('jid')),
                         'desc_remove': __('Click to remove %1$s as a contact', display_name),
                         'allow_contact_removal': _converse.allow_contact_removal,
-                        'num_unread': item.get('num_unread') || 0
+                        'num_unread': item.get('num_unread') || 0,
+                        classes: '',
+                        width: '24',
+                        height: '24',
```
jcbrand: Looks like the `width` and `height` values you're passing in here never get used.

src/converse-rosterview.js
```diff
@@ -22,6 +23,26 @@ import tpl_search_contact from "templates/search_contact.html";
...
+                image = this.model.vcard.get('image');
+
+        canvas_el.outerHTML = tpl_avatar({
+            'classes': canvas_el.getAttribute('class'),
+            'width': canvas_el.width,
+            'height': canvas_el.height,
+            'image': "data:" + image_type + ";base64," + image,
+        });
+    },
+};
```
jcbrand: There's no need to duplicate the code. Instead, in `src/converse-chatboxviews.js`, you can set `AvatarMixin` as an attribute on the `_converse` object.

Then further down you can do:

```_converse.ViewWithAvatar = Backbone.NativeView.extend(_converse.AvatarMixin);```

File: sass/_roster.scss [hunk 3]
```diff
@@ -119,7 +119,9 @@
 
                     .avatar {
                         float: left;
-                        display: inline-block;
+                        //display: inline-block;
+                        margin-right: 0.5em;
+                        //margin-top: -4px;
```
jcbrand: Please don't leave uncommented code anywhere.

sass/_roster.scss
```diff
@@ -119,7 +119,9 @@
 
                     .avatar {
                         float: left;
```
jcbrand: We're using flexbox for many things now, so this `float: left` statement might no longer be necessary.

sass/_roster.scss
raphaelSeguin: Oups ! OK !

src/converse-rosterview.js
raphaelSeguin: I'll remove them then.

sass/_roster.scss
raphaelSeguin: Yes. In fact, I found it here, and when I tried to remove it, it broke the design. I'll try to remove it again and fix the display with flexbox.

sass/_roster.scss
raphaelSeguin: removed

linkmauve:
This fixes #1322, btw.

jcbrand:
I've made some updates to this PR and then did a manual merge.

Issue #1353, opened 2018-11-27
Message Delivery Receipts not working because of the message "type" attribute


catalincic:
Hi, it's the first issue I open so please understand if I do not follow the exact guidelines.

While testing the new Message Delivery Receipts feature you have merged into master recently (few days ago), I have noticed that there is a minor and strange bug which makes the delivery receipts not working.

**Cause:** The message handler registered in converse-chatboxes.js at line 627 is filtering the stanzas by their type=chat attribute. Code:

```
registerMessageHandler () {
                _converse.connection.addHandler((stanza) => {
                    this.onMessage(stanza);
                    return true;
                }, null, 'message', 'chat'); // << This line here
               ....
            },
```

But the delivery receipt stanza sending code does not add the type=chat attribute to the message stanza, hence making the message handler miss the delivery receipts.

```
sendReceiptStanza (to_jid, id) {
                const receipt_stanza = $msg({
                    'from': _converse.connection.jid,
                    'id': _converse.connection.getUniqueId(),
                    'to': to_jid,

                   // This is not in the code but should be added; or change the message handler
                    'type': 'chat' 
                }).c('received', {'xmlns': Strophe.NS.RECEIPTS, 'id': id}).up();
                _converse.api.send(receipt_stanza);
            },
```

As far as I can tell, it is indeed a bug - maybe I have missed something, but while testing delivery receipts, I noticed they were not working and the cause was this attribute missing. Checking the XEP-0184 specs, I do not see any type=chat attribute in the guidelines, so probably the message handler needs to be changed instead of adding that attribute. I would happily do it, but I am afraid I do not have the required experience with this to make sure the code changes will work properly.

PR #1380, opened 2018-12-16
Fix Message Delivery Receipts not working because of the message "typ…

ChaosKid42:
Fixes https://github.com/conversejs/converse.js/issues/1353


jcbrand:
@ChaosKid42: Looks like we worked on this on parallel :(

My fix is already in...

jcbrand:
BTW, this bug didn't surface in your tests because you called `_converse.chatboxes.onMessage` manually.

I know you've probably seen this in other tests. I've learned over time however that it's generally better to test stanza traffic via `_converse.connection._dataRecv(test_utils.createRequest(stanza));`.

That way the handlers are for example also tested and in this case the bug would have become clear.

PR #1382, opened 2018-12-16
set store hint on receipts and type='chat'

ChaosKid42:
I believe it's necesary to store receipts. Otherwise they might not reach the original sender (e.g. in case sender and receiver are not online at the same time). If they are stored in the archive, they will be retrieved later.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@
 - #1376 Fixed some alignment issues in the sidebar
 - #1378 Message Delivery Receipts were being sent for carbons and own messages
 - #1379 MUC unread messages indicator is failing
+- #1382 Message Delivery Receipts: Set store hint and type='chat'
 
 ## 4.0.6 (2018-12-07)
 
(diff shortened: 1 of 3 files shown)


jcbrand:
AFAIK the `store` hint is for storing the message in MAM and I'm not sure that we want to store receipts in MAM.

The XMPP server already keeps track of messages that were sent to offline users and sends it to them when they come online (without the `store` hint). I assume this also applies to receipts.

ChaosKid42:
I believe you are right regarding offline storage. Nevertheless this won't help if the user is online with another device. If the user's first device comes online again, only mam will help to retrieve the receipt I believe.
btw: Conversations seems to set the `store` hint, too.

ChaosKid42:
@jcbrand I really believe this is necessary in order to synchronize state between different clients of the same user. If your unhappy with this PR I could also make this an configurable feature. What do you think?

jcbrand:
Thanks

PR #1388, opened 2018-12-22
implement /destroy command in muc

ChaosKid42:
This implements the possibility to destroy a MUC.

### DIFF
--- a/spec/chatroom.js
+++ b/spec/chatroom.js
@@ -2544,7 +2544,7 @@
                 });
 
                 const info_messages = Array.prototype.slice.call(view.el.querySelectorAll('.chat-info'), 0);
-                expect(info_messages.length).toBe(18);
+                expect(info_messages.length).toBe(19);
                 expect(info_messages.pop().textContent).toBe('/voice: Allow muted user to post messages');
                 expect(info_messages.pop().textContent).toBe('/topic: Set groupchat subject (alias for /subject)');
                 expect(info_messages.pop().textContent).toBe('/subject: Set groupchat subject');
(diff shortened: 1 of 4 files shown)


src/converse-muc-views.js
```diff
@@ -884,12 +895,21 @@ converse.plugins.add('converse-muc-views', {
                                 this.model.get('jid'), args[0], 'participant', args[1],
                                 undefined, this.onCommandError.bind(this));
                         break;
+                    case 'destroy':
+                        if (!this.verifyAffiliations(['owner'])) {
+                            break;
+                        }
+                        this.destroy(
+                                this.model.get('jid'), args[0],
+                                this.onDestroySuccess.bind(this), this.onCommandError.bind(this));
```
jcbrand: Instead of making a new method `onDestroySuccess`, you can simply use `this.close` here, and you can use an arrow function to avoid having to use `.bind`.

In `destroy`, you can simply return `_converse.api.sendIQ(iq)` so that it returns a promise.

So it could be something like this:
```
 this.destroy(this.model.get('jid'), args[0])
       .then(() => this.close())
       .catch(e => this.onCommandError(e));
```

src/converse-muc-views.js
```diff
@@ -797,6 +797,13 @@ converse.plugins.add('converse-muc-views', {
                 }
             },
 
+            destroy (groupchat, reason, onSuccess, onError) {
+                const destroy = $build("destroy");
+                const iq = $iq({to: groupchat, type: "set"}).c("query", {xmlns: Strophe.NS.MUC_OWNER}).cnode(destroy.node);
+                if (reason && reason.length > 0) { iq.c("reason", reason); }
+                return _converse.api.sendIQ(iq).then(onSuccess).catch(onError);
```
jcbrand: I'd like us to move away from using callbacks in favour of using promises.

Here, you can simply return `return _converse.api.sendIQ(iq)` and the later when you call `destroy` you can either use `then` and `catch` or `async/await`.

spec/chatroom.js
```diff
@@ -3326,6 +3327,54 @@
...
+                    null, ['rosterGroupsFetched'], {},
+                    async function (done, _converse) {
+
+                await test_utils.openAndEnterChatRoom(_converse, 'lounge', 'localhost', 'dummy');
+                var sent_IQ, IQ_id;
+                var sendIQ = _converse.connection.sendIQ;
+                spyOn(_converse.connection, 'sendIQ').and.callFake(function (iq, callback, errback) {
+                    sent_IQ = iq;
+                    IQ_id = sendIQ.bind(this)(iq, callback, errback);
+                });
```
jcbrand: Not a dealbreaker for merging, but I'd like us to move away from using `var` in favour of `let` and `const`.

ChaosKid42:
@jcbrand Thanks fo your review. Updated the PR.

jcbrand:
Thanks!

PR #1396, opened 2019-01-05
Force avatar refetch if new hash received

### LINKED ISSUE #1334: Force avatar refetch when receiving vcard-temp:x:update
It seems that converse will only fetch avatar for a specific user only if it didn't already do so on the same day.

When users change avatar and send a presence stanza with `vcard-temp:x:update`, I think we should force the update on converse side.

Edit: Just to add that it should - I'm not saying I know all the underlyings things going on in converse - be as easy as adding a `true` in the onAvatarChanged function of the model
```javascript
if (hash && vcard.get('image_hash') !== hash) {
    _converse.api.vcard.update(vcard, true); // <- here
}
```

Lynesth:
A very simple change that fixes #1334.
Tested it for some time now, didn't notice any issue.

### DIFF
--- a/src/headless/converse-muc.js
+++ b/src/headless/converse-muc.js
@@ -1138,7 +1138,7 @@ converse.plugins.add('converse-muc', {
 
                 _.forEach(_.filter(vcards, undefined), (vcard) => {
                     if (hash && vcard.get('image_hash') !== hash) {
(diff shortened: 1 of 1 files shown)


jcbrand:
Thanks!

Issue #1337, opened 2018-11-20
send_chat_state_notifications doesn't work in MUCs


Lynesth:
Chat state notifications are sent in MUCs without any consideration for the `send_chat_state_notifications` option.

I also think that `show_chatstate_notifications` and `send_chat_state_notifications` should be written the same way for consistency.

jcbrand:
Renaming is done in #1403

PR #1397, opened 2019-01-05
Fix send_chat_state_notifications option for MUCs

Lynesth:
Fixes half of #1337.


jcbrand:
Thanks

jcbrand:
Next time, please also update CHANGES.md

PR #1408, opened 2019-01-12
new config option "roomconfig_whitelist"

ChaosKid42:
I'd like to be able to limit the amount of room config options that are presented to the user. Therefore I created a new config option `roomconfig_whitelist`.


licaon-kter:
Typo in `empty`

PR #1412, opened 2019-01-14
muc moderator commands can be disabled selectively

ChaosKid42:
I'd like to be able to selectively disable some muc moderator commands.

### DIFF
--- a/src/converse-muc-views.js
+++ b/src/converse-muc-views.js
@@ -854,12 +854,18 @@ converse.plugins.add('converse-muc-views', {
             },
 
             parseMessageForCommands (text) {
-                if (_converse.muc_disable_moderator_commands) {
+                if (_converse.muc_disable_moderator_commands &&
+                        !_.isArray(_converse.muc_disable_moderator_commands)) {
                     return _converse.ChatBoxView.prototype.parseMessageForCommands.apply(this, arguments);
                 }
                 const match = text.replace(/^\s*/, "").match(/^\/(.*?)(?: (.*))?$/) || [false, '', ''],
                       args = match[2] && match[2].splitOnce(' ').filter(s => s) || [],
-                      command = match[1].toLowerCase();
+                      command = match[1].toLowerCase(),
+                      disabled_commands = _.isArray(_converse.muc_disable_moderator_commands) ?
+                        _converse.muc_disable_moderator_commands : [];
+                if (_.includes(disabled_commands, command)) {
+                    return false;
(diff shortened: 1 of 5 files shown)


src/converse-muc-views.js
```diff
@@ -853,12 +853,15 @@ converse.plugins.add('converse-muc-views', {
             },
 
             parseMessageForCommands (text) {
-                if (_converse.muc_disable_moderator_commands) {
+                if (_converse.muc_disable_moderator_commands && _converse.muc_disable_moderator_commands.constructor !== Array) {
```
jcbrand: Please use [_.isArray](https://lodash.com/docs/4.17.11#isArray) here instead and wrap the line.

src/converse-muc-views.js
```diff
@@ -853,12 +853,15 @@ converse.plugins.add('converse-muc-views', {
...
                     return _converse.ChatBoxView.prototype.parseMessageForCommands.apply(this, arguments);
                 }
                 const match = text.replace(/^\s*/, "").match(/^\/(.*?)(?: (.*))?$/) || [false, '', ''],
                       args = match[2] && match[2].splitOnce(' ').filter(s => s) || [],
                       command = match[1].toLowerCase();
+                if (_.includes(_converse.muc_disable_moderator_commands, command)) {
```
jcbrand: It's not guaranteed that `_converse.muc_disable_moderator_commands` is an array here.

I'd rather do something like this:

```const disabled_commands = _.isArray(_converse.muc_disable_moderator_commands) : _converse.muc_disable_moderator_commands ? [];```

And then use that here and further down below. Then you don't need to check if it's an array there either.

ChaosKid42:
@jcbrand: Thanks for your review. That was really helpful and easy to implement.

jcbrand:
Hi @ChaosKid42 

Looking good. Just one more thing please, and that is a test for this usecase.

It should be fairly quick and easy to do, you can copy paste [this test](https://github.com/conversejs/converse.js/blob/master/spec/muc.js#L2546) and then just pass in a value for `muc_disable_moderator_commands` and update the assertions.

The 3rd parameter of `initConverseWithPromises` takes the configuration object (that's usually passed to `converse.initialize`), so that's where you pass in the value.

Thanks!

ChaosKid42:
@jcbrand Done. Really quite easy to do.

jcbrand:
Thanks! :)

PR #1413, opened 2019-01-14
fix moderator commands that change affiliation

ChaosKid42:
I believe it should be possible to change affiliation of non-occupants. Currently converse rejects this. This tries to fix it.

### DIFF
--- a/src/converse-muc-views.js
+++ b/src/converse-muc-views.js
@@ -35,7 +35,7 @@ import xss from "xss";
 
 const { Backbone, Promise, Strophe, b64_sha1, moment, f, sizzle, _, $build, $iq, $msg, $pres } = converse.env;
 const u = converse.env.utils;
-
+const AFFILIATION_CHANGE_COMANDS = ['admin', 'ban', 'owner', 'member', 'revoke'];
 
 converse.plugins.add('converse-muc-views', {
     /* Dependencies are other plugins which might be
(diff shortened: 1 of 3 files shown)


jcbrand:
This will only work if you provide a valid JID instead of just a nickname right?

I think you'll need to do some validation that it's really a JID being passed in.
You can use the [isValidJID](https://github.com/conversejs/converse.js/blob/master/src/headless/utils/core.js#L47) utility method.

ChaosKid42:
@jcbrand Ok. Added the check.

src/converse-muc-views.js
```diff
@@ -841,7 +841,8 @@ converse.plugins.add('converse-muc-views', {
                     );
                     return false;
                 }
-                if (!this.model.occupants.findWhere({'nick': args[0]}) && !this.model.occupants.findWhere({'jid': args[0]})) {
+                if (!(_.includes(['admin', 'ban', 'owner', 'member', 'revoke'], command) && u.isValidJID(args[0])) &&
```
jcbrand: Please create a `const AFFILIATION_CHANGE_COMANDS` at the top of the file which is set to these commands and then use that here.

Would be nice if you also wrapped the line.

And then please rebase again, so that I can rebase from github.

Thanks.

ChaosKid42: @jcbrand Done.

ChaosKid42:
@jcbrand Friendlly reminder: Do you still plan to merge this PR?

jcbrand:
Yes, but I first need to figure out why the tests on Travis are failing. They're all green on my own machine.

ChaosKid42:
Thanks for fixing the tests. This PR now passes the checks on travis.

PR #1418, opened 2019-01-16
Fix #1375

### LINKED ISSUE #1375: Adding version in released asset filenames
Hi,

I've just packaged your project in Buildroot (see buildroot.org for more details). The only issue I have is the absence of version in the asset filenames you release in github. I suggest you add -version for all of them. For example:
https://github.com/conversejs/converse.js/releases/download/v4.0.7/converse-4.0.7.min.css
https://github.com/conversejs/converse.js/releases/download/v4.0.7/converse-4.0.7.min.js
https://github.com/conversejs/converse.js/releases/download/v4.0.7/webfonts-4.0.7.zip

Otherwise, the absence of such version in the filename makes it difficult to identify the version of a downloaded file. Thus, if I bump the version of this project in buildroot, the download infrastructure will think it has already downloaded the file and will not try to download the new version, because it has the same name.

Thanks. Best,

joudinet:
- Fix #1375 by adding the version to file names in the release rule.
- Also simplify the sed commands in the same Makefile rule and fix the first sed command that was adding a spurious space character in front of _converse_VERSION_NAME.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,7 @@ UGLIFYJS		?= node_modules/.bin/uglifyjs
 
 # Internal variables.
 ALLSPHINXOPTS   = -d $(BUILDDIR)/doctrees $(PAPEROPT_$(PAPER)) $(SPHINXOPTS) ./docs/source
+VERSION_FORMAT	= [0-9]+\.[0-9]+\.[0-9]+
 
 .PHONY: all
 all: dev dist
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -84,18 +84,15 @@ po2json:
...
-	$(SED) -ri "s/cdn.conversejs.org\/[0-9]+\.[0-9]+\.[0-9]+/cdn.conversejs.org\/$(VERSION)/" docs/source/quickstart.rst
+	$(SED) -i '/^_converse.VERSION_NAME =/s/=.*/= "v$(VERSION)";/' src/headless/converse-core.js
+	$(SED) -i '/Version:/s/:.*/: $(VERSION)/' COPYRIGHT
+	$(SED) -i '/Project-Id-Version:/s/:.*/: Converse.js $(VERSION)\n"/' locale/converse.pot
+	$(SED) -i '/"version":/s/:.*/: "$(VERSION)",/' package.json
+	$(SED) -i '/"version":/s/:.*/: "$(VERSION)",/' src/headless/package.json
+	$(SED) -ri 's/--package-version=[0-9]\+\.[0-9]\+\.[0-9]\+/--package-version=$(VERSION)/' Makefile
+	$(SED) -i -e "/version =/s/=.*/= '$(VERSION)'/" -e "/release =/s/=.*/= '$(VERSION)'/" docs/source/conf.py
+	$(SED) -i "s/(Unreleased)/`date +(%Y-%m-%d)`/" CHANGES.md
+	$(SED) -ri 's,cdn.conversejs.org/[0-9]+\.[0-9]+\.[0-9]+,cdn.conversejs.org/$(VERSION),' docs/source/quickstart.rst
```
jcbrand: Hi @joudinet, this looks much cleaner thanks, but I don't know the regex syntax. Would you mind explaining how it works?

Looks like you're first matching `Version:`, and then using `s` to designate substitution of `:*` with `: $(VERSION)`.

But the last bit doesn't seems to be now a much looser regex than before, or am I misunderstanding how it works?

Makefile
```diff
@@ -84,22 +84,24 @@ po2json:
...
-	$(SED) -ri s/\"version\":\ \"[0-9]\+\.[0-9]\+\.[0-9]\+\"/\"version\":\ \"$(VERSION)\"/ src/headless/package.json
-	$(SED) -ri s/--package-version=[0-9]\+\.[0-9]\+\.[0-9]\+/--package-version=$(VERSION)/ Makefile
-	$(SED) -ri s/v[0-9]\+\.[0-9]\+\.[0-9]\+\.zip/v$(VERSION)\.zip/ index.html
-	$(SED) -ri s/v[0-9]\+\.[0-9]\+\.[0-9]\+\.tar\.gz/v$(VERSION)\.tar\.gz/ index.html
-	$(SED) -ri s/version\ =\ \'[0-9]\+\.[0-9]\+\.[0-9]\+\'/version\ =\ \'$(VERSION)\'/ docs/source/conf.py
-	$(SED) -ri s/release\ =\ \'[0-9]\+\.[0-9]\+\.[0-9]\+\'/release\ =\ \'$(VERSION)\'/ docs/source/conf.py
-	$(SED) -ri "s/(Unreleased)/`date +%Y-%m-%d`/" CHANGES.md
-	$(SED) -ri "s/cdn.conversejs.org\/[0-9]+\.[0-9]+\.[0-9]+/cdn.conversejs.org\/$(VERSION)/" docs/source/quickstart.rst
+	$(SED) -i '/^_converse.VERSION_NAME =/s/=.*/= "v$(VERSION)";/' src/headless/converse-core.js
+	$(SED) -i '/Version:/s/:.*/: $(VERSION)/' COPYRIGHT
```
jcbrand: Hi @joudinet, this looks much cleaner thanks, but I don't know the regex syntax. Would you mind explaining how it works?

Looks like you're first matching `Version:`, and then using `s` to designate substitution of `:.*` with `: $(VERSION)`.

But the substitution match `:*` is now much looser than before, or am I misunderstanding how it works?

joudinet: Yes, it first matches lines that match /regexp/ and applies on each of them the command s/from/to/ to replace the portion matched by 'from' with 'to'. This avoids to repeat yourself in the replacement part.
I looked at instances were these sed commands are applied, an I loose the match regexp such that it still matches the correct portion to be replaced without matching any other part of the document. I don't see the point in making the match part over complex. For example, the ' Version:' in the COPYRIGHT replaces the end of the line after the colon by ': $(VERSION)', since there is nothing else present on this line after the version.
Feel free to harden them if you're afraid such commands might match something they shouldn't replace in a future version.

I also removed the -r (i.e. -E or --regex-extended) option whenever it was not needed. For example, it was confusing in the sed command for CHANGES.md. there were parenthesis in the match part but no \1 in the replacement part to insert the matched expression. I think this command wanted to actually match the parenthesis characters. So, I actually harden the match regexp for this sed command.

Makefile
```diff
@@ -84,22 +85,28 @@ po2json:
...
+	$(SED) -ri 's/--package-version=$(VERSION_FORMAT)/--package-version=$(VERSION)/' Makefile
+	$(SED) -i -e "/version =/s/=.*/= '$(VERSION)'/" -e "/release =/s/=.*/= '$(VERSION)'/" docs/source/conf.py
+	$(SED) -i "s/[Uu]nreleased/`date +%Y-%m-%d`/" CHANGES.md
+	$(SED) -ri 's,cdn.conversejs.org/$(VERSION_FORMAT),cdn.conversejs.org/$(VERSION),' docs/source/quickstart.rst
 	make pot
 	make po
 	make po2json
 	make build
+	mkdir -p 'converse-assets-$(VERSION)'
+	install -D dist/converse.min.js 'converse-assets-$(VERSION)/converse.js'
```
jcbrand: GNU `install` isn't available on MacOS and needs to be installed first and then it's only available as `ginstall`.

So in the Makefile `install` needs to be made configurable just like `sed` currently is, so that you can specify the `install` command to be used when running `make`.

Makefile
```diff
@@ -84,22 +85,28 @@ po2json:
...
+	install -D css/converse.min.css \
+		'converse-assets-$(VERSION)/css/converse.css'
+	unzip -d 'converse-assets-$(VERSION)' css/webfonts.zip
+	unzip -d 'converse-assets-$(VERSION)' locale.zip
+	unzip -d 'converse-assets-$(VERSION)' sounds.zip
```
jcbrand: Instead of unzipping here, you can just copy over the directories, right?

We can then remove the recipes for `css/webfonts.zip`, `sounds.zip` and `locale.zip`. I think if we have `converse-assets-$(VERSION).zip` then the need for the other zip files goes away.

Makefile
jcbrand: Thanks for the explanation, I think it's good as it is now.

jcbrand:
Thanks!

PR #1420, opened 2019-01-18
Add ARIA live region on chat-content

linkmauve:
This asks screen readers to read every message received, and is mandatory for blind users to get notified of new messages.


jcbrand:
Thanks!

PR #1421, opened 2019-01-21
fix direct invite for membersonly room

ChaosKid42:
Direct invite for a members only room must make sure that the invited user is added to the member list. This is currently not working.

This PR fixes it.


jcbrand:
Thanks!

PR #1436, opened 2019-02-06
Prevent duplicate messages on MUC join

### LINKED ISSUE #1414: Duplicate messages when first entering a group chat
When I first enter a group chat I receive all (historic) messages twice. Live incoming messages are fine. Scrolling further up in the history than what is initially loaded I also see each message only once.

So I suspect that on join it is not setting `maxchars` or `maxstanzas` to 0 and getting messages from both MAM and 'old' style muc history. I haven’t looked at debug output though to confirm this; so sorry if this ends up being something else and I’m being misleading.

It *feels* like a regression. I don’t remember seeing this on 4.0

laszlovl:
Fixes #1414

This was broken in https://github.com/conversejs/converse.js/commit/a4d608dcdfe9c0d33ad65279369b0e6bd4a97def

Thanks for the hint @iNPUTmice

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@
 - #1422 Resurrect the `muc_show_join_leave` option
 - #1412 muc moderator commands can be disabled selectively by config
 - #1413 fix moderator commands that change affiliation
+- #1414 Prevent duplicate messages on MUC join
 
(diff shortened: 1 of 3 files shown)


jcbrand:
This was such a bad and embarrassing bug, thanks for fixing it!

PR #1438, opened 2019-02-09
Update contact nickname when receive a roster push

aberenguel:
When we change the roster item name in the server, the client receives a roster push but does not update the respective contact. 
This PR fix that.


jcbrand:
Thanks @aberenguel.

There is now a failing test due to your change, can you please update [spec/protocol](https://github.com/conversejs/converse.js/blob/master/spec/protocol.js#L297) to fix the test?

You can put in a new value for `name` on line 273 and then check that value on line 297.

jcbrand:
I've merged manually and fixed the test.

PR #1441, opened 2019-02-11
Dont run fetchRosterContacts async

### LINKED ISSUE #1405: Status of contacts list are not displayed properly
Using last version of converse 4.1.0 (previous version 4.0.6 had the same problem), when I connect to converse in full screen, the status of the contacts on the left are not properly displayed, some of them appear as offline where they are online.

I don't mark the check for "trusted device" when I connect to converse in the login screen.
The problem is not always reproducible, sometimes I had to go offline and connect again.

laszlovl:
Fixes #1405

The problem I observed was as follows: upon login, your roster as well as your contacts' presence are contained in a single BOSH response. Converse would process the roster (`fetchRosterContacts`) async and move on to processing the presence in the mean time. Because of this, in `presenceHandler` the roster contacts would not exist yet (`contact = this.get(bare_jid)` returning NULL) causing the presence to be ignored. As a result, the contact would never show online until logging off & on.

This change makes sure we wait for `fetchRosterContacts` to be completed. There are a few possible ways to go about this, but I think this works well. Correct me if I'm wrong :)

### DIFF
--- a/src/headless/converse-roster.js
+++ b/src/headless/converse-roster.js
@@ -407,7 +407,7 @@ converse.plugins.add('converse-roster', {
                 if (collection.length === 0 ||
                         (this.rosterVersioningSupported() && !_converse.session.get('roster_fetched'))) {
                     _converse.send_initial_presence = true;
-                    _converse.roster.fetchFromServer();
+                    await _converse.roster.fetchFromServer();
                 } else {
                     _converse.emit('cachedRoster', collection);
                 }
--- a/src/headless/converse-roster.js
+++ b/src/headless/converse-roster.js
@@ -88,10 +88,14 @@ converse.plugins.add('converse-roster', {
                 }
             } else {
                 try {
-                    await _converse.rostergroups.fetchRosterGroups().then(() => {
-                        _converse.emit('rosterGroupsFetched');
-                        return _converse.roster.fetchRosterContacts();
-                    });
+                    /* Make sure not to run fetchRosterContacts async, since we need
+                     * the contacts to exist before processing contacts presence,
+                     * which might come in the same BOSH request.
+                     */
+                    await _converse.rostergroups.fetchRosterGroups();
+                    _converse.emit('rosterGroupsFetched');
+
+                    await _converse.roster.fetchRosterContacts();
                     _converse.emit('rosterContactsFetched');
                 } catch (reason) {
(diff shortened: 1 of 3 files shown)


src/headless/converse-roster.js
```diff
@@ -88,10 +88,14 @@ converse.plugins.add('converse-roster', {
                 }
             } else {
                 try {
-                    await _converse.rostergroups.fetchRosterGroups().then(() => {
-                        _converse.emit('rosterGroupsFetched');
-                        return _converse.roster.fetchRosterContacts();
-                    });
+                    /* Make sure not to run fetchRosterContacts async, since we need
```
jcbrand: eee

jcbrand: One way to test for this, is to check that `rosterGroupsFetched` is called *before* `rosterContacts` fetched.

I'm happy to help you with writing such a test.

laszlovl:
> Thanks @laszlovl :)

You're welcome!

Unfortunately after some more hours of usage, I'm afraid this patch doesn't fully fix the problem yet. `fetchRosterContacts` still calls the async function `fetchFromServer` which does the actual heavy lifting.

I think the following fixes it fully once and for all - but would this still be the right way to go? I still have to wrap my head around some of the async patterns here; if we `await` on every invocation does it still make sense to make it `async` in the first place?

```diff
--- a/src/headless/converse-roster.js
+++ b/src/headless/converse-roster.js
@@ -407,7 +409,7 @@ converse.plugins.add('converse-roster', {
                 if (collection.length === 0 ||
                         (this.rosterVersioningSupported() && !_converse.session.get('roster_fetched'))) {
                     _converse.send_initial_presence = true;
-                    _converse.roster.fetchFromServer();
+                    await _converse.roster.fetchFromServer();
```

jcbrand:
> if we `await` on every invocation does it still make sense to make it `async` in the first place?

Yes, otherwise the function wouldn't wait for the promises to resolve before executing the next line.

src/headless/converse-roster.js
laszlovl: Sure, I'll have a go at that this week. :)

jcbrand:
@laszlovl I had some time, so I manually merged your fix and added a test.

laszlovl:
@jcbrand Excellent! :)

PR #1451, opened 2019-02-21
Don't leak the real JIDs of participants when using mentions in semi-anonymous rooms.

lumi-me-not:
Converse is leaking real JIDs of participants in semi-anonymous MUCs when a moderator mentions anyone, it also will sometimes not include a "uri" attribute, even though the XEP states that this is required.

I've fixed both of these issues, though I'm not sure if this is a good way of doing so.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,8 @@
 - Bugfix. Prevent duplicate messages by comparing MAM archive id to XEP-0359 stanza ids.
 - Bugfix. Open groupchats not shown when logging in after disconnection.
 - #1406: `TypeError: e.devicelists is undefined` when unchecking the "trusted device" checkbox
+- Bugfix. Don't send real jids in mentions in a semi-anonymous or anonymous room when we are a moderator.
+- Bugfix. Always set the "uri" field in reference tags, as required by the XEP.
 
 ## 4.1.1 (2019-02-18)
 
--- a/spec/messages.js
+++ b/spec/messages.js
@@ -2800,6 +2800,34 @@
                 done();
             }));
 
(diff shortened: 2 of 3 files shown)


jcbrand:
Thank you for this PR @lumi-me-not 

I wrote to the standards list about this PR. 
https://mail.jabber.org/pipermail/standards/2019-March/035856.html

@iNPUTmice prefers to use real JIDs as far as possible, because nicknames aren't stable in MUCs. Anybody can take a nickname and thereby impersonate other users, or receive notifications for messages that were meant for someone else who used that nickname before them.

There is a workarounds for this, for example registering nicknames with MUCs, which @mwild1 and I worked on in the context of Prosody and Converse, see [auto_register_muc_nickname](https://conversejs.org/docs/html/configuration.html#auto-register-muc-nickname), but it's not widely deployed.

I think in light of all this, this change should be made configurable, with the current behavior being the default. The configuration setting can be called something like `muc_references' with the default value being `jid` and the other option being `nickname`.

(On a different, but related note, we should add validation for configuration settings to check that valid values are supplied).

Setting `auto_register_muc_nickname` to `true` and `muc_references` to `nickname` would solve the problem of unstable MUC nicknames (if the XMPP server supports [MUC nickname registrations](https://xmpp.org/extensions/xep-0045.html#register)).

mwild1:
Wondering aloud if we could just add stable ids to MUC participants...

jcbrand:
> Wondering aloud if we could just add stable ids to MUC participants...

It's probably worthwhile revisiting this MR now that we have stable MUC participant ids via XEP-0421 (which Converse already supports)

PR #1453, opened 2019-02-22
Makefile: Keep .min extension and add sourcemap to the assets archive

joudinet:


jcbrand:
Very timely, thanks :)

PR #1461, opened 2019-02-25
Makefile: Fix .min.css filename and include headless variant in assets

joudinet:
Also include the unminified versions for people who wants to modify
them.


jcbrand:
Thanks!

PR #1470, opened 2019-02-28
WIP: Add support for IndexedDB

jcbrand:
[LocalForage](https://localforage.github.io/localForage/) is a storage wrapper which provides a single localStorage-like API (but async) for localStorage, sessionStorage and IndexedDB.

The switch from synchronous storage to async storage makes this change very tricky and invasive. There are various places in the code where the implicit assumption is that saving to the browser's storage is synchronous.

TODO:
* Since saving might now be more expensive, when calling `.save` with a single attribute being set, check whether a new value is being set, otherwise don't save.


jcbrand:
Closed in favour of #1724

PR #1477, opened 2019-03-04
Assume "text-single" as default form field type

weiss:
As [per XEP-0004][1], the default `type` of data form fields is `text-single`, so a missing `type` attribute should not be treated differently.

This fixes handling of CAPTCHAs offered by ejabberd, which explicitly sets `type='text-single'`for [the `var='ocr'` field][2].

(For review, it might help to [ignore whitespace changes in the diff][3].)

[1]: https://xmpp.org/extensions/xep-0004.html#protocol-fieldtypes
[2]: https://xmpp.org/extensions/xep-0158.html
[3]: https://github.com/conversejs/converse.js/pull/1477/files?utf8=✓&diff=unified&w=1#diff-08112c1ee3f6b7f0cdbbae41576b6552

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 - Upgrade to Backbone 1.4.0
 - Fix "flashing" of roster filter when you have less than 5 roster contacts.
+- Fix handling of CAPTCHAs offered by ejabberd.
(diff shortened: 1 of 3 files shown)


jcbrand:
Thanks!

PR #1487, opened 2019-03-08
make it configurable whether to respect autojoin

ChaosKid42:
I'd like to be able to turn off converse.js' behaviour of autojoining MUCs bases on their bookmarks. This PR adds a new config option `muc_respect_autojoin` which allows this.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -17,6 +17,7 @@
 - #1437: List of groupchats in modal doesn't scroll
 - #1457: Wrong tooltip shown for "unbookmark" icon
 - #1479: Allow file upload by drag & drop also in MUCs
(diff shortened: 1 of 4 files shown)


CHANGES.md
```diff
@@ -17,6 +17,7 @@
 - #1437: List of groupchats in modal doesn't scroll
 - #1457: Wrong tooltip shown for "unbookmark" icon
 - #1479: Allow file upload by drag & drop also in MUCs
+- #1487: New config option `muc_respect_autojoin`
```
jcbrand: You can now already set it to the URL that it will be when it's released. Take a look at the URLs of other config options. You generally just have to replace underscores with dashes at the end of the URL.

ChaosKid42: Nice. Changed the PR accordingly.

PR #1489, opened 2019-03-09
Allow `http:` image URLs if Converse.js is loaded on an insecure origin.

meitar:
Currently, in order for Converse.js to recognize a pasted URL as an
image, it must be an URL whose protocol is `https:` (`https` in the
URI.js library's notation). This is sensible, but means that any
non-HTTPS image URL is not recognized as a valid URL (and thus will not
be rendered inline, even if `show_images_inline` is set to `true`).

It is important to always check for HTTPS URLs when in a secure context
(i.e., the initial page load was requested via HTTPS) in order to ensure
that non-secured content does not mix with secured content. However, the
inverse is not true: if the original page was loaded over HTTP, then
enforcing HTTPS for images adds arguably no meaningful protection while
also breaking the `show_images_inline` feature for the edge cases where
Converse.js is deployed without HTTPS and a user pastes an HTTP URL.

This patch changes the behavior of the `isImageURL` method such that the
requirement for the pasted URL's protocol to be `https:` is enforced
only when the `window.location.protocol` itself is also `https:`. By
doing this, we ensure that secure origins (i.e., when Converse.js is
loaded over HTTPS initially) are still secured and cannot have non-HTTPS
content introduced to the page via a pasted non-HTTPS URL, however it
also allows non-HTTPS origins to render both HTTP and HTTPS image URLs.

I did not create a test because I did not find existing tests for `isImageURL()` and assume that this change is therefore minor enough not to require the addition of a new test spec?

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@
 - Allow setting of debug mode via URL with `/#converse?debug=true`
 - New config setting [locked_muc_domain](https://conversejs.org/docs/html/configuration.html#locked-muc-domain)
 - New config setting [show_client_info](https://conversejs.org/docs/html/configuration.html#show-client-info)
+- Render inline images served over HTTP if Converse itself was loaded on an unsecured (HTTP) page.
 - #1149: With `xhr_user_search_url`, contact requests are not being sent out
 - #1213: Switch roster filter input and icons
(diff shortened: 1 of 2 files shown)


jcbrand:
Thanks!

PR #1501, opened 2019-03-18
Don't prompt for reason if auto_join_on_invite

ChaosKid42:
I believe Converse shouldn't prompt for a reason if `auto_join_on_invite` is `true` because the inivted user won't see it anyway.


jcbrand:
Thanks

PR #1507, opened 2019-03-22
fix LMC with Conversations

ChaosKid42:
If I send a correction to Conversations the message will be rendered as a new message by Conversations. C seems to expect the message-id and the origin-id to be identical (I'm not 100% sure about it, but seems to work). This PR tries to fix it for conversejs.


jcbrand:
Thanks.

Unfortunately it seems like the tests have become much more flaky on Travis the last while :(

Issue #1508, opened 2019-03-22
Minimized bookmarked chatboxes are always maximized after page reload


fabiomontefuscolo:
https://github.com/conversejs/converse.js/blob/10da92ec5b51a3d2275e202494a234d7e3ab881a/src/converse-bookmarks.js#L256

Should it be? 
```javascript
                    if (!(groupchat.get('hidden') || groupchat.get('minimized'))) {
                        groupchat.trigger('show');
                    }
```

jcbrand:
Yes, I think that's the fix for the bug. Will you please make a PR?

guusdk:
I don't claim to understand the issue or fix, but I've happily provided a PR with Fabio's changes in https://github.com/conversejs/converse.js/pull/1510

PR #1510, opened 2019-03-25
Minimized bookmarked chatboxes should not be always maximized after page reload

guusdk:
This is a fix for https://github.com/conversejs/converse.js/issues/1508


src/converse-bookmarks.js
```diff
@@ -253,7 +253,7 @@ converse.plugins.add('converse-bookmarks', {
             openBookmarkedRoom (bookmark) {
                 if ( _converse.muc_respect_autojoin && bookmark.get('autojoin')) {
                     const groupchat = _converse.api.rooms.create(bookmark.get('jid'), bookmark.get('nick'));
-                    if (!groupchat.get('hidden')) {
+                    if (!(groupchat.get('hidden') || groupchat.get('minimized'))) {
```
jcbrand: For better readability, would you please change this from `!(x || y)` to `!x && !y`?

guusdk: done

PR #1512, opened 2019-03-26
Allow manual entry of jid with xhr_user_search_url

ChaosKid42:
I'd like to be able to manually entry a JID even if using xhr_user_search_url.

Therfore I propose to show the JID field.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -32,6 +32,7 @@
 - #1501: Don't prompt for a reason if [auto_join_on_invite](https://conversejs.org/docs/html/configuration.html#auto-join-on-invite) is `true`
 - #1507: Make message id and origin-id identical in order to fix LMC with Conversations
 - #1508: Minimized bookmarked chatboxes should not be always maximized after page reload.
+- #1512: Allow manual entry of jid even with [xhr_user_search_url](https://conversejs.org/docs/html/configuration.html#xhr-user-search-url)
 
 ## 4.1.2 (2019-02-22)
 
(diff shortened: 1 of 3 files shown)


jcbrand:
AFAIK this isn't necessary, you can just type the JID in the existing input. Or are you prevented from submitting the form or something?

ChaosKid42:
It's not working because the jid field is empty.

jcbrand:
OK I understand now.

This definitely needs to be configurable though. This config option is used in commercial systems where you don't want people to ever be exposed to JIDs.

ChaosKid42:
Ok. What about `xhr_user_search_show_jid_field` which defaults to `false`?
I'd probably be able to add it to the PR tomorrow.

jcbrand:
`xhr_user_search_show_jid_field` is quite long and ugly, but I can't think of anything better right now.

ChaosKid42:
@jcbrand: Done

jcbrand:
@ChaosKid42: I have a counter-proposal for you.

Instead of showing the JID field, what if we instead get it to work when you enter a JID in the `input[name="name"]` (i.e. auto-completed) field. What we can do, is we can add a `change` event listener which updates the `input[name="jid"]` field once the `input[name="name"]` field has changed value. This is independent of auto-completion, so it'll also work when the user just manually enters a JID.

If you want users to be able to see the JIDs of contacts, we can try show them in the auto-complete list (in small grey text next to or under the name).

We would then still need to handle the case where you don't want to show JIDs at all. However, I think this might be a bit nicer UI. Having two input fields, and having one being updated by the other, but also being able to manually enter something there is likely to confuse users.

ChaosKid42:
@jcbrand: My counter-counter-proposal would be to have both options because your solution does not allow the user to manually enter a contact's nickname. You need two fields for that.
I tried to implement it. However, listening to the change event of the `name` field can als be triggered in case you select a value from auto-completion. This is bad and I can't figure out how to prevent it.

ChaosKid42:
@jcbrand: I believe this works now.

jcbrand:
Hi @ChaosKid42 

As it happens, I also had to work on the "Add Contact" modal for the `xhr_user_search_url` case. In particular, I needed to disable autocomplete suggestions.

However, I then took the time to remove Awesomplete and instead use `_converse.AutoComplete`.

I've had some time to think about your use-case. Generally, I get concerned about having so many different configuration settings and look for ways to achieve customizability without introducing many new ones. In light of this, I don't think we need `xhr_user_search_show_jid_field`. Instead, we can just hide the JID field via CSS. I think that's sufficient.

So, I think your use-case can be the default use-case, and if people don't want to show the JID input, they can simply hide it with CSS. This should make your PR much simpler.

Sorry for the confusion and for causing conflict errors and more work for you. I've been meaning to get rid of Awesomplete for quite a while now, saw and opportunity and took it.

ChaosKid42:
@jcbrand: No problem. The PR actually becomes trivial :-)

Issue #1548, opened 2019-05-01
Missed MAM messages beyond a number of 50 are permanently lost


laszlovl:
This is similar to #1534, but I think it's different enough to warrant a separate issue.

The problem is triggered as follows:

* User enters a MUC
* Someone writes message "1" in the MUC
* User closes his browser
* Someone writes messages "2" through "100" in the MUC
* User re-opens his browser
* Converse will now use MAM to request 50 messages after "1", e.g. messages "2" through "51". Because MAM paging is not implemented, Converse will not request additional messages after "51".
* Someone writes message "101" in the MUC

Messages "52" through "100" are now permanently lost. Even if the user refreshes/rejoins, Converse will only query MAM for messages created after message "101".

I propose the short-term fix of actually implementing paging in `fetchNewestMessages`. Thus, after requesting messages "2" through "51", Converse should query for additional MAM messages created after "51".

In the long term, this might not be very sustainable though. In case of a long absence from a MUC, this method will query for **all** messages left since the user was last online, which could be quite a lot in case of an active MUC. Ideally, Converse would request only the 50 most recent messages then page upwards/backwards as needed. However, implementing that in the current architecture would cause even more missed messages because of #1534 (Converse would request messages "50" through "100", but then messages "1" through "49" would be permanently lost, as it's not aware that there might be more unretrieved messages between "1" and "50")

PR #1549, opened 2019-05-01
Implement MAM paging to prevent losing messages

laszlovl:
Implement a fix for #1548. Please see that ticket and the commit messages for detailed explanations.

The way I handled https://github.com/laszlovl/converse.js/commit/56f577f6b226733734d91e7fb767ddd1dba131b4#diff-8279dbb754eb81e3e0ce2e30070bc8adR52067 isn't very elegant, please let me know if there's a better way to go about this.

### DIFF
--- a/src/converse-mam-views.js
+++ b/src/converse-mam-views.js
@@ -40,6 +40,7 @@ converse.plugins.add('converse-mam-views', {
                  * the last archived message in our local cache.
                  */
                 if (this.disable_mam) { return; }
+
                 const { _converse } = this.__super__,
                       most_recent_msg = u.getMostRecentMessage(this.model);
 
@@ -47,10 +48,22 @@ converse.plugins.add('converse-mam-views', {
                     this.fetchArchivedMessages();
                 } else {
                     const stanza_id = most_recent_msg.get(`stanza_id ${this.model.get('jid')}`);
+                    const that = this;
+
+                    const success = function(messages) {
+                        /* If we received exactly the maximum number of messages, there's a good
+                        chance more messages are available. Let's run this method again to check. */
+                        if (_.size(messages) == _converse.archived_messages_page_size) {
+                            /* must use setTimeout here; otherwise the MAM messages we added just now are
+                            not available yet in the next iteration of fetchNewestMessages() */
+                            setTimeout(that.fetchNewestMessages.bind(that), 1);
+                        }
+                    };
+
                     if (stanza_id) {
-                        this.fetchArchivedMessages({'after': stanza_id});
+                        this.fetchArchivedMessages({'after': stanza_id}).then(success);
                     } else {
-                        this.fetchArchivedMessages({'start': most_recent_msg.get('time')});
+                        this.fetchArchivedMessages({'start': most_recent_msg.get('time')}).then(success);
                     }
                 }
             },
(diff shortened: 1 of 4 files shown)


src/converse-mam-views.js
```diff
@@ -40,17 +40,30 @@ converse.plugins.add('converse-mam-views', {
...
                     this.fetchArchivedMessages();
                 } else {
                     const stanza_id = most_recent_msg.get(`stanza_id ${this.model.get('jid')}`);
+                    const that = this;
+
+                    const success = function(messages) {
+                        /* If we received exactly the maximum number of messages, there's a good
+                        chance more messages are available. Let's run this method again to check. */
+                        if (_.size(messages) == _converse.archived_messages_page_size) {
+                            /* must use setTimeout here; otherwise the MAM messages we added just now are
```
jcbrand: The XMPP server will inform you when the results are paged, so it should be necessary to do this here.
See the (relevant part of the XEP](https://xmpp.org/extensions/xep-0313.html#query-limit).

Specifically you should check for a `<set xmlns='http://jabber.org/protocol/rsm>` element and then use that to determine whether you need to query for more pages.

src/converse-mam-views.js
```diff
@@ -100,26 +113,34 @@ converse.plugins.add('converse-mam-views', {
...
-                );
+
+                return new Promise((resolve, reject) => {
+                    _converse.api.archive.query(
+                        Object.assign({
+                            'groupchat': is_groupchat,
+                            'before': '', // Page backwards from the most recent message
+                            'max': _converse.archived_messages_page_size,
+                            'with': this.model.get('jid'),
+                        }, options),
```
jcbrand: The next release of Converse is 5.0.0, so we can break the API.

I therefore think we should let `_converse.api.archive.query` return a Promise instead of accepting callback functions. That would make the code here much cleaner and nicer.

You won't need to do `return new Promise` and can instead use `await` statements.

src/headless/converse-chatboxes.js
```diff
@@ -747,7 +747,7 @@ converse.plugins.add('converse-chatboxes', {
                     'references': this.getReferencesFromStanza(stanza),
                     'subject': _.propertyOf(stanza.querySelector('subject'))('textContent'),
                     'thread': _.propertyOf(stanza.querySelector('thread'))('textContent'),
-                    'time': delay ? delay.getAttribute('stamp') : moment().format(),
+                    'time': delay ? moment(delay.getAttribute('stamp')).format() : moment().format(),
```
jcbrand: Can you please explain again why this is necessary?

laszlovl: As discussed on the chat, but for everyone else's reference: the explanation is in the commit message https://github.com/laszlovl/converse.js/commit/96da5c954b3c83ac6c7c74865f311923748b7976

linkmauve: Wouldn’t it be better to keep the time/moment object in memory, rather than its string representation, for further manipulation from other parts of the code?

jcbrand: You can only store string key/value pairs in localStorage.

jcbrand:
@laszlovl: I've cherry-picked your second commit and then also made some more updates.

We now use `toISOString` everywhere in the app, so that we don't run into timezone issues when users move across time zones.

I also updated the `api.archive.query` API method to return a promise instead of accepting callbacks.
While doing so, I noticed that this API method already supports paging, it's explained in the method's docstring.

We should be using this API method's paging apparatus and not a `setTimeout` hack.

laszlovl:
> We should be using this API method's paging apparatus and not a `setTimeout` hack.

Cool, nice work! Do you have a timeline yet on the change to remove archived messages from browser storage at the end of the session? (I think that should negate the need for any changes here altogether)

jcbrand:
> I think that should negate the need for any changes here altogether

I'm not so sure about that. For now I'll only clear messages upon logout. You'll still need to do this kind of query when Converse.js reconnects after the session dropped.

laszlovl:
I just checked with the latest master, and I can't seem to reproduce the problem anymore.

If I kill the browser's network connection for less than 60 seconds (my server's BOSH timeout), all missed messages will be sent automatically.

If I kill the browser's network connection for more than 60 seconds, Converse will notice that it's disconnected and upon automatic reconnect it seems to clear the message archive, since only the last 50 messages are fetched by mam. Scrolling up will properly fetch the missing ones.

Is that in line with your expectations @jcbrand?

jcbrand:
It's not exactly clear to me why it should be working now. The way I understand the bug, you should still be missing messages 51 to 100, but I haven't yet completely figured out what the actual bug in the code is. I just started refactoring because the MAM code was clearly in need of updating.

If we changed the order in which MAM messages are fetched then the bug might also be solved (in an indirect way) but I don't remember doing so or whether you did.

In any case, I think it's still worthwhile to add proper paging support for MAM results. This PR however can be closed.

laszlovl:
Oh, I understand why it's working now. If all messages are cleared during a reconnect, Converse won't do a `this.fetchArchivedMessages({'after': stanza_id});` but a plain `this.fetchArchivedMessages();` and thus simply retrieve the most recent 50 messages.

But is the behavior of clearing messages on reconnect intentional and permanent?

jcbrand:
> But is the behavior of clearing messages on reconnect intentional and permanent?

Yes

laszlovl:
> > But is the behavior of clearing messages on reconnect intentional and permanent?
> 
> Yes

Cool, that'll make life easier.

The logic to account for `most_recent_msg` in `fetchNewestMessages` might now be dead code, because history will always be cleared when this function is executed?

jcbrand:
> The logic to account for most_recent_msg in fetchNewestMessages might now be dead code, because history will always be cleared when this function is executed?

Yes, I'm a bit reluctant to remove it, I wonder whether we might still need it in some cases... same with paging.

Issue #1565, opened 2019-05-10
Documentation mentions non-existent OTR support repeatedly


chee:
Hey! 😊

Your [documentation](https://github.com/conversejs/converse.js/#features) [repeatedly](https://conversejs.org/#features) [mentions](https://conversejs.org/docs/html/configuration.html?highlight=otr#whitelisted-plugins) [OTR](https://conversejs.org/docs/html/configuration.html?highlight=otr#allow-otr) [support](https://conversejs.org/docs/html/configuration.html?highlight=otr#use-otr-by-default) but the [change log](https://github.com/conversejs/converse.js/blob/master/CHANGES.md#400-2018-09-07) for 4.0 [casually mentions](https://github.com/conversejs/converse.js/blob/master/CHANGES.md#other) that OTR support has been dropped.

Would you be interested in a PR removing all of the mentions of OTR support?

Also, is there an issue anywhere that explains why the support was dropped?

Also would the v3 OTR code be a good starting point for creating an OTR plugin?

jcbrand:
Hi, a PR that removes references to OTR in the docs would be very welcome.

Support for OTR was dropped because breaking changes were introduced in the v4 release and we don't have capacity to support both OTR and OMEMO. Besides that, other major clients also dropped support for it, most notably Conversations. 

It should be possible to copy the v3 OTR code to the community plugins repo and update it to work with v4. I'd be very happy if someone were to do it. 

https://github.com/conversejs/community-plugins

PR #1567, opened 2019-05-11
Remove references to unsupported OTR encryption

chee:
Relates to issue #1565, "documentation mentions non-existent OTR support repeatedly".

Should i add an item to CHANGES.md for a doc-only change?


jcbrand:
Great, thanks!

I don't think a changelog entry is necessary.

PR #1572, opened 2019-05-14
ios mobile viewport

supun19:
ios mobile viewport was lessthan 100vh so had to redefine vh units each time


sass/_chatbox.scss
```diff
@@ -635,8 +635,10 @@
         .box-flyout {
             background-color: var(--chat-head-color);
             box-shadow: none;
-            height: var(--fullpage-chat-height);
-            min-height: calc(var(--fullpage-chat-height) / 2);
+            // height: var(--fullpage-chat-height);
+            height: calc(var(--vh, 1vh) * 100);
```
jcbrand: You need to set `--fullpage-chat-height` to `calc(var(--vh, 1vh) * 100);` in _variables.scss and then undo the edits in this file.

dev.html
```diff
@@ -35,7 +36,22 @@
...
+            // First we get the viewport height and we multiple it by 1% to get a value for a vh unit
+            const vh = window.innerHeight * 0.01;
+            // Then we set the value in the --vh custom property to the root of the document
+            document.documentElement.style.setProperty('--vh', `${vh}px`);
+            window.addEventListener('resize', () => {
+            // We execute the same script as before
+            const vh = window.innerHeight * 0.01;
+            document.documentElement.style.setProperty('--vh', `${vh}px`);
+            });
+        }, 1000);
```
jcbrand: This code needs to go into one of the core modules so that it's always available. Otherwise you'd need to remember to always add this when you call `_converse.initialize`.

Also, instead of using `setTimeout`, we should rather listen for an event that signifies that the viewport height is calculable.

The most low-level view module is converse-chatboxviews.js.

I think you can register an event handler for `chatBoxViewsInitialized` and then in that handler add the above code.

Also, you still need to add an event handler for the `resize` event so that the `vh` value gets recalculated when the viewport changes size.
See here:  https://css-tricks.com/the-trick-to-viewport-units-on-mobile/#article-header-id-2

CHANGES.md
```diff
@@ -1,7 +1,7 @@
 # Changelog
 
 ## 5.0.0 (Unreleased)
-
+- ios mobile viewport was lessthan 100vh so had to redefine vh units each time
```
jcbrand: Please fix `lessthan` and write `ios` as `iOS`. Also add the ticket number in front and make sure that it's placed in the right position (ordered by ticket number).

src/converse-chatboxviews.js
```diff
@@ -159,7 +159,10 @@ converse.plugins.add('converse-chatboxviews', {
              * @event _converse#chatBoxViewsInitialized
              * @example _converse.api.listen.on('chatBoxViewsInitialized', () => { ... });
              */
+            _converse.api.trigger('chatBoxViewPortCalculate');
+            _converse.api.trigger('chatBoxViewPortCaclulateWhenResize');
```
jcbrand: This is not correct. When the user resizes the viewport it won't trigger this custom event that you've made. You need to listen for the `resize` event, like it's shown in the article I linked to earlier.

src/converse-controlbox.js
```diff
@@ -610,8 +610,18 @@ converse.plugins.add('converse-controlbox', {
...
+        _converse.api.listen.on('chatBoxViewPortCalculate', () => {
+            const vh = window.innerHeight * 0.01;
+            document.documentElement.style.setProperty('--vh', `${vh}px`);
+        });
+        _converse.api.listen.on('chatBoxViewPortCaclulateWhenResize', () => {
+            window.addEventListener('resize', () => {
+                const vh = window.innerHeight * 0.01;
+                document.documentElement.style.setProperty('--vh', `${vh}px`);
+            });
+        });
```
jcbrand: Instead of making custom events, you can just listen for the `chatBoxViewsInitialized` event.

src/converse-controlbox.js
```diff
@@ -610,8 +610,18 @@ converse.plugins.add('converse-controlbox', {
                     }
                 }
             });
-        });
 
+        });
+        _converse.api.listen.on('chatBoxViewPortCalculate', () => {
+            const vh = window.innerHeight * 0.01;
+            document.documentElement.style.setProperty('--vh', `${vh}px`);
+        });
```
jcbrand: Instead of triggering and listening for a new event, you can just listen for `chatBoxViewsInitialized`

src/converse-controlbox.js
jcbrand: You don't need to wrap this inside `_converse.api.listen.on`.

jcbrand:
Thanks, merged locally.

PR #1573, opened 2019-05-16
fix foreach in fileupload

ChaosKid42:
For me file upload is not working in master. This seems to fix the problem


jcbrand:
Thanks!

Issue #1611, opened 2019-06-19
Clicking chat title should open info modal instead of requiring separate icon


Nyco:
# Context: solve major pain points one by one

We conducted a study:
Major pain points of Converse after a small user test series #1580
https://github.com/conversejs/converse.js/issues/1580

We have already designed a solution to the major pain points (scored 9):
Create/join channel (groupchat) #1593
https://github.com/conversejs/converse.js/issues/1593
Not delivered yet.

Now, in this current ticket, we are addressing these:
* "i"
* Channel menu
* Contact profile

![58100928-e9fae180-7bde-11e9-9db7-a431b7ae3aaf](https://user-images.githubusercontent.com/25348/59784870-c3a29180-92c3-11e9-8bbe-024a17aba9c4.png)

Thanks to feedback, we will address this in another ticket:
* Quit/exit/leave chat

# Vocabulary 

The "chat pane" is the right side of Converse (not left pane, list of conversations).
The "banner" is the infos and buttons at the top of the chat pane.

Please correct the vocabulary, so we have common wording.

# Problems

## Pain points

Users don't understand why there is:
* duplicate infos when they click in the "i" button (channels)
* a card while the avatar+name is not clickable

## Expressed needs

Users want to click on the name, JID and avatar:
* one-to-one: to see the contact's profile
* channels: to see the channels details

# Solution

## Channels chat banner

* make JID and name clickable (left)
  * when clicked, display the content of the current "i" dialog box
* remove the "i" button (right)

For the quit/leave/exit, config and bookmarks, we'll see in another issue.

## One-to-one chat banner

* make avatar, JID and name clickable (left)
  * when clicked, display the content of the current card dialog box
* remove card (right)

For the quit/exit/leave, we'll see in another issue.

## Both cases

Same height for the banner (pixel perfect)

jcbrand:
> remove "quit" (right)

How do you leave a groupchat then?

Nyco:
> How do you leave a groupchat then?
"Why quit?"
That's what most users said during our tests.

Leaving a channel belongs to the IRC and ICQ era.

In modern/contemporary team chats like Slack, Mattermost, RocketChat, Zulip (even in mobile interpersonal chats like WhatsApp, FB Messenger, Telegram), you just enter and stay. It is a very rare use case to leave a channel.

jcbrand:
[...]

PR #1612, opened 2019-06-19
issue 1611 fixed: no info icon in group/one-to-one chat panels

chaizeg:
no info, leave groupchat/one-to-one chat icons in chat panels, avatar and jid are clickable and showcase the information about the contact/groupchat

updates #1611

### DIFF
--- a/sass/_chatrooms.scss
+++ b/sass/_chatrooms.scss
@@ -92,8 +92,6 @@
             .chatroom-jid {
                 font-size: var(--font-size-small);
             }
-        }
-        .show-room-details-modal{
             cursor: pointer;
         }
     }
--- a/sass/_chatrooms.scss
+++ b/sass/_chatrooms.scss
@@ -92,6 +92,8 @@
             .chatroom-jid {
                 font-size: var(--font-size-small);
             }
+        }
+        .show-room-details-modal{
             cursor: pointer;
         }
     }
--- a/sass/_chatrooms.scss
+++ b/sass/_chatrooms.scss
@@ -93,6 +93,9 @@
                 font-size: var(--font-size-small);
             }
         }
+        .show-room-details-modal{
+            cursor: pointer;
+        }
     }
 
     .chatroom {
--- a/src/templates/chatbox_head.html
+++ b/src/templates/chatbox_head.html
@@ -3,7 +3,7 @@
     <div class="chatbox-title">
         <div class="row no-gutters">
             {[ if (o.type !== o._converse.HEADLINES_TYPE) { ]}
-               <div style="cursor:pointer;"></a> <canvas class="avatar show-user-details-modal" height="36" width="36"></canvas></div>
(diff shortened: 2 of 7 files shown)


src/templates/chatbox_head.html
```diff
@@ -3,11 +3,11 @@
     <div class="chatbox-title">
         <div class="row no-gutters">
             {[ if (o.type !== o._converse.HEADLINES_TYPE) { ]}
-                <canvas class="avatar" height="36" width="36"></canvas>
+               <div style="cursor:pointer;"></a> <canvas class="avatar show-user-details-modal" height="36" width="36"></canvas></div>
```
jcbrand: Please don't put `style` attributes in the markup. All styles should go in the Sass/CSS.

src/templates/chatbox_head.html
```diff
@@ -18,7 +18,6 @@
         </div>
     </div>
     <div class="chatbox-buttons row no-gutters">
-        <a class="chatbox-btn close-chatbox-button fa fa-times" title="{{{o.info_close}}}"></a>
-        <a class="chatbox-btn show-user-details-modal fa fa-id-card" title="{{{o.info_details}}}"></a>
+        <!----a class="chatbox-btn show-user-details-modal fa fa-id-card" title="{{{o.info_details}}}"--><!----/a-->
```
jcbrand: Please re-add the close button, see the discussion in #1611.

Also, don't comment out markup, delete it entirely if it's no longer used.

sass/_chatrooms.scss
```diff
@@ -93,6 +93,9 @@
                 font-size: var(--font-size-small);
             }
         }
+        .show-room-details-modal{
+            cursor: pointer;
+        }
```
jcbrand: I'd like us to avoid a proliferation of CSS classes. Can you not just add `cursor: pointer` to  `.show-user-details-modal` directly, instead of wrapping it in a div with a new class?

Also, don't we need to modify the `click` event handler in JavaScript to now listen for `click` on a different DOM element?

chaizeg: I deleted the div wrapping the avatar. As for the click event, there is no need to do that because I added .show-user-details-modal and .show-room-details-modal classes to the DOM elements that are meant to showcase groupchat/user info on click, so It's already handled.

jcbrand: Ok, looks much cleaner already. One more suggestion, how about putting `show-user-details-modal` on the `div.chatbox-title` element. You're already letting that element show a cursor, so it makes sense to also make it the same element that's clickable.

chaizeg: Done, although it actually makes the entire top area with the avatar and the nickname and the area next to it until right before the close button clickable..

jcbrand: Sorry but that doesn't sound right... then there shouldn't have been `cursor: pointer` on `.chat-title`.

jcbrand:
So much has changed in the meantime that I don't think this PR is relevant anymore.

PR #1614, opened 2019-06-19
Fixes #1613: X clear now of same size as text in presence dialog 

### LINKED ISSUE #1613: "x" clear cross now of same size as text in presence dialog box
In the status/presence dialog box, the "x" cross to clear the text field is presented like an exponent. It should be displayed the same way as text.

Almost the same issue as https://github.com/conversejs/converse.js/issues/1071

Capture d’écran 2019-06-19 à 13 40 30

chaizeg:


sass/_forms.scss
```diff
@@ -44,6 +44,11 @@
             font-size: var(--font-size);
         }
 
+        .clear-pres-dialog{
+            margin-top : 0.75em;
+            margin-right: 0.2em;
+        }
```
jcbrand: Why don't you reuse the `.clear-input` class?

jcbrand: I see it's being used in `src/templates/chat_status_modal` and in `src/templates/roster_filter.html`

chaizeg: I realized that clear-input also controls the same icon for filtering contacts. But as the placeholders for the presence dialog and the roster filter are of different sizes, using only clear-input will make the clear cross not well situated in the placeholder for the roster filter.

jcbrand: Can you not use CSS specificity rules to overcome this? Sass lets you nest rules within one another, you can probably use that to style `.clear-input` differently for each use-case.

PR #1615, opened 2019-06-19
Double entry

Nyco:
"- Properly handle message correction being received before the corrected message"
was there twice

Thanks for converse.js! :)


jcbrand:
Thanks :)

PR #1626, opened 2019-07-03
Support sending files from clipboard

Zash:
Allows fancy things like taking a screenshot and pasting into chat.

Tested in Firefox 60.7 ESR


jcbrand:
Thanks!

Issue #1476, opened 2019-03-04
Configuration settings to automatically activate OMEMO


knoy:
There should be configuration options to configure encryption policy, i.e: 

auto_omemo = true/false, automatically attempt to activate OMEMO whenever possible
force_omemo = true/false, do not allow OTR or plaintext communication

PR #1631, opened 2019-07-06
ADD Omemo default support

orbitz:


src/headless/converse-core.js
```diff
@@ -223,6 +223,7 @@ _converse.default_settings = {
     ],
     message_carbons: true,
     nickname: undefined,
+    omemo_default: false,
```
jcbrand: This needs to go into [src/converse-omemo.js](https://github.com/conversejs/converse.js/blob/master/src/converse-omemo.js) and you need to use [_converse.api.settings.update](https://conversejs.org/docs/html/api/-_converse.api.settings.html#.update) inside the plugin's [initialize](https://github.com/conversejs/converse.js/blob/master/src/converse-omemo.js#L240) method.

Other than that, looks good :+1:

orbitz: How's that?

jcbrand:
Thanks, one more thing please.

Mention the ticket number for https://github.com/conversejs/converse.js/issues/1476 in the changelog and make sure that your entry is in the right order.

Bonus points if you mention the ticket number in the commit message.

orbitz:
Done!

orbitz:
Looks like I broke something, I'll have to look tomorrow.

orbitz:
I'm also not sure this is working with the latest change, does the variable go to a different place than I'm checking with the method you said to use?

jcbrand:
Thanks!

PR #1634, opened 2019-07-08
improve self-contact features

worlword:
This commit adds some advanced functionality to the self-contact. When adding yourself as a contact the subscription will be completed automatically. Furthermore the xmpp-status will be tracked and correctly displayed for the self-contact roster-element.

### DIFF
--- a/src/converse-profile.js
+++ b/src/converse-profile.js
@@ -257,7 +257,21 @@ converse.plugins.add('converse-profile', {
             },
 
             afterRender () {
+                this.updateStatusDisplayForSelfContact();
                 this.renderAvatar();
+            },
+
+            updateStatusDisplayForSelfContact () {
+                const self = this;
+                // changing the status-display of the self-contact after the user changed the xmpp-status
+                const contacts = _converse.roster.models;
+                _.each(contacts, function(contact) {
+                    const contact_jid = contact.get('id');
+                    if (u.isSameBareJID(contact_jid, _converse.connection.jid)) {
+                        const chat_status = self.model.get('status') || 'offline';
+                        contact.presence.set('show', chat_status);
+                    }
+                });
             },
 
             showProfileModal (ev) {
--- a/src/converse-rosterview.js
+++ b/src/converse-rosterview.js
@@ -354,10 +354,20 @@ converse.plugins.add('converse-rosterview', {
                     u.hideElement(this.el);
                     return this;
                 }
-                const ask = this.model.get('ask'),
-                    show = this.model.presence.get('show'),
-                    requesting  = this.model.get('requesting'),
-                    subscription = this.model.get('subscription');
+
+                const contact_jid = this.model.get('jid'),
+                    requesting = this.model.get('requesting'),
+                    subscription = u.isSameBareJID(contact_jid, _converse.connection.jid) ? 'both' : this.model.get('subscription'),
+                    ask = u.isSameBareJID(contact_jid, _converse.connection.jid) ? 'none' : this.model.get('ask');
+                let show = this.model.presence.get('show');
+                
+                // set status of self-contact to current xmpp-status
+                // without this the self-contact will always be offline after login
+                if (u.isSameBareJID(contact_jid, _converse.connection.jid) && show === 'offline') {
+                    const status_self = _converse.xmppstatus.get('status');
+                    this.model.presence.set('show', status_self);
+                    show = status_self;
+                }
 
                 const classes_to_remove = [
                     'current-xmpp-contact',
--- a/src/headless/converse-roster.js
+++ b/src/headless/converse-roster.js
@@ -550,6 +550,12 @@ converse.plugins.add('converse-roster', {
              * @param { Object } attributes - Any additional attributes to be stored on the user's model.
              */
             async addContactToRoster (jid, name, groups, attributes) {
+
+                // set status-display of self-contact. Does not change xmpp-status
+                if (this.isSelf(jid)) {
+                    attributes = { chat_status: 'online' };
+                }
+
                 groups = groups || [];
                 try {
                     await this.sendContactAddIQ(jid, name, groups);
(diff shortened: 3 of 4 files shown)


src/converse-profile.js
```diff
@@ -257,9 +257,23 @@ converse.plugins.add('converse-profile', {
             },
 
             afterRender () {
+                this.updateStatusDisplayForSelfContact();
                 this.renderAvatar();
             },
 
+            updateStatusDisplayForSelfContact () {
+                const self = this;
```
jcbrand: Please don't do this. You can use an arrow function below to ensure that `this` is bound to the right context.

src/converse-profile.js
```diff
@@ -257,9 +257,23 @@ converse.plugins.add('converse-profile', {
...
+                // changing the status-display of the self-contact after the user changed the xmpp-status
+                const contacts = _converse.roster.models;
+                _.each(contacts, function(contact) {
```
jcbrand: You can use `contacts.forEach` instead of `_.each` (I'm trying to move away from lodash as much as possible).

Also, `function(contact)` can be an arrow function: `contact => {`, then you don't need to use `self`.

jcbrand: Looking at the code further, instead of looping through all contacts, you can instead do

 ```
const self_contact = contacts.findWhere({'jid': _converse.connection.jid});
if (self_contact) {
    self_contact.presence.save('show', this.model.get('status') || 'offline');
}
```

However, I don't think a View is the right place to put this kind of code.

Look here how I listen for the `add` event for a roster contact and then do stuff in the listener.
https://github.com/conversejs/converse.js/blob/master/src/headless/converse-roster.js#L935

You can do the same kind of thing and then use the code above. Alternatively you can add an `initialize` method to [_converse.RosterContacts](https://github.com/conversejs/converse.js/blob/master/src/headless/converse-roster.js#L391) and then create a listener there for the `add` event.

Either approach is probably fine, but because your listener depends on `converse-profile`, it's probably better to take the first approach and to put the code inside `converse-profile.js`.

Then, whenever the contact is added (after fetching it from cache or from the XMPP server), the right `status` value will be set.

Also, you should probably use `save` instead of `set` so that the value gets persisted in the cache.

Then, another thing to think about is to register an event handler for when `status` changes, so that you can update the contact's presence.

File: src/converse-rosterview.js [hunk 3]
```diff
@@ -354,10 +354,20 @@ converse.plugins.add('converse-rosterview', {
...
                     return this;
                 }
-                const ask = this.model.get('ask'),
-                    show = this.model.presence.get('show'),
-                    requesting  = this.model.get('requesting'),
-                    subscription = this.model.get('subscription');
+
+                const contact_jid = this.model.get('jid'),
+                    requesting = this.model.get('requesting'),
+                    subscription = u.isSameBareJID(contact_jid, _converse.connection.jid) ? 'both' : this.model.get('subscription'),
```
jcbrand: You shouldn't do this.

Instead, when the contact gets created, check whether it's yourself and then if so, set `subscription` as `both`, so that the value is persisted.

However, I'm not sure this is right in general, the right value must be stored on the XMPP server, then we don't need any checks like this in Converse and it'll also work properly on other clients.

So I think it's still better to let yourself show up as a requesting contact and then to accept your own contact request, then doing it like this here.

src/converse-rosterview.js
```diff
@@ -354,10 +354,20 @@ converse.plugins.add('converse-rosterview', {
...
+                    ask = u.isSameBareJID(contact_jid, _converse.connection.jid) ? 'none' : this.model.get('ask');
+                let show = this.model.presence.get('show');
+                
+                // set status of self-contact to current xmpp-status
+                // without this the self-contact will always be offline after login
+                if (u.isSameBareJID(contact_jid, _converse.connection.jid) && show === 'offline') {
+                    const status_self = _converse.xmppstatus.get('status');
+                    this.model.presence.set('show', status_self);
+                    show = status_self;
```
jcbrand: Why is it necessary to do this here *and* in `converse-profile`? Seems like duplication to me.

src/headless/converse-roster.js
```diff
@@ -559,12 +565,12 @@ converse.plugins.add('converse-roster', {
...
                 }
                 return this.create(_.assignIn({
-                    'ask': undefined,
+                    'ask': (this.isSelf(jid) ? 'none' : undefined),
                     'nickname': name,
                     groups,
                     jid,
                     'requesting': false,
-                    'subscription': 'none'
+                    'subscription': (this.isSelf(jid) ? 'both' : 'none')
```
jcbrand: I don't think this is right, due to what I've written above.

src/headless/converse-roster.js
```diff
@@ -691,10 +697,12 @@ converse.plugins.add('converse-roster', {
                  * received in the IQ from the server.
                  */
                 const jid = item.getAttribute('jid');
-                if (this.isSelf(jid)) { return; }
+                let subscription = item.getAttribute("subscription");
+                if (this.isSelf(jid) && subscription !== 'remove') { 
+                    subscription = 'both'; 
+                }
```
jcbrand: Also here...

worlword:
Hey, thank you for your feedback. I will look into it. 

As for the subscription --> both issue. I will look into it, but it was implemented because converse.js (+ eJabberd) does not dispaly the option to accept yourself as a contact. So the contact would always be pending with no chance to accept it. That's why we wanted to send the subscription as "both" so that the server already knows we accepted ourselfs. But I will look into it.
Thanks again for the tips.

jcbrand:
OK, I think in that case you need to add code that detects that the pending contact is yourself and then automatically accepts the request.

I'm not at my laptop now, so can't look up the details.

ChaosKid42:
I believe this can be closed because of https://github.com/conversejs/converse.js/commit/40726e06e9a17e41dc0869bd4832f03eb4b19ef2

Issue #1635, opened 2019-07-09
credentials_url not fetching JID


s-sokolko:
**Describe the bug**
When initializing converse using credentials_url config parameter conversejs fetches and uses the password provided, but it keeps looking for JID in config params rather than use the JID provided by credentials_url response

**To Reproduce**
Init converse like that:
```javascript 
converse.initialize({
	    authentication: 'login',
	    auto_away: 300,
	    auto_reconnect: true,
	    keepalive: true,
	    auto_list_rooms: true,
	    websocket_url: 'wss://xmpp.o-it-s.ru:5280/websocket/',
	    message_archiving: 'always',
	    view_mode: 'overlayed',
	    allow_logout: false,
	    allow_registration: false,
	    play_sounds: true,
	    locales_url: '/locale/' + language + '/LC_MESSAGES/converse.json',
	    auto_login: true,
	    credentials_url: 'http://localhost/creds.php',
	    debug: true,
	});
```

Here https://github.com/conversejs/converse.js/blob/master/src/headless/converse-core.js#L1272 
and here https://github.com/conversejs/converse.js/blob/master/src/headless/converse-core.js#L1287
one can see that credentials.jid is not used

**Expected behavior**
Converse is expected to pull both jid and password according to the docs https://conversejs.org/docs/html/configuration.html#credentials-url

**Screenshots**

**Environment (please complete the following information):**
I use desktop PC with Google Chrome web browser. ConverseJS is latest master (sha 5c88d3e836d347dbbd5afc0bbd1cb5fb3348ddc6).

**Additional context**

jcbrand:
@s-sokolko: I use `credentials_url` a lot in various projects, and I just checked why I don't seem to have the problem you reported.

In `fetchLoginCredentials` we call `setUserJID` which sets `this.jid` and also creates a JID resource if one wasn't provided by the `credentials_url` endpoint.
See here: https://github.com/conversejs/converse.js/blob/8a405ee88cf8ba277a3a04bf04b7185f2f1562db/src/headless/converse-core.js#L618

That last point is important, thats why we use `this.jid` here:
https://github.com/conversejs/converse.js/blob/8a405ee88cf8ba277a3a04bf04b7185f2f1562db/src/headless/converse-core.js#L1288

What is actually the problem that you're having? Is `this.jid` undefined when calling `this.connection.connect`?

Based on line 618, which I link to above, I don't see any problem and it is correct that `this.jid` is used later on.

[...]

PR #1638, opened 2019-07-09
Fix: use JID from credentials when credentials_url is set

s-sokolko:
I am not sure if I can make tests for this, but this PR is an attempt to fix https://github.com/conversejs/converse.js/issues/1635

Works for me

### DIFF
--- a/src/headless/converse-core.js
+++ b/src/headless/converse-core.js
@@ -1272,9 +1272,6 @@ _converse.initialize = async function (settings, callback) {
             const password = _.isNil(credentials)
                 ? _converse.connection.pass || this.password
                 : credentials.password;
-            const jid = _.isNil(credentials)
-                ? this.jid
-                : credentials.jid;
             if (!password) {
                 if (this.auto_login) {
                     throw new Error("autoLogin: If you use auto_login and "+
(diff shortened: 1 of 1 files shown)


src/headless/converse-core.js
```diff
@@ -1272,6 +1272,9 @@ _converse.initialize = async function (settings, callback) {
             const password = _.isNil(credentials)
                 ? _converse.connection.pass || this.password
                 : credentials.password;
+            const jid = _.isNil(credentials)
+                ? _converse.connection.authcid || this.jid
+                : credentials.jid;
```
jcbrand: Why do you add `_converse.connection.authcid` there as well?

s-sokolko: @jcbrand I am quite new to Strophe and I tried to make my code similar to password obtaining process (from Strophe connection). I can just use 
```javascript
const jid = _.isNil(credentials)
                ? this.jid
                : credentials.jid;
```
Is it a better option?

jcbrand: Yes, I don't think authcid is necessary

s-sokolko: @jcbrand done. But tests are failing for some strange reason

jcbrand:
Well done, thanks!

PR #1639, opened 2019-07-10
Add Broadcast-/MOTD-Message functionality (XEP-0133)

worlword:
Adds the ability to send Announcements and to set the Message of the
Day. After Login a stanza is send to the xmpp-server to check if the
user is allowed to send these messages. If the user has permission a
submenu will be added to the controlbox.

This implementation does not fully implement the service-administration
XEP-0133. This is because eJabberd handles the standard differently. So
it is not certain it will work with other xmpp-servers. Mainly eJabberd
expects a message-stanza instead of an iq-Stanza.

Furthermore adds a "-- send from <jid>" at the end of the announcement.
This helps users to contact the sender.

Even if this does send a different stanza to the server, I believe the code is still valuable. For other servers the msg-stanza need to be changed to a iq-stanza (I cannot do this because I'm not able to test it with another server).

I also tried to add the views as a "chat" so that converse can handle the visibility automatically. But this attempt failed :( So now I added code to clean the view manually.

An example-Screenshot:
01_examplePic

### DIFF
--- a/sass/_chatbox.scss
+++ b/sass/_chatbox.scss
@@ -548,6 +548,27 @@
 
 }
 
+#conversejs {
+    #service-administration {
+        .flyout {
+            border-color: var(--service-administration-color);
+        }
+        .chat-head-chatbox {
+            background-color: var(--service-administration-color);
+        }
+        .chat-body {
+            background-color: var(--chat-textarea-background-color);
+            display: block;
+            padding: 1em;
+            overflow: auto;
+        }
+        .send-button {
+            margin-top: 1em;
+            width: 100%;
+        }
+    }
+}
+
 #conversejs.converse-embedded {
     .converse-chatboxes {
         z-index: 1031; // One more than bootstrap navbar
--- a/src/converse-service-administration.js
+++ b/src/converse-service-administration.js
@@ -1,0 +1,433 @@
+// Converse.js (A browser based XMPP chat client)
+// https://conversejs.org
+//
+// Copyright (c) 2012-2017, Jan-Carel Brand <jc@opkode.com>
+// Licensed under the Mozilla Public License (MPLv2)
+//
+
+import "converse-chatview";
+import "converse-profile";
+import "converse-rosterview";
+import converse from "@converse/headless/converse-core";
+import sizzle from "sizzle";
+import tpl_form_checkbox from "templates/form_checkbox.html";
+import tpl_form_input from "templates/form_input.html";
+import tpl_form_textarea from "templates/form_textarea.html";
+import tpl_service_admin_chat from "templates/service_administration_chat.html";
+import tpl_service_admin_option from "templates/service_administration_option.html";
+import tpl_service_admin_selector from "templates/service_administration_selector.html";
+
+const { $iq, Backbone, Strophe } = converse.env;
+const u = converse.env.utils;
+
+converse.plugins.add('converse-service-administration', {
+
+    /* Plugin dependencies are other plugins which might be
+     * overridden or relied upon, and therefore need to be loaded before
+     * this plugin.
+     *
+     * If the setting "strict_plugin_dependencies" is set to true,
+     * an error will be raised if the plugin is not found. By default it's
+     * false, which means these plugins are only loaded opportunistically.
+     *
+     * NB: These plugins need to have already been loaded via require.js.
+     */
+    dependencies: ["converse-chatboxes", "converse-rosterview", "converse-chatview"],
+
+    initialize () {
+        const { _converse } = this,
+            { __ } = _converse;
+
+        _converse.SERVICE_ADMIN_TYPE = 'service-administration';
+
+        _converse.ServiceAdministration = Backbone.Model.extend({
+            async discoverSupport () {
+                const supported = await _converse.api.disco.supports(Strophe.NS.COMMANDS, _converse.bare_jid);
+                if (supported) {
+                    const stanza = $iq({
+                        'from': _converse.connection.jid,
+                        'id': 'test-supported-commands',
+                        'to': _converse.domain,
+                        'type': 'get'
+                    }).c('query', {
+                        xmlns: Strophe.NS.DISCO_ITEMS,
+                        node: 'announce'
+                    }).up();
+
+                    _converse.connection.sendIQ(stanza, result => {
+                        const commands = sizzle('item[node^="http://jabber.org/protocol/admin"]', result);
+                        _converse.serviceAdminCommands = new _converse.ServiceAdminCommandCollection();
+                        commands.forEach(command => {
+                            const command_description = command.getAttribute('name');
+                            const command_node = command.getAttribute('node');
+                            const command_name = command_node.split('#').pop();
+                            _converse.serviceAdminCommands.add(new _converse.ServiceAdminCommandItem({
+                                command_node: command_node,
+                                description: command_description,
+                                command_name: command_name
+                            }));
+                        });
+                        this.renderControlboxElement();
+                    });
+                }
+            },
+
+            renderControlboxElement () {
+                _converse.serviceAdministrationCommandView = new _converse.ServiceAdminCommandView({collection: _converse.serviceAdminCommands});
+                const groupchat_element = _converse.chatboxviews.get('controlbox').el.querySelector('#chatrooms');
+                if (groupchat_element !== null) {
+                    groupchat_element.insertAdjacentElement('beforebegin', _converse.serviceAdministrationCommandView.render());
+                }
+            }
+        });
+
+        _converse.ServiceAdminCommandItem = Backbone.Model.extend({});
+
+        _converse.ServiceAdminCommandItemView = Backbone.NativeView.extend({
+            tagName: 'div',
+            className: 'list-item controlbox-padded d-flex flex-row',
+            events: {
+                'click': 'openServiceAdminChat'
+            },
+            stanza: 'server_response',
+
+            render () {
+                this.el.innerHTML = tpl_service_admin_option({
+                    'name': this.model.get('command_name'),
+                    'service_admin_name': this.model.get('command_name')
+                });
+                return this.el;
+            },
+
+            openServiceAdminChat () {
+                const stanza = $iq({
+                    'from': _converse.connection.jid,
+                    'id': this.model.attributes.command_name + '-1',
+                    'to': _converse.domain,
+                    'type': 'set',
+                    'xml:lang': _converse.locale
+                }).c('command', {
+                    xmlns: Strophe.NS.COMMANDS,
+                    action: 'execute',
+                    node: this.model.attributes.command_node
+                }).up();
+
+                _converse.connection.sendIQ(stanza, result => {
+                    this.server_response = result;
+                    const session_id = sizzle('command', this.server_response).pop().getAttribute('sessionid');
+                    const title = sizzle('title', this.server_response).pop();
+                    const fields = sizzle('field', this.server_response);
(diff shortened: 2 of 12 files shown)


jcbrand:
Hi @worlword 

Thanks a lot for this, it looks like a great feature to have.

I recently broke the build, which is why the tests are failing currently for this PR. Can you please rebase your commits on the latest HEAD master branch?

worlword:
Hey,

I updated the pull-Request to the latest commit.

sass/_chatbox.scss
```diff
@@ -121,6 +121,20 @@
             width: var(--mobile-chat-width);
         }
 
+        #service_administration_description {
+            color: #777;
```
jcbrand: To make theming easier we should try to avoid hardcoding colors and instead use CSS properties.

Looks to me like here you can use `color: var(--secondary-color);`

sass/_chatbox.scss
```diff
@@ -121,6 +121,20 @@
...
+            background-color:white;
+            display:block;
+            padding:5px;
```
jcbrand: Use `em` for padding. That way padding is relative to the font-size and scales when people make their browser text larger or smaller (e.g with <ctrl> - and <ctrl> =).

sass/_chatbox.scss
```diff
@@ -121,6 +121,20 @@
             width: var(--mobile-chat-width);
         }
 
+        #service_administration_description {
+            color: #777;
+            background-color:white;
```
jcbrand: Looks like you should use `var(--chat-textarea-background-color);` instead of `white;`.

sass/_chatbox.scss
```diff
@@ -121,6 +121,20 @@
...
+            display:block;
+            padding:5px;
+        }
+        #mod_announce_text_message {
+            width: 100%;
+            height: calc(100% - var(--fullpage-chat-head-height));
+            margin-top: 10px;
+            margin-bottom: 10px;
```
jcbrand: Use `em` units instead of `px`.

src/converse-controlbox.js
```diff
@@ -355,7 +362,7 @@ converse.plugins.add('converse-controlbox', {
                 'errors': [],
             }
         });
-
+
```
jcbrand: This whitespace needs to be removed.

src/converse-service-administration.js
```diff
@@ -0,0 +1,452 @@
...
+     *
+     * If the setting "strict_plugin_dependencies" is set to true,
+     * an error will be raised if the plugin is not found. By default it's
+     * false, which means these plugins are only loaded opportunistically.
+     *
+     * NB: These plugins need to have already been loaded via require.js.
+     */
+    dependencies: ["converse-chatboxes", "converse-rosterview", "converse-chatview"],
+
+    overrides: {},
```
jcbrand: You don't need to specify `overrides` if it's empty.

src/converse-service-administration.js
```diff
@@ -0,0 +1,452 @@
...
+                    'controlbox_el': _converse.chatboxviews.get('controlbox').el.querySelector('.controlbox-panes') 
+                });
+                _converse.serviceAdministrationDiscovery.canUserSendBroadcastMessage();
+            }    
+        });
+
+        _converse.ServiceAdministrationDiscovery = Backbone.Model.extend({
+
+            // check if the user is allowed to send Broadcast-/MOTD-Messages
+            canUserSendBroadcastMessage () {
```
jcbrand: This is not a good way to check for support.

From the XEP:
_In order to interact with a particular component attached to a server, an administrator needs to first discover that component and the commands it support, then send the appropriate command to the component itself._


Converse has general support for service discovery, so you can use the [_converse.api.disco](https://conversejs.org/docs/html/api/-_converse.api.disco.html) API.

For example:

`const supported = await _converse.api.disco.supports(Strophe.NS.COMMANDS, _converse.bare_jid);`.

You'll need to define `Strophe.NS.COMMANDS` in converse-core.js via `Strophe.addNamespace` (there are lots of examples). The value should be `http://jabber.org/protocol/commands`.

Once you know whether ad-hoc commands are supported, you can get a list of supported commands via:

`const commands = await _converse.api.disco.items(_converse.bare_jid, Strophe.NS.COMMANDS);`

You'll then know exactly which commands are supported by the server.

src/converse-service-administration.js
```diff
@@ -0,0 +1,452 @@
...
+                    const converse_html = _converse.chatboxviews.el.querySelector('#controlbox');
+                    converse_html.insertAdjacentElement('afterend', element);
+                }
+
+                u.slideIn(this.el.parentElement);
+            }
+        });
+
+        // Model for all the service-administration-options provided (XEP-0133)
+        _converse.servive_administration_model = Backbone.Model.extend({ });
```
jcbrand: Object names which can be instantiated with `new` should be in BumpyCase (with a starting capital letter), not with underscores.

src/converse-service-administration.js
```diff
@@ -0,0 +1,452 @@
...
+
+        _converse.AdminAnnounceAll = new _converse.servive_administration_model({
```
jcbrand: Once you've received a list of supported commands via disco (as explained above), you can create these models only for the actual commands supported.

src/converse-controlbox.js
```diff
@@ -103,6 +104,12 @@ converse.plugins.add('converse-controlbox', {
                     }
                     view.close();
                 });
+
+                const servive_administration_chat = _converse.chatboxviews.el.querySelector('#sendAdminCommandId');
```
jcbrand: There's a typo here, `servive` needs to be `service`. Also, this variable name is very long, everywhere where you use `administration` you can shorten it to `admin`.

src/converse-chatview.js
```diff
@@ -1312,6 +1312,10 @@ converse.plugins.add('converse-chatview', {
                 if (!views.get(item.get('id')) && item.get('type') === _converse.PRIVATE_CHAT_TYPE) {
                     views.add(item.get('id'), new _converse.ChatBoxView({model: item}));
                 }
+                const servive_administration_chat = _converse.chatboxviews.el.querySelector('#sendAdminCommandId');
+                if(!_.isNil(servive_administration_chat)) {
+                    servive_administration_chat.remove();
+                }
```
jcbrand: This looks wrong. The chatview plugin does not have a dependency on the service-admin plugin, so it should not have to be aware of anything related to that plugin and therefore this code should not be here.

Looks like you're adding the `#sendAdminCommandId` element to *all* chat boxes and then removing it again here, right? You shouldn't be adding it to all of them in the first place, only to the relevant one.

worlword: Thanks for the feedback. This one is a litte more complicated. That's the problem i refered to when I wrote 

> I also tried to add the views as a "chat" so that converse can handle the visibility automatically. But this attempt failed :( So now I added code to clean the view manually.

The problem here (mainly in fullscreen-mode) is that the service-administration-window (left on the screenshot) is not a real chat. So when the user opens a normal chat/muc I have to manually close the service-administration-windows so that the chat will be displayed. Normally converse.js does this by hiding the current chat and then opening the new chat (in fullscreen). 

The best way would be if the service-administration-window would be handled like the other chats. That way no manual cleanup would be required. That way the code inside `converse-controlbox.js` could also be removed.

I assume that the new view has to be added to `_converse.chatboxviews` and `_converse.chatboxes`. This way converse can handle displaying/removing the view. But I couldn't get it to work. That's why I clean it manually here, when a new chat is opened. I could try to create my own event-listeners inside `converse-service-administration.js` to clean the view. That way the code wouldn't pollute the other plugins.

jcbrand: > The best way would be if the service-administration-window would be handled like the other chats.

Take a look at the `HeadlinesBox` and `HeadlinesBoxView` in [converse-headline.js](https://github.com/conversejs/converse.js/blob/master/src/converse-headline.js).

That's also not a "real chat" but they both extend the chatbox base classes and therefore don't have this problem. You should be able to do the same with the service administration box. I think it's conceptually very similar. Same with the `ControlBoxView`, it also extends the `ChatBoxView` baseclass.

Currently the headlines box doesn't appear in fullscreen but that's more a UX issue than anything else. There's nowhere to click to bring it to the foreground :D
You don't have this problem because you added a section in the controlbox for opening the service admin box.

I'd like to someday refactor the base classes so that the one-on-one and MUC boxes both extend a common base class instead of everything extending the current one-on-one class, but I don't have time for it currently and it isn't (yet) really an issue.

worlword:
Hello JCBrand,

I updated the pull-request. But there are still some problems. But first the stuff that works...

The whole feature now uses iq-stanzas (like the xep-0133) suggests. There are iq-stanzas to discover support, to fetch the commands, to fetch the form and to send the filled-out form back to the server. The form is rendered dynamically.
I made the changes to the css you suggested. All the other stuff should be done as well. Well, it was basically a re-write. 

Issues I used a workaround for: 
* I cannot import `sizzle`. I also can't use `const { sizzle } from converse.env;`. I don't know why. That's why I use `converse.env.sizzle` everywhere. 
* `_converse.api.disco.items` did not work. I always got an empty list of commands back. So I wrote my own stanza.

Now there are some problems I would like your help with. You know converse.js better than me and you have more knowledge about backbone.js than me :) So I hope you can take a look. I marked both both places with an 'ERROR' comment. 
* At the moment this feature only works in fullscreen. In overlay the dragresize-module tries to call save on the model. That returns an error that `no "url" property was specified`. I tried many things but this error still persists. When I override the `url()`-Funktion I get the error that the `apply` property does not exist. I really don't know what's wrong here so I would be greatful for any hints/help. 
* `close() `get's into an inifinite loop. `this.model.close()` tirggers an event that calls `close()` again. So there is a loop created. I tried to copy your code there, but I couldn't get it to work. So again, I would be greatful for any advice. 

Lastly a bit of a letdown. Currently there are only checkboxes, input-fields and multi-text-fields supported in the forms. That's because those are the only ones used by ejabberd. I also don't read any `<instruction>`-tag because ejabberd does not send those as well. I simply don't want to commit code that I cannot test. But it should be relatively easy to add new form-elements and to read the `instruction`-tag.

worlword:
I updated the pull-Request. 
Sizzle is now imported properly and used in the code. The other problems still persist...
I read about Backbone.js again. As far as I understand now, Backbone.js should manage the uri itself. So I don't know why mine is missing the uri. I also found out that normal chat-windows also don't have an `uri`-attribute set. It still passes through the breakpoint and `save` is called. But there is no error for normal chats. I'm not sure if I need to use `_converse.router` for this to work. But I think that is used to "redirect" the user to another "Page" and has nothing to do with the storage of the models.

jcbrand:
The URI error is very low-level and is related to the `browserStorage` attribute set on models or whether the model has a `collection` attribute linking to the collection that contains it.

Debugging these kinds of errors can be tricky.

ChaosKid42:
@worlword This seems to resolve the issue regarding backbone.js:

```js
        _converse.ServiceAdminBox = _converse.ChatBox.extend({
            defaults () {
                return {
                    'bookmarked': false,
                    'hidden': ['mobile', 'fullscreen'].includes(_converse.view_mode),
                    'message_type': 'service-administration',
                    'num_unread': 0,
                    'time_opened': this.get('time_opened') || (new Date()).getTime(),
                    'type': _converse.SERVICE_ADMIN_TYPE
                };
            },
            initialize () {
                this.set({'box_id': `box-${btoa(this.get('jid'))}`});
            }
        });
```

jcbrand:
Closing this in favour of #1834

Issue #1416, opened 2019-01-16
Hide participant list in "embedded" mode on mobile devices


ge0rg:
It would be great to have "embedded" mode work both on mobile and on desktop devices. One of the last missing pieces is that the participant list of a MUC is always shown by default. It would be awesome to have it hidden on mobile devices and only show on the desktop / in landscape mode by default.

Ppjet6:
I was going to report a similar issue. I would argue the participant list should be hidden by default on mobile at all, (not just in embedded mode).

deleolajide:
I think the participant list should always be hidden by default. I use a plugin to ensure that and also provide a command to open and peek at **who** is in the groupchat  without leaving the input text-area.
```
        if (command == "who")
        {
            view.toggleOccupants(null, false);
            view.scrollDown();
            return true;
        }
```

licaon-kter:
Not sure if the "need" to currently open the list of users on join is the issue, but large rooms will just balloon memory trying to load it until killed by the OS.

Eg.
![bigroomfail](https://user-images.githubusercontent.com/2282799/53331965-da7b3d80-38fa-11e9-8531-a2e0e1709dde.png)

Steps? Join `xmpp:#matrix#matrix.org@bridge.xmpp.matrix.org?join`

FYI, Conversations (at least 2.4.0 or later) and apparently Yaxim handle these ok.

Ppjet6:
@licaon-kter, I don't think your issue here is related to the display of the participant list itself, it's probably that converse gets swamped processing presences for 5k participants. That should probably be a separate issue.

tasn:
If it helps anyone, I worked around it locally by emulating a click on the "hide participant list plugin":
```
  converse.initialize({
    ...
  }).then(() => {
    var toggleButton = document.getElementsByClassName("toggle-occupants fa-angle-double-right")[0];
    if (toggleButton) {
      toggleButton.click();
    }
  });
```

ge0rg:
In Converse 5.x, the class name is `toggle-occupants float-right fa fa-angle-double-right` but it seems not to work for me. Maybe this code needs to be called after the nickname is entered and the join completes?

tasn:
@ge0rg, it also works intermittently for me. It used to work reliably, not sure what's going on.

PR #1640, opened 2019-07-10
Add resize-functionality to MUC-Member-List

worlword:
This adds the ability to resize the MUC-Member-List in all MUCs. The
MUC-Member-List can be scaled between 20% width of the MUC and 75% of
the MUC.

Everything was implemented as a plugin and can be disabled in the
converse.js-File. Furthermore this functionality can be enabled or
disabled through the allow_muc_roster_resizing parameter. Setting the
allow_muc_roster_resizing true will enable the feature. Setting it to
false will disable the feature.

### DIFF
--- a/src/converse-dragresize.js
+++ b/src/converse-dragresize.js
@@ -36,7 +36,7 @@ converse.plugins.add('converse-dragresize', {
      *
      * NB: These plugins need to have already been loaded via require.js.
      */
-    dependencies: ["converse-chatview", "converse-headline", "converse-muc-views"],
+    dependencies: ["converse-chatview", "converse-headline", "converse-muc-views", "converse-mouse-events"],
 
     enabled (_converse) {
         return _converse.view_mode == 'overlayed';
@@ -146,7 +146,6 @@ converse.plugins.add('converse-dragresize', {
             'allow_dragresize': true,
         });
 
-
         const dragResizable = {
 
             initDragResize () {
@@ -182,7 +181,7 @@ converse.plugins.add('converse-dragresize', {
                 return this;
             },
 
-            resizeChatBox (ev) {
+            resize (ev) {
                 let diff;
                 if (_converse.resizing.direction.indexOf('top') === 0) {
                     diff = ev.pageY - this.prev_pageY;
@@ -314,59 +313,7 @@ converse.plugins.add('converse-dragresize', {
         };
         Object.assign(_converse.ChatBoxView.prototype, dragResizable);
 
-
-        _converse.applyDragResistance = function (value, default_value) {
-            /* This method applies some resistance around the
-            * default_value. If value is close enough to
-            * default_value, then default_value is returned instead.
-            */
-            if (value === undefined) {
-                return undefined;
-            } else if (default_value === undefined) {
-                return value;
-            }
-            const resistance = 10;
-            if ((value !== default_value) &&
-                (Math.abs(value- default_value) < resistance)) {
-                return default_value;
-            }
-            return value;
-        };
-
-
-        /************************ BEGIN Event Handlers ************************/
-        function registerGlobalEventHandlers () {
-
-            document.addEventListener('mousemove', function (ev) {
-                if (!_converse.resizing || !_converse.allow_dragresize) { return true; }
-                ev.preventDefault();
-                _converse.resizing.chatbox.resizeChatBox(ev);
-            });
-
-            document.addEventListener('mouseup', function (ev) {
-                if (!_converse.resizing || !_converse.allow_dragresize) { return true; }
-                ev.preventDefault();
-                const height = _converse.applyDragResistance(
-                        _converse.resizing.chatbox.height,
-                        _converse.resizing.chatbox.model.get('default_height')
-                );
-                const width = _converse.applyDragResistance(
-                        _converse.resizing.chatbox.width,
-                        _converse.resizing.chatbox.model.get('default_width')
-                );
-                if (_converse.api.connection.connected()) {
-                    _converse.resizing.chatbox.model.save({'height': height});
-                    _converse.resizing.chatbox.model.save({'width': width});
-                } else {
-                    _converse.resizing.chatbox.model.set({'height': height});
-                    _converse.resizing.chatbox.model.set({'width': width});
-                }
-                _converse.resizing = null;
-            });
-        }
-        _converse.api.listen.on('registeredGlobalEventHandlers', registerGlobalEventHandlers);
         _converse.api.listen.on('beforeShowingChatView', view => view.initDragResize().setDimensions());
         /************************ END Event Handlers ************************/
     }
 });
-
--- a/src/headless/converse-core.js
+++ b/src/headless/converse-core.js
@@ -234,6 +234,7 @@ _converse.default_connection_options = {'explicitResourceBinding': true};
 // ----------------------------
 _converse.default_settings = {
     allow_non_roster_messaging: false,
+    allow_occupants_view_resizing: false,
     authentication: 'login', // Available values are "login", "prebind", "anonymous" and "external".
     auto_away: 0, // Seconds after which user status is set to 'away'
     auto_login: false, // Currently only used in connection with anonymous login
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## 6.0.0 (Unreleased)
 
+- Add: Resizing of the occupants-list in Multi-User-Chats
 - #129: Add support for XEP-0156: Disovering Alternative XMPP Connection Methods. Only XML is supported for now.
 - #1105: Preliminary support for storing persistent data in IndexedDB instead of localStorage
 - #1089: When filtering the roster for `online` users, show all non-offline users.
--- a/sass/_chatbox.scss
+++ b/sass/_chatbox.scss
@@ -386,7 +386,8 @@
                 height: 5px;
                 width: 100%;
             }
-            &-left {
+            &-left,
+            &-occupants-left {
                 cursor: w-resize;
                 width: 5px;
                 height: 100%;
--- a/sass/_chatrooms.scss
+++ b/sass/_chatrooms.scss
@@ -155,6 +155,8 @@
                     display: flex;
                     flex-direction: column;
                     word-wrap: break-word;
+                    flex: 0 1 100%; 
+                    min-width: 25%;
                     .new-msgs-indicator {
                         background-color: var(--chatroom-head-color);
                     }
@@ -173,7 +175,9 @@
                     border-left: var(--occupants-border-left);
                     border-bottom-right-radius: var(--chatbox-border-radius);
                     padding: 0.5em;
-                    max-width: var(--occupants-max-width);
+                    max-width: 75%;
+                    min-width: 20%;
+                    flex: 0 0 25%;
 
                     .occupants-header {
                         display: flex;
--- a/src/converse-mouse-events.js
+++ b/src/converse-mouse-events.js
@@ -1,0 +1,91 @@
+// Converse.js (A browser based XMPP chat client)
+// https://conversejs.org
+//
+// Copyright (c) 2012-2017, Jan-Carel Brand <jc@opkode.com>
+// Licensed under the Mozilla Public License (MPLv2)
+//
+
+import "converse-chatview";
+import "converse-controlbox";
+import converse from "@converse/headless/converse-core";
+
+const { _ } = converse.env;
+
+
+converse.plugins.add('converse-mouse-events', {
+
+    dependencies: ["converse-chatview", "converse-headline", "converse-muc-views"],
+
+    enabled (_converse) {
+        return _converse.view_mode == 'overlayed' ||
+               _converse.allow_occupants_view_resizing;
+    },
+
+    initialize () {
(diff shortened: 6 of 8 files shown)


jcbrand:
Hi @worlword, can you also rebase this PR onto the latest master please?

worlword:
Hey, I don't have time right now. I will do it tomorrow.

worlword:
Ok. I rebased and created another pull-request. But I don't know, why the error is thrown. When I run "make serve" and then run the tests in the browser, all tests complete fine. Is the build still broken?

jcbrand:
Unfortunately on Travis the tests are much more flaky and prone to fail due to timing issues, even if they're 100% green locally.

I've restarted the tests on Travis, now they're green.

docs/source/configuration.rst
```diff
@@ -198,6 +198,14 @@ feature and you don't mind that your room bookmarks are visible to all
...
 able to have any room bookmarks at all for an account on that XMPP server.
 
+allow_occupants_view_resizing
+----------------------
+
+* Default: ``true``
+
+Allows users to resize the occupants-list inside multi-user-chats. Setting this to ``false`` 
+deactivates the resizing-capability.
+
```
jcbrand: I'm not convinced that this config setting is necessary. We already have the `dragresize` config seting (which I also wonder whether it's really necessary), so perhaps we can just let that setting determine drag-resizing for the boxes and the sidebar.

Continuously adding new config options makes the code more complex and creates more places for bugs to hide, so I actually want to try and reduce the amount of config options, or at least keep them to a minimum.

src/converse-resize-occupants-view.js
```diff
@@ -0,0 +1,161 @@
...
+/*global define, window, document */
+
+import "converse-chatview";
+import "converse-controlbox";
+import converse from "@converse/headless/converse-core";
+
+const { _ } = converse.env;
+
+converse.plugins.add('converse-resize-occupants-view', {
+    /* Plugin dependencies are other plugins which might be
```
jcbrand: Can all this code rather go into `converse-dragresize.js`? They are very similar and both have to do with dragging and resizing elements, and by combining them we don't have to keep track of an extra plugin.

If you do that, then you also don't need to create the extra `converse-mouse-events` plugin either, since all the code can be in one plugin.

src/converse-resize-occupants-view.js
```diff
@@ -0,0 +1,161 @@
...
+                const flyout = this.el,
+                      style = window.getComputedStyle(flyout);
+                this.width = parseInt(style.width.replace(/px$/, ''), 10);
+                const { _converse } = this.__super__;
+                _converse.resizing = {
+                    'chatbox': this,
+                    'direction': 'left'
+                };
+                this.prev_pageX = ev.pageX;
+            },
```
jcbrand: Methods such as this one, which don't exist on the original model being overridden, should rather be added via composition with `Object.assign` (you'll see lots of examples of this in other plugins).

Using `Object.assign` is a bit cleaner and explicit way of doing it, relying less on `pluggable.js` magic.

src/headless/converse-core.js
```diff
@@ -200,11 +200,13 @@ _converse.default_connection_options = {'explicitResourceBinding': true};
 // ----------------------------
 _converse.default_settings = {
     allow_non_roster_messaging: false,
+    allow_muc_roster_resizing: true,
```
jcbrand: Looks like this is unused.

worlword:
Hallo JCBrand, thanks for the feedback. 

As for the problem with remembering the size on reload. Normally you save these values in the event-listener for the "mouseup" event. The problem here is, that ChatRoomOccupantsView is a Backbone.OrderedListView. This prevents me from using the "save"-method like you did for the chatboxes when you resize. So I'm stuck on a solution for this currently. So if you have any advice i will take it :D

worlword:
Ok, I solved the problem. The size will now be saved during reload. I also added added your other suggestions. Sadly I'm out of time right now. I will review the changes again on the weekend and then update the pull-request.

jcbrand:
Thanks @worlword!

I don't think I'll be able to review this again soon, I'm under some time pressure myself, but I'll get back to it ASAP.

worlword:
Hey, I just found a problem with moving all code back to converse-dragresize.js. It only works inside overlay mode. If you move to fullscreen the resizing of the occupants does not work anymore. I think this a features for both view_modes. You could change the following code but then the controlbox is scaled small in fullscreen mode.
`enabled (_converse) {
        return _converse.view_mode == 'overlayed';
    },`

I still have everything in my commit-history so i could move back quickly to separate modules.  But maybe you have a different idea.

jcbrand:
I think this just means that the converse-dragresize plugin should no longer be only for the overlayed view mode.

So we should have more fine-grained disabling of code in the plugin based on which view mode is active.

worlword:
I updated the pull request. I didn't know if I should make a new commit for this. I hope it's okay if I just put everything in the same commit again. 

The sidebar now stays closed on relaod. The width is now also saved on reload. I decided to save the width inside the controlbox since the controlbox is always there and the width can be used for all MUCs. This way the users have to do the resizing only once if they have some people with long names. This works best for us at least. Otherwise one could save the width in the MUC-Model.

I also moved the code back to converse-dragresize.js. 

Travis seeems to be having problems again. On my server all tests complete without errors.

jcbrand:
> I hope it's okay if I just put everything in the same commit again.

Yes, this is preferable, thanks.

> The sidebar now stays closed on relaod. The width is now also saved on reload.

Great.

> I decided to save the width inside the controlbox since the controlbox is always there

Currentlly `converse-dragresize` imports `converse-controlbox`, so it'll break if someone removes that plugin, but I don't actually see anything in the existing code (before your PR) that has a hard dependency on the controlbox.

We try to develop Converse.js in such a way so that where feasible, plugins can be removed entirely and everything else still works. For example, `converse-dragresize`, `converse-minimize`, `converse-omemo` and `converse-rosterview`, `converse-muc` and `converse-muc-views` can all be removed and you'll still have a functioning application (albeit with reduced features).

In the `embedded` view mode (with `singleton` set to `false`), you see only a single MUC or one-on-one chat, and no controlbox. So there's at least once use-case where the controlbox is not needed (and the plugin could be removed from the build) but where you still have drag resizing of the occupants sidebar.

So my point is that it's probably not safe to assume that the controlbox is always there. Perhaps that might practically be the case currently, but if we continue doing so, we're further preventing ourselves from making the `converse-controlbox` plugin removable in the future.

> and the width can be used for all MUCs. This way the users have to do the resizing only once if they have some people with long names.

When you have multiple MUCs open in `overlay` view mode and you resize only one MUCs sidebar, then the others stay the same, but then when you reload the page, suddenly the other MUCs's sidebars are the same length as the resized one. So, besides the issues with depending on the controlbox, I also think this behaviour is confusing and that it's better to store the width on the MUC's model.

You might as well argue the opposite way, that you only want to change the sidebar for one particular MUC and don't want all other MUCs' sidebars to be changed.

If the user wants to have *all* MUC sidebars bigger, then we can solve that problem with user configuration settings.

I've been wanting to add user settings (based on PEP) to Converse for a long time, but haven't done so yet. This would be a good example of a user setting, where you set the MUC occupants width in one place and its then applied in the whole app.

If the "webmaster" wants to set a particular width for the sidebar, they can do so in CSS, like is currently already the case with the default sizes of the chat boxes.

> I also moved the code back to converse-dragresize.js.

Thanks, I think this is much better now.

> Travis seeems to be having problems again. On my server all tests complete without errors.

I restarted it and it's green again.

Thanks for all your effort on this so far.

jcbrand:
One more thing, I see that when you reload the page in `overlayed` view mode, that the MUC sidebars are their default size and then only after a noticeable delay are they resized to the custom size.

I think this is because you're applying the new size in the `afterShown` method.

I very recently refactored the code around showing chat boxes a little bit and there's now a new event [beforeShowingChatView](https://github.com/conversejs/converse.js/blob/2fca442f57f64d1f0b174db21d879c170b866168/src/converse-chatview.js#L1326) which you can use instead of overriding the `afterShow` method.

This should avoid the noticeable resizing of the sidebars and using events is also preferable to using `overrides`.

worlword:
Hello JCBrand,

sorry for the long long delay. 
I updated the pull-request. I should have made all the changes you requested, 
Sadly there is still a problem with the saving/loading of the width. I listen to the beforeShowingChatView event, like you suggested, but the model does not contain the value occupants_width anymore. The values of width and height also seem to be the same. My guess would be that a new instance of the model will be created when the MUC is reopened and therefore the values are missing. But I could not figure out how to fix it. 
Other than that the feature works fine.

src/converse-dragresize.js
```diff
@@ -143,6 +159,81 @@ converse.plugins.add('converse-dragresize', {
...
+                const flyout = view.el;
+                const div = document.createElement('div');
+                div.innerHTML = '<div class="dragresize dragresize-roster-left"></div>';
+                flyout.insertBefore(
+                    div,
+                    flyout.firstChild
+                );
+            },
+
+            onStartHorizontalResizeRoster (ev) {
```
jcbrand: Don't refer to the occupants list as a roster, in XMPP roster means something very specific and we want to avoid confusion. The same applies to the `dragresize-roster-left` class above.

src/converse-dragresize.js
```diff
@@ -351,19 +439,38 @@ converse.plugins.add('converse-dragresize', {
...
+            const occupants_width_pixel = chatroomview.model.get('width_occupants');
+            const occupants_view = chatroomview.el.querySelector('.occupants');
+            if (occupants_view && occupants_width_pixel !== undefined) {
+                occupants_view.style.flex = "0 0 " + occupants_width_pixel + "px";
+            }
+        });
+
+        /*_converse.api.listen.on('chatBoxInitialized', view => {
+            window.addEventListener('resize', _.debounce(() => view.setDimensions(), 100));
+        });*/
```
jcbrand: Uncommented code needs to be removed.

jcbrand:
Hi @worlword 

> I updated the pull-request. I should have made all the changes you requested,
Sadly there is still a problem with the saving/loading of the width. I listen to the beforeShowingChatView event, like you suggested, but the model does not contain the value occupants_width anymore.

You probably mean `width_occupants`. I'm not sure what the issue is here, or whether there really is an issue. I guess I'll have to test your code sometime.

I left to more smallish review requests.

worlword:
Hello JCBrand,

I updated the pull-request with the changes you requested

jcbrand:
Thanks @worlword, I don't have much time currently but I'll test out this PR as soon as possible.

jcbrand:
I tested this feature out tonight and it seems to work fine. At first it didn't work because I use `fullscreen` `view_mode` and the `converse-dragresize` plugin is only enabled in `overlayed` `view_mode`.

This is actually a problem. We don't want to enable drag-resizing of chats in the other view modes. So I now think that the code for drag-resizing the occupants sidebar should actually go into `converse-muc-views`.

That way it'll work in `fullscreen` view mode as well.

worlword:
Well that was kinda my first implementation :D 

If I move it to `converse-muc-views` than I habe to also move all the mouse events into another plugin again so that I can use them. Otherwise the mouse-events will also be deactivated in `fullscreen`.
I hope that's ok with you

jcbrand:
> Well that was kinda my first implementation :D

Yeah I'm sorry. Doing code reviews is actually quite hard because I have to try and keep in mind all kinds of edge-cases as well as how the PR might affect future development.

I'll definitely make mistakes while doing so.

When I write code myself I also often try out multiple ideas before I end up with what I think is the best solution. Particularly figuring out whether something should go into a separate plugin or not can be harder than might seem at first glance.

worlword:
Just updated the pull-request. 

I moved the event-handlers into it's own plugin again. Also added the `_converse.allow_occupants_view_resizing` again. I don't know. I feel like we should users the option to turn the feature off. If you want the feature always active, than we can remove `_converse.allow_occupants_view_resizing`.
I moved the mouse-events into it's own plugin again because I don't know what performance-issues arise if there are multiple event-listeners to the `mousemove` event. If we create a second event-handler we don't need the extra plugin. But maybe you have other the add more mouse-events or keybinds. Than you could reuse the plugin :)
I moved the resizing code into `converse-muc-views` like you suggested.

Btw the test will propably fail again because of the two missing emojis.

worlword:
I updated the pull-request again. I forgot to clean up the dragresize-module...

poVoq:
Am I guessing correct that this could also be used to hide the participant list by default? That has been requested a few times, for example here: https://github.com/conversejs/converse.js/issues/1416

Anyways, and news on merging this?

jcbrand:
@worlword: Instead of spending more time going back and forth, I've merged your commit and then made some changes of my own.

Take a look at https://github.com/conversejs/converse.js/commit/b1c9af3ed0015e9fea6cc60b6c2c6a796a154332 to see what I did.

> Am I guessing correct that this could also be used to hide the participant list by default? 

@poVoq  That's something else, not covered by this PR.

poVoq:
Would it be possible to have a setting for not only "resizeable" on/off but also for the default size? So that one could for example set it to 0% by default?

jcbrand:
I'm not sure that's a good idea, would be better to just have an option to hide it by default.

PR #1656, opened 2019-07-29
Fixed Russian translation for "Groupchats"

s-sokolko:
Fixed Russian translation for "Groupchats". It is "Ко**мн**аты" instead of "Ко**нм**аты"


licaon-kter:
Why not use the special Weblate site?

s-sokolko:
@licaon-kter sorry. didn't get the docs right: I thouth the website was only for new translations.
I've made an offer on the website, so the MR can be closed if needed https://hosted.weblate.org/translate/conversejs/translations/ru/?checksum=f401a7f5394468ca#suggestions

jcbrand:
@s-sokolko: If you simply make a "suggestion" on Weblate, then someone else first needs to accept it. This is probably the right way to do things when there are multiple translators and/or when the translation in question is up for debate.

Just be aware that your suggestions won't automatically be used (if someone doesn't accept) them. So sometimes it's probably just better to just make the translation instead of making a suggestion.

s-sokolko:
Thanks, @jcbrand ! And is the process of "bugfixing" translations documented anywhere? In my case I found a simple typo (two letters scrambled). I'm a native Russian speaker, so I am sure about that. If I can do some better thing than just a suggestion - I am ready to help

jcbrand:
@s-sokolko Just fix the translations directly in Weblate (without making a "suggestion"). Thanks!

s-sokolko:
@s-sokolko Thank you! Done!

PR #1658, opened 2019-07-30
Remove useless rm

licaon-kter:
...else `make clean` will error out:
```
npm run clean

> converse.js@4.2.0 clean /home/user/converse.js
> rm -rf node_modules stamp-npm dist *.zip

rm -r lib bin include parts
rm: cannot remove 'lib': No such file or directory
rm: cannot remove 'bin': No such file or directory
rm: cannot remove 'include': No such file or directory
rm: cannot remove 'parts': No such file or directory
make: *** [Makefile:123: clean] Error 1
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,7 +120,6 @@ stamp-npm: $(LERNA) package.json package-lock.json src/headless/package.json
 .PHONY: clean
 clean:
 	npm run clean
-	rm -r lib bin include parts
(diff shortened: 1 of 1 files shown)


jcbrand:
It's not a useless `rm` those directories get created when generating the documentation.

The fix is to include `-f` after `rm`.

PR #1660, opened 2019-08-02
Shadow dom: TypeError: Cannot read property 'parentElement' of null caused by MinimizedChatsToggleView

reda-alaoui:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -57,6 +57,7 @@
 - #1579: Trim spaces at the beginning and end of a JID (when adding contact)
 - #1585: Upload files by pasting from clipboard
 - #1586: Not possible to kick someone with a space in their nickname
+- Bugfix: "Cannot read property 'parentElement' of null" in shadow dom
 
 ### Breaking changes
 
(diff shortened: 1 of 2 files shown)


reda-alaoui:
For the test, please tell me what form it should have and where I should put it.

jcbrand:
Thanks for your pull request.

Can you please explain more about the problem and why this fixes it?
Why is this error happening for this View, but not all the others?

I've used Converse.js in a web component with shadow DOM before and didn't have this problem.

It looks to me like Converse tries to get  `#toggle-minimized-chats` from the parent (light) DOM? Instead of the shadow DOM which it is actually in.

reda-alaoui:
I had this issue with Chrome 75 and conversejs 4.2.0.

> It looks to me like Converse tries to get #toggle-minimized-chats from the parent (light) DOM? Instead of the shadow DOM which it is actually in.

Yes indeed. I think this is the only converse.js module doing this. The others retrieve/create their element themselves by using the root element.

jcbrand:
Ok, I think I know why... it's the only view that specifies `el: #....`, so it's the only View that tells Backbone to go look for an element in the DOM that represents it.

I don't think a test is necessary.

Thanks again.

PR #1662, opened 2019-08-03
Allow to disable the user profile modal

reda-alaoui:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -58,6 +58,7 @@
 - #1585: Upload files by pasting from clipboard
 - #1586: Not possible to kick someone with a space in their nickname
 - Bugfix: "Cannot read property 'parentElement' of null" in shadow dom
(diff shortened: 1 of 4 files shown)


jcbrand:
There's a better way to disable the profile, just add `converse-profile` to the [blacklisted_plugins](https://conversejs.org/docs/html/configuration.html#blacklisted-plugins) setting.

reda-alaoui:
@jcbrand , I tried this first.
When I do this, the whole control box is blank.

jcbrand:
>  When I blacklist converse-profile, the whole control box is blank.

Then that should be fixed instead of adding another config variable.

reda-alaoui:
@jcbrand , `converse-profile` is responsible for displaying the whole profile_view.html.
For my need, I don't want to hide the whole profile view (e.g I want to keep the avatar display) , I just want to prevent the detailed profile modal to be opened by the end user because he is not allowed to modify his profile information in my application.
I think blacklisting `converse-profile` is not what I need.

PR #1663, opened 2019-08-03
Allow to disable the contact user details display

reda-alaoui:


jcbrand:
Thank you for making this pull request.

I think this is a change that can rather just be implemented in your own plugin instead of merging it into core and adding yet another configuration variable.

You could for example hide that button via CSS and to be extra safe override `showUserDetailsModal` from `ChatBoxView` to not do anything.

PR #1669, opened 2019-08-08
[XXXX] Added Pop Out Feature to Chat

henrychen11:

### DIFF
--- a/src/converse-controlbox.js
+++ b/src/converse-controlbox.js
@@ -7,33 +7,33 @@
 /**
  * @module converse-controlbox
  */
-import "converse-chatview";
-import "converse-profile";
-import "converse-rosterview";
-import _FormData from "formdata-polyfill";
-import bootstrap from "bootstrap.native";
-import converse from "@converse/headless/converse-core";
-import tpl_brand_heading from "templates/converse_brand_heading.html";
-import tpl_controlbox from "templates/controlbox.html";
-import tpl_controlbox_toggle from "templates/controlbox_toggle.html";
-import tpl_login_panel from "templates/login_panel.html";
+import 'converse-chatview';
+import 'converse-profile';
+import 'converse-rosterview';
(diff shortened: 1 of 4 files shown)


src/converse-controlbox.js
```diff
@@ -7,33 +7,33 @@
...
+import 'converse-chatview';
+import 'converse-profile';
+import 'converse-rosterview';
+import _FormData from 'formdata-polyfill';
+import bootstrap from 'bootstrap.native';
+import converse from '@converse/headless/converse-core';
+import tpl_brand_heading from 'templates/converse_brand_heading.html';
+import tpl_controlbox from 'templates/controlbox.html';
+import tpl_controlbox_toggle from 'templates/controlbox_toggle.html';
+import tpl_login_panel from 'templates/login_panel.html';
```
jcbrand: Looks like you ran this file through a code formatter (prettier?) which makes it impossible to see what changes you actually made to it.

Please add an exception for this repo to your code formatter.

src/templates/controlbox.html
```diff
@@ -2,6 +2,7 @@
     <div class="chat-head controlbox-head">
         {[ if (!o.sticky_controlbox) { ]}
             <a class="chatbox-btn close-chatbox-button fa fa-times"></a>
+            <a class="chatbox-btn popout-chatbox-button fa fa-arrows-alt"></a>
```
jcbrand: This button shouldn't be inside the condition, which is only to check whether a close button is relevant, if `sticky_controlbox` is `true`, then the controlbox can't be closed and the `close` button doesn't render.

We'll probably want to add a new configuration setting to determine whether the chat can pop out as well.

jcbrand:
Hi Henry, instead of working in a fork in your own github profile, please make a branch inside this organization (https://github.com/conversejs ), that way I can also push changes if need be.

PR #1672, opened 2019-08-09
Fix fonts path back

licaon-kter:
...as documented in https://github.com/conversejs/converse.js/blob/master/docs/source/quickstart.rst#L97 else they'll need to be in `website.tld/dist/dist/webfonts` instead

Partially reverts https://github.com/conversejs/converse.js/commit/a116a1ec8f8fbe01834c94f97043af439adfe412

### DIFF
--- a/sass/font-awesome.scss
+++ b/sass/font-awesome.scss
@@ -2,38 +2,38 @@
   font-family: 'Baumans';
   font-style: normal;
   font-weight: 400;
-  src: local('Baumans Regular'), local('Baumans-Regular'), url('../dist/webfonts/baumans.ttf') format('truetype');
+  src: local('Baumans Regular'), local('Baumans-Regular'), url('../webfonts/baumans.ttf') format('truetype');
 }
 
 @font-face {
   font-family: 'Muli';
   font-style: normal;
   font-weight: 400;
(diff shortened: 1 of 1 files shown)


jcbrand:
The webfonts indeed should be in the `dist` folder and I added webpack configuration to copy them there (it might not be working properly currently, not sure).

The quickstart guide says webfonts should be in the same dir as the CSS file, which is `dist`.

jcbrand:
Here's the webpack config: https://github.com/conversejs/converse.js/blob/master/webpack.config.js#L48

licaon-kter:
Umm, ok, whatever the problem is, this PR fixes it, if you want to fix it differently, go ahead. :)

jcbrand:
You didn't describe what the actual problem is.

licaon-kter:
It's there in the OP... :P

With HEAD it tries to load the fonts from `domain.tld/dist/dist/webfonts/filename` which doesn't exist since the `webfonts` folder, the .css and .js files are in `/dist/`

licaon-kter:
Ok so the fonts are now in /dist/dist/ ? Should I PR the docs then ?

jcbrand:
> Ok so the fonts are now in /dist/dist/

I don't know how I managed to do that... should be fixed in master

Issue #1679, opened 2019-08-16
Room invitation fails with singleton and random server assigned room


deleolajide:
**Describe the bug**
When **auto_join_private_chats**,  **singleton** and **auto_join_on_invite** are set and a conference invite is received for a random server-generated conference room , the invitation is rejected and a message is shown in the console log

```
xxxx@conference.domain is not allowed because singleton is true and it's not being auto_joined
```

**To Reproduce**
```
    converse.initialize({
        auto_login: true,
        theme: 'concord',
        allow_non_roster_messaging: true,
        auto_join_on_invite: true,
        auto_join_private_chats: ['demo@workgroup.domain.com'],
        authentication: 'anonymous',
        jid: 'domain.com',
        auto_away: 300,
        auto_reconnect: true,
        debug: true,
        singleton: true,
        websocket_url: 'wss://domain.com/ws/',
        message_archiving: 'always'
    });
```

**Expected behavior**
Converse is validating the conference room against the **auto_join_rooms** list which is empty. As there is no way to specify the generated room name, the validation should accept any room name if all **auto_join_private_chats**,  **singleton** and **auto_join_on_invite** are set.

**Screenshots**
not applicable

**Environment (please complete the following information):**
 - Desktop?
 - Browser [Chrome]
 - Converse.js version [e.g. 5.0.1]

**Additional context**
```
validate (attrs, options) {
    if (!attrs.jid) {
        return 'Ignored ChatBox without JID';
    }
    // my fix
    if (_converse.auto_join_on_invite && _converse.singleton) return null;

    const room_jids = _converse.auto_join_rooms.map(s => _.isObject(s) ? s.jid : s);
    const auto_join = _converse.auto_join_private_chats.concat(room_jids);
    if (_converse.singleton && !_.includes(auto_join, attrs.jid)) {
        const msg = `${attrs.jid} is not allowed because singleton is true and it's not being auto_joined`;
        _converse.log(msg, Strophe.LogLevel.WARN);
        return msg;
    }
},
```
```

PR #1681, opened 2019-08-16
Fix Issue #1679

deleolajide:
This fix forces Converse.js to respect auto_join_rooms and allow auto join room to occur with server-side generated random room names as required for XEP 0142 - workgroup queues

### DIFF
--- a/src/headless/converse-chatboxes.js
+++ b/src/headless/converse-chatboxes.js
@@ -405,7 +405,7 @@ converse.plugins.add('converse-chatboxes', {
                 }
                 const room_jids = _converse.auto_join_rooms.map(s => _.isObject(s) ? s.jid : s);
                 const auto_join = _converse.auto_join_private_chats.concat(room_jids);
-                if (_converse.singleton && !_.includes(auto_join, attrs.jid)) {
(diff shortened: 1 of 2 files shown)


src/headless/converse-chatboxes.js
```diff
@@ -405,7 +405,7 @@ converse.plugins.add('converse-chatboxes', {
                 }
                 const room_jids = _converse.auto_join_rooms.map(s => _.isObject(s) ? s.jid : s);
                 const auto_join = _converse.auto_join_private_chats.concat(room_jids);
-                if (_converse.singleton && !_.includes(auto_join, attrs.jid)) {
+                if (_converse.singleton && !_.includes(auto_join, attrs.jid) && !_converse.auto_join_on_invite) {
```
jcbrand: Thanks @deleolajide. Please remove the trailing whitespace and also add an entry to the changelog.

deleolajide: I have updated changelog. Not sure about white spaces. I made changes directly from github.com website. I can't tell if they are removed or not.

jcbrand:
Thanks!

PR #1689, opened 2019-08-26
Fix incorrectly detecting every message with a duplicate ID as correction

laszlovl:
Some clients don't use globally unique message ID's, but probably restart their sequence every time they (re)start. Since `correctMessage()` only checked for `id` and `from`, this would make completely unrelated messages from the same user overwrite each other's contents.

Tested that the fix prevents the problem and still properly handles actual corrections.

### DIFF
--- a/src/headless/converse-chatboxes.js
+++ b/src/headless/converse-chatboxes.js
@@ -549,10 +549,10 @@ converse.plugins.add('converse-chatboxes', {
              *     {@link _converse.ChatBox.getMessageAttributesFromStanza}
              */
             correctMessage (attrs) {
-                if (!attrs.msgid || !attrs.from) {
+                if (!attrs.replaced_id || !attrs.from) {
                     return;
                 }
-                const message = this.messages.findWhere({'msgid': attrs.msgid, 'from': attrs.from});
+                const message = this.messages.findWhere({'msgid': attrs.replaced_id, 'from': attrs.from});
                 if (!message) {
                     return;
                 }
@@ -969,6 +969,7 @@ converse.plugins.add('converse-chatboxes', {
                     'is_spoiler': !!spoiler,
                     'message': text,
                     'msgid': msgid,
+                    'replaced_id': replaced_id,
                     'references': this.getReferencesFromStanza(stanza),
                     'subject': _.propertyOf(stanza.querySelector('subject'))('textContent'),
                     'thread': _.propertyOf(stanza.querySelector('thread'))('textContent'),
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 5.0.4
+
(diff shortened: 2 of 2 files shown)


marclaporte:
Just curious: which clients?

src/headless/converse-chatboxes.js
```diff
@@ -964,6 +966,7 @@ converse.plugins.add('converse-chatboxes', {
                     'is_archived': this.isArchived(original_stanza),
                     'is_delayed': !!delay,
                     'is_spoiler': !!spoiler,
+                    'is_replace': !!replaced_id,
```
jcbrand: `is_replace` is grammatically incorrect and I think it's less helpful to store a boolean than storing the actual ID of the message being replaced.

So instead of adding `is_replace`, use `replaced_id` directly.

Then in `correctMessage` on line 555 please change

    const message = this.messages.findWhere({'msgid': attrs.msgid, 'from': attrs.from});

to 

     const message = this.messages.findWhere({'msgid': attrs.replaced_id, 'from': attrs.from});

I'm not sure why I didn't initially do it like this, it now seems so obvious. Hopefully I'm not missing some important detail.

laszlovl: With this new logic, does it still make sense to set the replaced_id as msgid? (`const msgid = replaced_id || stanza.getAttribute('id') || original_stanza.getAttribute('id');`)

jcbrand: I think so yes. We at this point don't yet know whether we have the message being replaced and `replaced_id` points to that messsage's ID. If such a message doesn't exist, we use the data from this message to create it.

jcbrand: In fact, all corrections and the original corrected message are all collapsed into one Message object, but the point still stands.

laszlovl:
I've been looking at this, but message corrections now appear to be broken in general - regardless of these changes. I believe this wasn't the case when I originally tested it two weeks ago, but perhaps my setup is now slightly different.

Either way, the message correction is now caught by the "duplicate message check":

```js
                const message = await this.getDuplicateMessage(original_stanza);
                if (message) {
                    this.updateMessage(message, original_stanza);
                }
                if (message ||
                        this.isReceipt(stanza) ||
                        this.isChatMarker(stanza)) {
                    return _converse.api.trigger('message', {'stanza': original_stanza});
                }
```

which means it'll never even reach `this.correctMessage()`.

laszlovl:
The problem above was actually caused by an older (4.x) version of Converse.js sending out invalid origin ID's for message corrections.

src/headless/converse-chatboxes.js
laszlovl: OK done, and verified that message corrections still work.

jcbrand:
Dank je wel!

PR #1696, opened 2019-08-30
Fix muc_domain link

licaon-kter:
...since `locked_domain` needs to be "domain.tld" and can't be just `true`... or maybe `locked_domain` needs editing instead?


jcbrand:
Thanks!

PR #1706, opened 2019-09-11
Update Oauth login to current API format

rgrunbla:
Hey,

The login API format changed a while back, and the oauth module wasn't updated. This should fix it.
As the module is not being currently used, I didn't put anything in CHANGES.md (as it's mostly minor).


jcbrand:
Thanks!

PR #1707, opened 2019-09-11
Capitalize "groupchat"

licaon-kter:


jcbrand:
Thanks but you'll need to fix the failing tests as well. Just grep for those strings in the `./spec` directory and fix.

jcbrand:
Thanks!

PR #1723, opened 2019-09-19
Process stanzas from mam one-by-one

ChaosKid42:
Process stanzas coming from mam one-by-one in order to fix processing of message receipts.


ChaosKid42:
@jcbrand Had to add a "try-catch" block in order to prevent it from processing mam messages in case an exception occurs.

jcbrand:
Thanks!

PR #1728, opened 2019-09-23
allow to inject stanzas for debugging

ChaosKid42:
This PR allows to inject stanzas for debugging purposes if debugging is set to `true`.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -604,6 +604,11 @@ debug
 
 If set to ``true``, debugging output will be logged to the browser console.
 
+Additionally the ``/inject`` command will be enabled, which allows you to inject message
+stanzas for debugging purposes. The stanzas don't actually get sent out but receiving of
+these stanzas is mimicked. This allows for example for replaying message stanzas logged to
+the browser console before.
+
 You can also set this value by changing the URL fragment to `#converse?debug=true` or `#converse?debug=false`.
 
 
--- a/src/converse-chatview.js
+++ b/src/converse-chatview.js
@@ -763,18 +763,26 @@ converse.plugins.add('converse-chatview', {
             },
 
             parseMessageForCommands (text) {
-                const match = text.replace(/^\s*/, "").match(/^\/(.*)\s*$/);
-                if (match) {
-                    if (match[1] === "clear") {
+                text = text.replace(/^\s*/, "");
+                const command = (text.match(/^\/([a-zA-Z]*) ?/) || ['']).pop().toLowerCase();
+                if (command) {
+                    const args = text.slice(('/'+command).length+1).trim();
+                    if (command === "clear") {
                         this.clearMessages();
                         return true;
+                    } else if (command === "inject") {
+                        this.handleDebugTextAsReceivedMessages(args);
+                        return true;
                     }
-                    else if (match[1] === "help") {
+                    else if (command === "help") {
                         const msgs = [
                             `<strong>/clear</strong>: ${__('Remove messages')}`,
                             `<strong>/me</strong>: ${__('Write in the third person')}`,
                             `<strong>/help</strong>: ${__('Show this menu')}`
                             ];
+                        if (_converse.debug) {
+                            msgs.push(`<strong>/inject</strong>: ${__('Inject stanzas for debugging purposes')}`);
+                        }
                         this.showHelpMessages(msgs);
                         return true;
                     }
@@ -1014,6 +1022,25 @@ converse.plugins.add('converse-chatview', {
                 const result = confirm(__("Are you sure you want to clear the messages from this conversation?"));
                 if (result === true) {
                     this.model.clearMessages();
+                }
+                return this;
+            },
+
+            async handleDebugTextAsReceivedMessages (text) {
+                if (!_converse.debug) {
+                    this.showHelpMessages([__('Inject command not available because debugging turned off.')], 'error');
+                    return this;
+                }
+                const msgs = new DOMParser().parseFromString('<s>'+text+'</s>', 'text/xml').firstElementChild.children;
+                /* eslint-disable no-await-in-loop */
+                for (const msg of msgs) {
+                    try {
+                        await _converse.chatboxes.onMessage(msg);
+                   } catch (e) {
+                        _converse.log(e, Strophe.LogLevel.ERROR);
+                        _converse.log('Message that lead to that error:', Strophe.LogLevel.ERROR);
+                        _converse.log(msg, Strophe.LogLevel.ERROR);
+                    }
                 }
                 return this;
             },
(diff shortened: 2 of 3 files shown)


src/converse-chatview.js
```diff
@@ -763,18 +763,26 @@ converse.plugins.add('converse-chatview', {
...
-                if (match) {
-                    if (match[1] === "clear") {
+                text = text.replace(/^\s*/, "");
+                const command = (text.match(/^\/([a-zA-Z]*) ?/) || ['']).pop().toLowerCase();
+                if (command) {
+                    const args = text.slice(('/'+command).length+1).trim();
+                    if (command === "clear") {
                         this.clearMessages();
                         return true;
+                    } else if (command === "inject") {
```
jcbrand: Here you need to check if debugging is turned on.

src/converse-chatview.js
```diff
@@ -1018,6 +1026,25 @@ converse.plugins.add('converse-chatview', {
                 return this;
             },
 
+            async inject (ev) {
+                if (!_converse.debug) {
+                    return this;
```
jcbrand: I think this should raise an error, or at least log an error.

src/converse-chatview.js
```diff
@@ -1018,6 +1026,25 @@ converse.plugins.add('converse-chatview', {
...
+                }
+                const message_handler = _converse.chatboxes.onMessage.bind(_converse.chatboxes);
```
jcbrand: You don't have to assign to `message_handler`, just call `_converse.chatboxes.onMessage` directly further down below.

src/converse-chatview.js
```diff
@@ -1018,6 +1026,25 @@ converse.plugins.add('converse-chatview', {
                 return this;
             },
 
+            async inject (ev) {
```
jcbrand: `ev` is generally used to refer to an Event, looks like a string is passed in here?

src/converse-chatview.js
```diff
@@ -1018,6 +1026,25 @@ converse.plugins.add('converse-chatview', {
...
+                if (!_converse.debug) {
+                    return this;
+                }
+                const message_handler = _converse.chatboxes.onMessage.bind(_converse.chatboxes);
+                const msgs = new DOMParser().parseFromString('<s>'+ev+'</s>', 'text/xml').children[0].children;
```
jcbrand: `firstElementChild` might be better than `children[0]`. I think `children[0]` might return a text node or other non-Element nodes.

src/converse-chatview.js
ChaosKid42: I'd rather not check it here. Otherwise you might end up accidently entering raw xmpp-stanzas in a chat because you forgot to turn on debugging.

jcbrand: Makes sense, then an error message needs to be shown informing the user that debug mode is not on.

ChaosKid42: Done.

docs/source/configuration.rst
```diff
@@ -602,7 +602,8 @@ debug
 
 * Default:  ``false``
 
-If set to ``true``, debugging output will be logged to the browser console.
+If set to ``true``, debugging output will be logged to the browser console. Additionally the ``/inject`` command will be enabled, which allows to inject stanzas for debugging
```
jcbrand: allows *you* to

jcbrand: The code only supports injecting `<message>` stanzas, which is not made clear here.

ChaosKid42: Updated the documentation. Hope it's clearer now.

src/converse-chatview.js
```diff
@@ -1018,6 +1026,25 @@ converse.plugins.add('converse-chatview', {
                 return this;
             },
 
+            async inject (stanzas) {
+                if (!_converse.debug) {
```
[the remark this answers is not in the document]
jcbrand: Thinking about this a bit more, this feature wouldn't actually have helped you solve the issue with receipts from MAM being processed correctly because you `await` in the `inject` method and that was the problem in the MAM code and this code would have just masked the problem.

Don't be mad at me, but I'm leaning towards rejecting this PR and instead just ask that you write a test in the future, unless you can make a good argument why it's worth having this feature (which needs support going forward and can break when we refactor).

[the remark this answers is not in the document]
ChaosKid42: I'm absolutely not mad at you. I can follow your reasoning. However I used that feature for tracking down another bug unrelated to the MAM issue. I found that empty messages with just chat-states (see example below) would generate errors. I still think that this feature might be useful for tracking down these kinds of errors:

```xml
<message from="csztest@scholzbande.de" to="csztest@scholzbande.de/converse.js-9009601" xmlns="jabber:client">
    <result id="1566456555786157" queryid="35fcabfb-bbe9-4b9d-957a-83bb1c4df8a8" xmlns="urn:xmpp:mam:2">
        <forwarded xmlns="urn:xmpp:forward:0">
            <message from="csztest@scholzbande.de/converse.js-52677026" id="5b8e6d92-d1ff-4195-afdd-958df0d518c7" to="christoph@scholzbande.de" type="chat" xml:lang="de" xmlns="jabber:client">
                <archived by="csztest@scholzbande.de" id="1566456555786157" xmlns="urn:xmpp:mam:tmp"/>
                <stanza-id by="csztest@scholzbande.de" id="1566456555786157" xmlns="urn:xmpp:sid:0"/>
                <active xmlns="http://jabber.org/protocol/chatstates"/>
                <received id="aa466f2b-2943-4109-9da6-7220285403b6" xmlns="urn:xmpp:receipts"/>
                <store xmlns="urn:xmpp:hints"/>
            </message>
            <delay from="scholzbande.de" stamp="2019-08-22T06:49:15.786157Z" xmlns="urn:xmpp:delay"/>
        </forwarded>
    </result>
</message>
```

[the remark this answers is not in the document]
jcbrand: Ok, so in my opinion, the best way to track down and fix such a bug is to write a test that passes in this message and then to check what happens.

For example, here's a test that checks that a malformed message is rejected:
https://github.com/conversejs/converse.js/blob/bce8dc9113e802e184e31ab52272a5bd3bebf164/spec/muc_messages.js#L14

You could fairly easily create a similar test and just replace the stanza with your own.

This has the added benefit of automatically adding a regression test for any fix you make.

Concerning this PR, I'm going to reject it because it's IMO a very narrow use-case and because it adds more code that needs to be kept up to date and maintained.

You can still put all these changes in a 3rd party plugin. You can use `overrides` to modify `parseMessageForCommands` and the `inject` method can  be assigned with `Object.assign` to `_converse.ChatBox.prototype`. There are various examples of this in the core plugins. Feel free to ask if it's still unclear how to do this.

You can put that plugin in the [community-plugins](https://github.com/conversejs/community-plugins) repo. Maybe eventually other debug functions can be added to it as well.

Thanks for making the PR and for implementing my suggestions.

Issue #1714, opened 2019-09-13
Notification sound playing erroneusly


licaon-kter:
HEAD (fe34b7eaa09010ff289cd2adb94525ab11f783c6), Firefox 69

1:1 chat
User A Converse
User A Gajim
User B Conversations

User A writes in Converse, Converse plays a sound (user A presses enter in Firefox and Firefox makes a sound)
User A writes in Gajim, no sound
User B writes in Conversations, no sound

This appears to be reversed, I should be getting notifications ONLY when user B is writing, any other time (user A in Converse or user A in Gajim) Firefox should not notify/make a sound.

licaon-kter:
Yup, appears fixed, thanks.

PR #1730, opened 2019-09-24
don't notify if only message delivery receipt

ChaosKid42:
This PR prevents notification sounds in case we receive a message delivery receipt only.


licaon-kter:
This should fix https://github.com/conversejs/converse.js/issues/1714 ?

jcbrand:
> This should fix #1714 ?

Indeed, @ChaosKid42 could you please add `fixes #1714` in the commit message and mention that ticket number in the changelog?

ChaosKid42:
Mentioning #1714 now.

PR #1731, opened 2019-09-25
only send bookmark stanza if connected

ChaosKid42:
I always see the following error message when logging out:

error_improper_addressing

I found that this is due to the fact, that converse tries to send a bookmark stanza even after the connection is down.

This PR prevents this and thus the error message is gone.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,8 @@
 - Bugfix: process stanzas from mam one-by-one in order to correctly process message
   receipts
 - #1714 Bugfix: Don't notify the user in case we're receiving a message delivery receipt only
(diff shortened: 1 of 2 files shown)


jcbrand:
Looks to me like there is a deeper underlying problem here. Why is Converse trying to send out a bookmark stanza during logout? Probably because models are being removed during teardown and Converse thinks the user deleted them and that they should therefore be removed on the server as well.

Can you please cherry pick this commit and see whether it fixes the issue for you?
https://github.com/conversejs/converse.js/commit/6d53405748fa468d53217e36128bb916e6854f94

ChaosKid42:
Your solution fixes my issue. Thanks!

jcbrand:
Cool, thanks for checking.

PR #1732, opened 2019-09-25
add message-forwarding functionality

worlword:
Adds a new action button to messages in the chat-history for message-
forwarding. The Button opens a modal dialog where the user can enter
a destination for the message. The destination can be a user in the
roster or a MUC that is currently opened.

So another pull-request XD

The following screenshot shows how forwarded messages are displayed in chat (I just saw that the text is in german but it should be understandable).
message_forwarding

The user can forward any message in the chat-history. During mouse-over a new button will be displayed. When the user clicks on the button the following dialog will be displayed.
message_forwarding_modal

### DIFF
--- a/spec/messages.js
+++ b/spec/messages.js
@@ -17,6 +17,7 @@
                 ['rosterGroupsFetched', 'chatBoxesFetched'], {},
                 async function (done, _converse) {
 
+            _converse.whitelisted_plugins = _converse.whitelisted_plugins.filter(e => e !== 'converse-forward-message');
             await test_utils.waitForRoster(_converse, 'current', 2);
             const contact_jid = mock.cur_names[0].replace(/ /g,'.').toLowerCase() + '@montague.lit';
             const forwarded_contact_jid = mock.cur_names[1].replace(/ /g,'.').toLowerCase() + '@montague.lit';
@@ -83,7 +84,7 @@
             expect(textarea.value).toBe('');
 
             const first_msg = view.model.messages.findWhere({'message': 'But soft, what light through yonder airlock breaks?'});
-            expect(view.el.querySelectorAll('.chat-msg .chat-msg__action').length).toBe(1);
+            expect(view.el.querySelectorAll('.chat-msg .chat-msg__action').length).toBe(2);
             let action = view.el.querySelector('.chat-msg .chat-msg__action');
             expect(action.getAttribute('title')).toBe('Edit this message');
 
@@ -160,7 +161,7 @@
                 .c('active', {'xmlns': 'http://jabber.org/protocol/chatstates'}).tree()
             );
             await new Promise(resolve => view.once('messageInserted', resolve));
-            expect(view.el.querySelectorAll('.chat-msg .chat-msg__action').length).toBe(1);
+            expect(view.el.querySelectorAll('.chat-msg .chat-msg__action').length).toBe(3);
 
             // Test confirmation dialog
             spyOn(window, 'confirm').and.returnValue(true);
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## 6.0.0 (Unreleased)
 
+- add message-forward (XEP-0297) capabilities
 - #129: Add support for XEP-0156: Disovering Alternative XMPP Connection Methods. Only XML is supported for now.
 - #1105: Preliminary support for storing persistent data in IndexedDB instead of localStorage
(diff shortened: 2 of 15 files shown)


jcbrand:
This is great, thanks @worlword! Can you please look at the two issues the LGTM bot raised?

I'll look into this ASAP.

worlword:
hey, 

i already updated :) Everything should be fixed.

src/headless/converse-chatboxes.js
```diff
@@ -1191,13 +1190,6 @@ converse.plugins.add('converse-chatboxes', {
                     );
                 }
 
-                const bare_forward = sizzle(`message > forwarded[xmlns="${Strophe.NS.FORWARD}"]`, stanza).length;
-                if (bare_forward) {
-                    return this.rejectMessage(
-                        stanza,
-                        'Forwarded messages not part of an encapsulating protocol are not supported'
-                    );
-                }
```
jcbrand: This is dangerous, because if the `forward-message` plugin is not loaded, then removing this code causes a security vulnerability which allows malicious users to spoof messages as if they're coming from someone else.

I'm not sure currently what the right way is to go about adding this feature. One option is to make message forwarding a more "core" feature such as message corrections and spoilers, meaning that it doesn't go into its own plugin but instead goes into `converse-chatboxes` and `converse-chatview`. However I'm not yet convinced that this is the right way.

I've been thinking about using [JXT] from @legastero to convert incoming XML stanzas to JSON. I think if we did that, we could let the `forward-message` plugin register a JXT definition for message forwarding. That way, if the plugin is not loaded, forwarded messages are dropped because they can't be parsed and we avoid this problem. That's at least the theory.

worlword: Hm. I've been thinking about this. I haven't tested this, but would it be possible to register our own strophe.js-handler (http://strophe.im/strophejs/doc/1.1.3/files/strophe-js.html#Strophe.Connection.addHandler) inside the plugin? That way the handler should not be registered if the plugin is missing and forwarded messages should be dropped again. When the plugin is active, the handler takes the stanza and calls the new functions. If it is not a forwarded message it will be dropped (I don't know if strophe.js allows to register a handler just for XEP-0297-stanzas). 
I can't test this right now, but I think this could be a solution without relying on JXT. What do you think?

jcbrand: It doesn't look like you can register a handler only for forwarded messages, however you could register a more broader handler and then inside the handler check if the message is a forwarded one.

worlword:
I just updated the pull-request. Sorry for the delay I was busy the past weeks. 
I use my own handler now and moved almost all code into `converse-message-forwarding`. I copied most of your code to handle the incoming stanzas. I wanted to avaid code duplication but the module breaks very quickly when the code inside `converse-muc` or `converse-chat` is changed. 

I don't know why I get the error in the emoji-test, I hoped it was because of my local configuration... I shouldn't have changed anything there. But I don't get the XEP-0363 errors when I test it locally. I have to look into that tomorrow

spec/messages.js
```diff
@@ -17,6 +17,7 @@
                 ['rosterGroupsFetched', 'chatBoxesFetched'], {},
                 async function (done, _converse) {
 
+            _converse.whitelisted_plugins = _converse.whitelisted_plugins.filter(e => e !== 'converse-forward-message');
```
jcbrand: The second parameter of `initConverse` 3 lines above is a settings map.
So you can pass in `{blacklisted_plugins: ['converse-forward-message']}` there instead of doing this here.

jcbrand (changes requested):
Thanks, looks pretty good, but I'll need to look at it in more detail later.

I've noticed a small thing for which I've made a suggestion.

spec/messages.js
worlword: I tried it. Then the test fails again. The problem is, that I check with the following if the plugin is enabled:
`if (!_converse.whitelisted_plugins.includes('converse-forward-message'))`
So when I add it to the blacklisted plugins it is also a whitelisted plugin. That's why it will not reject the message and the test fails. I'm thinking about also adding `|| _converse.blacklisted_plugins.includes('converse-forward-message')` to the if-statement. I don't really like the solution anyways, but could not find a better way from preventing converse to send an error-message on forwarded-messages, if the plugin is enabled.

I also tried to find out why the tests for xep-0363. For me all tests work and I cannot find anything in the code (by just looking) that jumps out. I updated the pull-Request anyway to the latest commit.

PR #1736, opened 2019-09-29
new config option: allow_message_corrections

ChaosKid42:
I'd like to be able to configure Converse for allowing edits of the message sent most recently.

This PR introduces a new config option `allow_message_corrections` which makes this possible.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -780,6 +780,17 @@ keepalive
 
 Determines whether Converse will attempt to keep you logged in across page loads.
 
+.. _`allow_message_corrections`:
+
+allow_message_corrections
+-------------------------
+
+* Default:  ``'all'``
+
+Configures the last message correction (LMC) feature of Converse. By default you can edit all of your own
+messages. Setting this to ``'last'`` will limit this feature to the message sent most recently as suggested by
+`XEP-0308: Last Message Correction <https://xmpp.org/extensions/xep-0308.html>`_.
+Setting it to anything else (including ``false``) will disable the ability to correct sent messages.
 
 .. _`locales`:
 
--- a/src/headless/converse-chatboxes.js
+++ b/src/headless/converse-chatboxes.js
@@ -41,6 +41,7 @@ converse.plugins.add('converse-chatboxes', {
             'auto_join_private_chats': [],
             'clear_messages_on_reconnection': false,
             'filter_by_resource': false,
+            'allow_message_corrections': 'all',
             'send_chat_state_notifications': true
         });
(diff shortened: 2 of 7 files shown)


src/headless/converse-chatboxes.js
```diff
@@ -765,6 +767,32 @@ converse.plugins.add('converse-chatboxes', {
...
+                }
+                if (!_converse.lmc_only_for_last) {
+                    attrs.editable = true;
+                    return;
+                }
+                if (send_time > this.get('time_last_sent_msg')) {
+                    this.set({'time_last_sent_msg': send_time});
+                    this.messages.where({'editable': true}).forEach(msg => {
+                        msg.save({'editable': false});
+                    });
```
jcbrand: What I like about having an `editable` attribute is that we can also later use it in semi-anon MUCs together with [XEP-0421](https://xmpp.org/extensions/xep-0421.html) to determine whether a message from the same nick is actually from the same person and therefore whether they're allowed to edit it.

If `time_last_sent_msg` is `true`, then at any one time only one message should have `editable` to `true`, which means you can use `this.messages.findWhere` which stops iterating once the first match has been found.

Note, `findWhere` returns the message itself, not an array.

src/headless/converse-chatboxes.js
```diff
@@ -1258,6 +1287,9 @@ converse.plugins.add('converse-chatboxes', {
                             !chatbox.handleChatMarker(stanza, from_jid, is_carbon, is_roster_contact, is_mam)) {
 
                         const attrs = await chatbox.getMessageAttributesFromStanza(stanza, original_stanza);
+                        if (!utils.isHeadlineMessage(_converse, stanza)) {
```
jcbrand: I think the `isHeadlineMessage` check should go into `set_editable`, that way you don't have to always remember doing this check before calling it.

docs/source/configuration.rst
```diff
@@ -780,6 +780,16 @@ keepalive
 
 Determines whether Converse will attempt to keep you logged in across page loads.
 
+.. _`lmc_only_for_last`:
+
+lmc_only_for_last
+-----------------
```
jcbrand: I'm not a fan of the name `lmc_only_for_last`. If you expand it, it reads "Last message correction only for last" which is weird and it also requires people to know what "lmc" means.

How about `allow_message_corrections` which takes 3 values: `'all'`, `'last'` or anything falsy?

ChaosKid42:
Thanks for your review @jcbrand. Tried to implement all of your suggestions.

PR #1748, opened 2019-10-08
Startin blox autocomplete pr

deleolajide:
- New config option [muc_mention_autocomplete_filter](https://conversejs.org/docs/html/configuration.html#muc_mention_autocomplete_filter)
- Fix issue [Add user pictures in mention autocomplete list](https://git.happy-dev.fr/startinblox/components/sib-chat/issues/50).
- Fix issue [Mention names autocomplete](https://git.happy-dev.fr/startinblox/components/sib-chat/issues/95).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 5.0.5
+
+- New config option [muc_mention_autocomplete_filter](https://conversejs.org/docs/html/configuration.html#muc_mention_autocomplete_filter)
+- Fix issue [Add user pictures in mention autocomplete list](https://git.happy-dev.fr/startinblox/components/sib-chat/issues/50).
+- Fix issue [Mention names autocomplete](https://git.happy-dev.fr/startinblox/components/sib-chat/issues/95).
+
 ## 5.0.4 (2019-10-08)
 
 - New config option [allow_message_corrections](https://conversejs.org/docs/html/configuration.html#allow_message_corrections)
--- a/src/converse-autocomplete.js
+++ b/src/converse-autocomplete.js
@@ -40,6 +40,19 @@ converse.plugins.add("converse-autocomplete", {
             const element = document.createElement("li");
             element.setAttribute("aria-selected", "false");
 
+            const img = document.createElement("img");
+            let dataUri = "data:" + _converse.DEFAULT_IMAGE_TYPE + ";base64," + _converse.DEFAULT_IMAGE;
+
+            if (_converse.vcards) {
+                const vcard = _converse.vcards.findWhere({'nickname': text});
+                if (vcard) dataUri = "data:" + vcard.get('image_type') + ";base64," + vcard.get('image');
+            }
+
+            img.setAttribute("src", dataUri);
+            img.setAttribute("width", "22");
+            img.setAttribute("class", "avatar avatar-autocomplete");
+            element.appendChild(img);
+
             const regex = new RegExp("("+input+")", "ig");
             const parts = input ? text.split(regex) : [text];
             parts.forEach((txt) => {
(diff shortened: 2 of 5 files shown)


src/converse-autocomplete.js
```diff
@@ -40,6 +40,19 @@ converse.plugins.add("converse-autocomplete", {
...
+            if (_converse.vcards) {
+                const vcard = _converse.vcards.findWhere({'nickname': text});
+                if (vcard) dataUri = "data:" + vcard.get('image_type') + ";base64," + vcard.get('image');
+            }
+
+            img.setAttribute("src", dataUri);
+            img.setAttribute("width", "22");
+            img.setAttribute("class", "avatar avatar-autocomplete");
+            element.appendChild(img);
+
```
jcbrand: This code is too specific to the MUC nickname autocomplete use-case, which is a subset of all different autocomplete implementations, and should therefore not be here. The autocomplete plugin is meant to be generic and we cannot assume that the text values are nicknames or that we want to show avatars for them.

So this code needs to be put into a different function (e.g. MUC_NICKNAME) which can in turn call ITEM and then it should be possible to pass that function in to the `AutoComplete` constructor in the `config` map so that it can be used instead of ITEM (which can stay the default).

jcbrand:
The tests are failing because I recently changed the eslint config to make it stricter, I'll disable that again soon.

CHANGES.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## 5.0.5
+
+- New config option [muc_mention_autocomplete_filter](https://conversejs.org/docs/html/configuration.html#muc_mention_autocomplete_filter)
+- Fix issue [Add user pictures in mention autocomplete list](https://git.happy-dev.fr/startinblox/components/sib-chat/issues/50).
+- Fix issue [Mention names autocomplete](https://git.happy-dev.fr/startinblox/components/sib-chat/issues/95).
```
jcbrand: Please don't link to issues from another project, it's not directly related to core Converse development.

deleolajide: > Please don't link to issues from another project, it's not directly related to core Converse development.

I appreciate that, but duplicating the issues on converse is an additional administrative task that would be nice to avoid. Does this mean I should stop linking Pade issues to Converse and vice-versa?

The ability to cross-reference the issues may prove to be very useful when an issue is revisited sometime in the future and everyone has forgotten the linkage especially if the project has public access.

src/converse-autocomplete.js
deleolajide: OK. I shall take another pass at this. Thanks

CHANGES.md
jcbrand: I think it's OK to link to other issues where relevant, but it's more about *how* one does it.

Many of the bugs fixed and features implemented for this project where done on behalf of (and paid for) other projects and yet we don't link to their issue trackers. It's not clear to me that we should make an exception here.

It makes it look (to me) like this project is being managed by some other issue tracker that belongs to a different project.

>  Does this mean I should stop linking Pade issues to Converse and vice-versa?

What you do in Pade's changelog doesn't concern this project directly. You're free to link however you see fit. Converse is an upstream dependency of Pade and Startin'Blox. In my mind linking to an upstream dependency is different than linking to a downstream project.

>  duplicating the issues on converse is an additional administrative task that would be nice to avoid

I don't expect you to duplicate the issues here, I don't create an issue for every thing I do.

deleolajide: OK. Got it

deleolajide:
I am going to resubmit this PR

jcbrand:
> I am going to resubmit this PR

It's often better to just continue working on the same branch so that the context is not lost.

If you make new commits to this branch, this PR will automatically be updated, so you don't need to close and resubmit.

PR #1749, opened 2019-10-09
prevent message correction for files

ChaosKid42:
I believe Converse should not allow editing of sent file upload messages, as this would only make sense if you actually upload another file. This PR prevents editing of sent file uploads.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 5.0.5
+
+- Prevent editing of sent file uploads.
+
 ## 5.0.4 (2019-10-08)
 
 - New config option [allow_message_corrections](https://conversejs.org/docs/html/configuration.html#allow_message_corrections)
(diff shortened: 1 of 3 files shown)


jcbrand:
> as this would only make sense if you actually upload another file

I'm not sure I agree. An uploaded file is just a link, so it's just text.

ChaosKid42:
> > as this would only make sense if you actually upload another file
> 
> I'm not sure I agree. An uploaded file is just a link, so it's just text.
Is it really? I believe the changed link would need to be included in the `<oob>` tag which I think is not the case right now. Additionally I believe editing an upload would have to offer a reupload (of the new file version) instead of giving the user the opportunity to editing the link directly.

Anyway: Would you be willing to accept this PR if I introduced anew config option `message_correction_for_files` which defaults to `true`?

jcbrand:
I'm mostly concerned that users can then not remove a sent image, but we can use message retractions for that, I'm going to add support for it soon.

Issue #1735, opened 2019-09-26
Chatboxes stay empty after logging out and in again


ChaosKid42:
The MUC chatboxes open but stay empty if you log in immediately after log off.

1. Go to converse.js
2. Trusted device off and log in
3. Open MUC chatbox (messages are shown)
4. Log off
5. Log in (no page reload in between)
6. Open MUC (messages are not shown)

**Expected behavior**
Messages are shown

**Environment (please complete the following information):**
 - Desktop
 - Browser Chrome
 - Converse.js version 5.0.3

Ejabberd 19.08 self-hosted

licaon-kter:
Seeing the same too sometimes, after logoff you need to CTRL-F5 force reload. 

As with other issues reported by me, unless you cleanly logout and force reload (some sort of the cleanest state possible) stuff will randomly be broken :(

ChaosKid42:
For now I'm simply using the following ugly hack:
```diff
diff --git a/src/headless/converse-core.js b/src/headless/converse-core.js
index d27a9c9f8..eaaa3ff0f 100644
--- a/src/headless/converse-core.js
+++ b/src/headless/converse-core.js
@@ -548,6 +548,7 @@ function clearSession  () {
      * @event _converse#clearSession
      */
     _converse.api.trigger('clearSession');
+    location.reload();
 }
```
But maybe this should be solved correctly by someone more educated than me :-)

MotoGP-Fan:
Maybe this helps to locate the issue

Tested with:
Prosody 0.11.2
Firefox 69.0 + Chrome 77.0.3865.75

Converse 5.0.3 on local installation
Converse 5.0.3 over external "cdn" setup
(both with minified js+css)

converse.initialize({
debug: true,
authentication: 'login',
bosh_service_url: 'https://example.de/http-bind',
locked_domain: 'example.de',
allow_registration: false,
show_send_button: true,
roster_groups: true,
archived_messages_page_size: 25,
message_carbons: true,
theme: 'default',
view_mode: 'fullscreen'
});

**Problem:**
As soon as you add a MUC to your bookmarks and quit Conversejs, this message appears - see pic
If you login again immediately after logout no chat messages will be displayed.
To display received messages you need to press Ctrl+F5 when you are logged in.
It is also possible to press Ctrl+F5 after log out.

**If you delete all bookmarked MUCs, everything seems to work as usual (no "error message appears after logout).**

If you need more information pls. let me know.
[...]

PR #1750, opened 2019-10-09
add config option 'reload_on_disconnect'

ChaosKid42:
This is supposed to provide a solution for https://github.com/conversejs/converse.js/issues/1735.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## 5.0.5 (unreleased)
+
+- New config option [reload_on_disconnect](https://conversejs.org/docs/html/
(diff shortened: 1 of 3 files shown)


jcbrand:
As you're well aware, this is a hack :)

I think we should rather try to fix the underlying problem (not that I currently know what it is).

ChaosKid42:
Yes. It's a hack. So from my point of view it would be ok if you close this PR. It can still be referenced in case any user cares to use this hack until the real solution is available.

Do you have any idea what the underlying problem could be? I would like to investigate this further but I can't due to lack of ideas where to look :-)

jcbrand:
I've commented on the ticket.

jcbrand:
#1735 has now been fixed.

PR #1788, opened 2019-11-21
Added examples for bosh & websocket connection to documentaton

poVoq:
also  added recommendation to enable smacks when websockets are used.


jcbrand:
Thanks!

PR #1799, opened 2019-11-29
Handle notifications about affiliation change for users not in a room

deleolajide:
Add support for [MUC notifications about affiliation change for users not in a room](https://xmpp.org/extensions/xep-0045.html#example-196)
See https://git.happy-dev.fr/startinblox/components/sib-chat/issues/96


src/headless/converse-muc.js
```diff
@@ -507,6 +507,40 @@ converse.plugins.add('converse-muc', {
                     }, null, 'message', 'groupchat', null, room_jid,
                     {'matchBareFromJid': true}
                 );
+                this.muc_notifications_handler = _converse.connection.addHandler(stanza => {
+                    const item = converse_muc_sizzle("x[xmlns=\"".concat(converse_muc_Strophe.NS.MUC_USER, "\"] item"), stanza).pop();
```
jcbrand: What is `converse_muc_sizzle`? I don't see it being defined anywhere.

Looks like you copied code that was generated by a  bundler or something?

Also, this selector string is difficult to parse:
`x[xmlns=\"".concat(converse_muc_Strophe.NS.MUC_USER, "\"] item"`

Rather use a [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).

For example: 
    `x[xmlns="${Strophe.NS.MUC_USER}"], item`

src/headless/converse-muc.js
```diff
@@ -507,6 +507,40 @@ converse.plugins.add('converse-muc', {
...
+                          'nick': converse_muc_Strophe.getNodeFromJid(jid),
+                          'type': type,
+                          'states': [],
+                          'show': type == 'unavailable' ? 'offline' : 'online',
+                          'affiliation': affiliation,
+                          'role': item.getAttribute('role'),
+                          'jid': jid
+                        }
+
+                        console.debug("muc_handler - muc occupant push", data);
```
jcbrand: Don't use the `console` directly, if you want to log information, then import and use the `log` utility. There are lots of examples in the code of how this is done.

src/headless/converse-muc.js
```diff
@@ -507,6 +507,40 @@ converse.plugins.add('converse-muc', {
...
+
+                        const attributes = Object.assign(data, {'jid': converse_muc_Strophe.getBareJidFromJid(data.jid),'resource': converse_muc_Strophe.getResourceFromJid(data.jid)});
```
jcbrand: Please wrap lines before they become 120 chars.

src/headless/converse-muc.js
deleolajide: >Looks like you copied code that was generated by a bundler or something?

Sorry, I pasted the wrong code with my debugging console.log from converse.js. I shall redo

src/headless/converse-muc.js
```diff
@@ -507,6 +507,42 @@ converse.plugins.add('converse-muc', {
...
+                          'states': [],
+                          'show': type == 'unavailable' ? 'offline' : 'online',
+                          'affiliation': affiliation,
+                          'role': item.getAttribute('role'),
+                          'jid': jid
+                        }
+
+                        const attributes = Object.assign(data, {
+                            'jid': converse_muc_Strophe.getBareJidFromJid(data.jid),
+                            'resource': converse_muc_Strophe.getResourceFromJid(data.jid)
```
jcbrand: `converse_muc_Strophe` this is also wrong.

src/headless/converse-muc.js
```diff
@@ -507,6 +507,42 @@ converse.plugins.add('converse-muc', {
...
+
+                    if (item) {
+                        const from = stanza.getAttribute("from");
+                        const type = stanza.getAttribute("type");
+                        const affiliation = item.getAttribute('affiliation');
+                        const jid = item.getAttribute('jid');
+
+                        const data = {
+                          'from': from,
+                          'nick': converse_muc_Strophe.getNodeFromJid(jid),
```
jcbrand: This is also wrong.

src/headless/converse-muc.js
```diff
@@ -507,6 +507,42 @@ converse.plugins.add('converse-muc', {
...
+                        }
+
+                        const attributes = Object.assign(data, {
+                            'jid': Strophe.getBareJidFromJid(data.jid),
+                            'resource': Strophe.getResourceFromJid(data.jid)
+                        });
+                        const occupant = this.occupants.findOccupant({'jid': data.jid});
+
+                        if (occupant) {
+                          occupant.save(attributes);
```
jcbrand: Looks like `attributes` must actually be `data`.

I think you should write a test for this change. That would have caught all the problems identified so far and it will also make sure that this feature doesn't accidentally get removed or broken in the future.
Here's a test that checks for affiliation changes via `presence` stanzas:

https://github.com/conversejs/converse.js/blob/f1e1e538b63eb72f755975d27912d99cf16670b7/spec/muc.js#L2841

src/headless/converse-muc.js
```diff
@@ -507,6 +507,42 @@ converse.plugins.add('converse-muc', {
...
+                        } else {
+                          this.occupants.create(attributes);
+                        }
+                    }
+                    return true;
+
+                }, null, 'message', null, null, room_jid,
+                {'matchBareFromJid': true });
```
jcbrand: `matchBareFromJid` doesn't make sense in this case.

src/headless/converse-muc.js
```diff
@@ -507,6 +507,42 @@ converse.plugins.add('converse-muc', {
...
+
+                        if (occupant) {
+                          occupant.save(attributes);
+                        } else {
+                          this.occupants.create(attributes);
+                        }
+                    }
+                    return true;
+
+                }, null, 'message', null, null, room_jid,
```
jcbrand: You can specify the xmlns here (replace the first `null` with `Strophe.NS.MUC_USER`, then you don't have to check whether the item exists on line 513 and then this handler will only fire when a child of the stanza has that xmlns.

deleolajide: We might want to handle other &lt;message&gt; child stanzas with different namespaces here in future. Don't you think?

src/headless/converse-muc.js
deleolajide: ok

src/headless/converse-muc.js
deleolajide: ok

src/headless/converse-muc.js
jcbrand: No, probably not, and even if that were the case, then we can simply loosen the restriction again.

The way it currently is, this handler gets called unnecessarily probably 99% of the time.

deleolajide: I still need to extract the child element from the stanza, so the test is still required

deleolajide: > we can simply loosen the restriction again.

ok

jcbrand:
Hi @deleolajide, the test you wrote is failing on Jenkins. Did you run it locally?

deleolajide:
> Hi @deleolajide, the test you wrote is failing on Jenkins. Did you run it locally?

No. Sorry, I have not configured my windows dev environment to run the converse tests. I might need to configure a Docker instance for this.

deleolajide:
Working on a fix...

jcbrand:
Thanks @deleolajide!

I've manually merged this PR and made a few small, cosmetic changes in a followup commit.

deleolajide:
>I've manually merged this PR and made a few small, cosmetic changes in a followup commit.

Thank you @jcbrand That is very much appreciated

PR #1802, opened 2019-11-29
add UT and electron app to website

poVoq:
HTML untested, but should work ;)


jcbrand:
Thanks, merged manually.

PR #1818, opened 2019-12-26
Update converse-mam.js - add `groupchat` option in documantation

yaskhan:
I think why do not come archives from group chats, and it turns out that you need to add "groupchat" in options. I spent half an hour searching in the documentation, and half an hour searching in the convers tests. And finally found in tests.


jcbrand:
Thanks!

PR #1819, opened 2019-12-29
notification is now clickable

ChaosKid42:
This PR makes desktop notifications clickable: A click on a desktop notification now opens the corresponding chat.

### DIFF
--- a/src/converse-notification.js
+++ b/src/converse-notification.js
@@ -183,6 +183,13 @@ converse.plugins.add('converse-notification', {
             if (_converse.notification_delay) {
                 setTimeout(n.close.bind(n), _converse.notification_delay);
             }
+            n.onclick = function (event) {
+                event.preventDefault();
+                window.focus();
+                const chat = _converse.chatboxes.get(from_jid);
+                chat.maybeShow(true);
+            }
+            n.onclick.bind(_converse);
         };
 
         _converse.showChatStateNotification = function (contact) {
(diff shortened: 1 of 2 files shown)


src/converse-notification.js
```diff
@@ -183,6 +183,12 @@ converse.plugins.add('converse-notification', {
             if (_converse.notification_delay) {
                 setTimeout(n.close.bind(n), _converse.notification_delay);
             }
+            n.onclick = function (event) {
+                event.preventDefault();
+                window.focus();
+                _converse.api.chats.open(from_jid);
```
jcbrand: This only handles one-on-one chats, not MUCs, and you also need to pass in `true` for the `force` parameter, otherwise minimized or obscured chats aren't brought to the foreground.

Instead of using the API, you can try:
```
const chat = _converse.chatboxes.get(from_jid);
chat.maybeShow(true);
```

jcbrand:
This is a very nice enhancement!

src/converse-notification.js
jcbrand: Just to clarify, I think `_converse.api.chats.open` will work for MUCs, but `open` will also open a new one-on-one chat if it doesn't exist.

That won't happen in this case, because the chat will (probably) always exist, but semantically it doesn't fit.

ChaosKid42: Thanks for the review. Updated the PR accordingly.

Issue #1791, opened 2019-11-22
auto_focus set to false only works on initial connection to MUC


poVoq:
**Describe the bug**
I am testing `auto_focus` set to false in view-mode `mobile` to prevent the on-screen keyboard popping up, and it seems to do the trick when I first connect to a MUC. However on subsequent switching to already opened MUCs it seem to fail.

**To Reproduce**
1. open mobile view mode in a mobile browser
2. open a MUC from a bookmark and see it work (no OSK pop up)
3. navigate back to the controlbox
4. click on previously opened MUC to re-enter and see the OSK pop up immediately

**Expected behavior**
OSK should really only be activate when the user clicks the chat input field

**Environment (please complete the following information):**
 - Desktop or Mobile? Mobile
 - Browser: Android Firefox and Ubuntu Touch Morph (chromium based)
 - Converse.js version: 5.0.5

poVoq:
Any chance this could be fixed for 6.0?

Probably the most annoying bug on mobile right now :(

Thanks!

jcbrand:
Why don't you try and see whether you can contribute a fix yourself?

poVoq:
Well... for this one I wouldn't even know where to look. But yeah :(

jcbrand:
Search through the code (for example with `grep`) for `auto_focus` and see where it's used. Most likely the necessary changes need to be made in converse-chatview.js or converse-muc-views.js.

poVoq:
Thanks a lot!

PR #1820, opened 2019-12-29
focus now on jid field after controlbox is loaded

ChaosKid42:
This PR assures that the inital focus is on the jid field after the login-panel in the controlbox is rendered.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 - #1313: Stylistic improvements to the send button
 - #1793: Send button doesn't appear in Firefox in 1:1 chats
 - #1822: Don't log error if user has no bookmarks
+- #1820: Set focus on jid field after controlbox is loaded
 
 ## 6.0.0 (2020-01-09)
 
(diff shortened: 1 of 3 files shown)


jcbrand:
Mozilla says it's bad for accessibility and mobile usability to autofocus.

See here: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#htmlattrdefautofocus

@poVoq also complains about autofocus in #1791. In the least you should first check the `auto_focus` configuration setting.

poVoq:
Yes, on mobile devices this should result in the immediate popup of the onscreen keyboard which is usually undesirable.

ChaosKid42:
Updated the PR accordingly.

CHANGES.md
```diff
@@ -36,6 +36,7 @@
 - #1796: Don't show "back" arrow navigation (on mobile) in the chat header when in `singleton` mode
 - #1821: Errors caused by malformed URLs are now handled
 - #1819: Click on a desktop notification now opens corresponding chat.
+- #1820: Set focus on jid field after controlbox is loaded
```
jcbrand: Please rebase and move this line to the top of the changelog.

Otherwise, looks good.

ChaosKid42: Done

jcbrand:
Thanks

PR #1821, opened 2019-12-29
Errors caused by malformed URLs are now handled

ChaosKid42:
Entering malformed URLs such as `www.%` or `http://a:%p@c` causes exceptions which leads to that these messages are not shown at all.

This PR introduces some exception handling which corrects this.


jcbrand:
Thanks @ChaosKid42 

I merged manually and then did some refactoring because I wanted to restrict try/catch to be only around lines of code that we can reasonably expect to fail for malformed URLs.

Generally you don't want to put a try/catch around too much code, because then other unanticipated errors can also get caught which can make debugging more difficult.

I've also added logging so that it's possible to see when an exception was raised.

I got a bit carried away with the refactoring, hopefully it's still working properly. This is why it's very helpful to always write a test with a PR such as this one, so that one can be sure that one doesn't break the fix refactoring.

PR #1822, opened 2019-12-30
don't log error if user has no bookmarks

ChaosKid42:
I think converse should not log an error in case the user simply has no bookmarks.

### DIFF
--- a/src/headless/converse-bookmarks.js
+++ b/src/headless/converse-bookmarks.js
@@ -237,18 +237,19 @@ converse.plugins.add('converse-bookmarks', {
                         [__("The server did not return your bookmarks within the allowed time. "+
                             "You can reload the page to request them again.")]
                     );
-                } else {
-                    log.error('Error while fetching bookmarks');
-                    log.error(iq);
-                }
-                if (deferred) {
+                } else if (deferred) {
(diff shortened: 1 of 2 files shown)


src/headless/converse-bookmarks.js
```diff
@@ -237,18 +237,17 @@ converse.plugins.add('converse-bookmarks', {
...
-                    log.error(iq);
-                }
-                if (deferred) {
+                } else if (deferred) {
                     if (iq.querySelector('error[type="cancel"] item-not-found')) {
                         // Not an exception, the user simply doesn't have any bookmarks.
                         window.sessionStorage.setItem(this.fetched_flag, true);
                         return deferred.resolve();
                     } else {
                         return deferred.reject(new Error("Could not fetch bookmarks"));
```
jcbrand: Here you still need to log an error though, otherwise we won't know what really went wrong.

ChaosKid42: Updated accordingly.

jcbrand: Thanks, sorry but the changelog entry needs to be moved to the top as well.

ChaosKid42: Done

PR #1823, opened 2019-12-30
config option for roomid validation

ChaosKid42:
By this PR I'd like to introduce an option that defines a regular expression that a room id must satisfy to allow room creation. Server administrators may want to impose restrictions on the minimum and maximum length and the allowed character set allowed for room ids. Otherwise users might create rooms which are difficult to handle.

However, restricting that on the server only leads to bad UX as users might learn of the servers policy only after they have tried to create a room. Furthermore error messages from XMPP-servers might not be self-explanatory.

Therefore this option allows converse to already check the policy and disallow the user from even trying to entering/creating such a room.

### DIFF
--- a/locale/converse.pot
+++ b/locale/converse.pot
@@ -181,6 +181,9 @@ msgstr ""
 msgid "Groupchat address (JID)"
 msgstr ""
 
+msgid "Groupchat id is invalid."
+msgstr ""
+
 #: dist/converse-no-dependencies.js:8955
 msgid "Description"
 msgstr ""
--- a/sass/_modal.scss
+++ b/sass/_modal.scss
@@ -22,6 +22,11 @@
             .modal-footer {
                 justify-content: flex-start;
             }
+            .roomid-policy-error {
+                color: var(--error-color);
+                font-size: var(--font-size-small);
+                float: right;
+            }
         }
 
         .scrollable-container {
--- a/src/converse-muc-views.js
+++ b/src/converse-muc-views.js
@@ -110,6 +110,8 @@ converse.plugins.add('converse-muc-views', {
             'muc_mention_autocomplete_filter': 'contains',
             'muc_mention_autocomplete_show_avatar': true,
             'roomconfig_whitelist': [],
+            'muc_roomid_policy': null,
+            'muc_roomid_policy_hint': null,
             'visible_toolbar_buttons': {
                 'toggle_occupants': true
             }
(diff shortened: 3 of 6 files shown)


src/converse-muc-views.js
```diff
@@ -542,11 +547,17 @@ converse.plugins.add('converse-muc-views', {
...
+                    'muc_roomid_policy_error_msg': this.muc_roomid_policy_error_msg,
+                    'muc_roomid_policy_hint': _converse.muc_roomid_policy_hint
                 }));
             },
 
             afterRender () {
+                if (_converse.muc_roomid_policy_hint) {
+                    this.el.querySelector('.roomid-policy-hint').innerHTML
+                        = _converse.muc_roomid_policy_hint;
+                }
```
jcbrand: It's a bad idea to allow arbitrary HTML to be passed in to `converse.initialize` and then to render that since it creates a potential opening for malicious code to be injected.

I think the hint should just be text or alternatively you should use the `xss` library to filter out harmful markup and scripts (for example see converse-message-view.js).

Also, you should rather render the hint declaratively in the template, than using `innerHTML`.
If you use only two curly brackets, then the markup is rendered, three curly brackets cause text-only rendering.

ChaosKid42: Again thanks for the review. Pushed the necessary changes.

src/templates/add_chatroom_modal.html
```diff
@@ -12,15 +12,19 @@
                 <form class="converse-form add-chatroom">
                     <div class="form-group">
                         <label for="chatroom">{{{o.label_room_address}}}:</label>
-                        <input type="text" required="required" name="chatroom" class="form-control" placeholder="{{{o.chatroom_placeholder}}}"/>
+                        <label class="roomid-policy-error">{{{o.muc_roomid_policy_error_msg}}}</label>
+                        <input type="text" required="required" name="chatroom" class="form-control roomjid-input" placeholder="{{{o.chatroom_placeholder}}}"/>
+                    </div>
+                    <div class="form-group">
+                        {{o.muc_roomid_policy_hint}}
```
jcbrand: This also.

locale/converse.pot
```diff
@@ -181,6 +181,9 @@ msgstr ""
 msgid "Groupchat address (JID)"
 msgstr ""
 
+msgid "Groupchat id is invalid!"
+msgstr ""
+
```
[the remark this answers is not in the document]
ChaosKid42: Done

sass/_modal.scss
```diff
@@ -22,6 +22,11 @@
             .modal-footer {
                 justify-content: flex-start;
             }
+            .roomid-policy-error {
+                color: red;
+                font-size: 12px;
+                float: right;
+            }
```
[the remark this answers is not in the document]
ChaosKid42: Done

PR #1826, opened 2019-12-31
A user can now add himself as a contact

ChaosKid42:
This is a very much simplified version of https://github.com/conversejs/converse.js/pull/1634.
It does not do anything about the subscription of the self contact and also does not show the correct user status (online/away etc). It is however sufficient to safe your own contact to the roster and use it for whatever you want to do with it (might e.g. be useful to send data such as images from your computer to your smartphone).
The subscription of the self contact is always shown as "both" regardless of what is actually saved in the roster. This is of course not entirely correct but might nevertheless be acceptable taking into account that this solution only needs minimum changes to the code.

### DIFF
--- a/src/headless/converse-roster.js
+++ b/src/headless/converse-roster.js
@@ -655,7 +655,6 @@ converse.plugins.add('converse-roster', {
              */
             updateContact (item) {
                 const jid = item.getAttribute('jid');
-                if (this.isSelf(jid)) { return; }
 
                 const contact = this.get(jid);
                 const subscription = item.getAttribute("subscription");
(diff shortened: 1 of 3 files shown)


src/headless/converse-roster.js
```diff
@@ -655,10 +655,9 @@ converse.plugins.add('converse-roster', {
              */
             updateContact (item) {
                 const jid = item.getAttribute('jid');
-                if (this.isSelf(jid)) { return; }
 
                 const contact = this.get(jid);
-                const subscription = item.getAttribute("subscription");
+                const subscription = this.isSelf(jid) ? 'both' : item.getAttribute("subscription");
```
jcbrand: This feels too hacky. What happens when you just remove line 658, without making this change?

ChaosKid42: Because of `subscription === null` the roster item won't be shown.
I updated the PR to make it less hacky: Now the case of a roster item belonging to yourself is handled in rosterview.

jcbrand: Looks better thanks.

jcbrand:
@ChaosKid42 Please resolve the conflicts, then I'll merge.

ChaosKid42:
@jcbrand Resolved the conflicts.

Issue #1274, opened 2018-10-27
Server Broadcasts will not be shown


mightymop:
Server Broadcasts will not be shown.

What i have done:

I've extended the openfire broadcastplugin and tries to send server broadcasts to users or muc.
I tested it with Conversations. 

Message Packets:

>Users:
`DEBUG: <body xmlns="http://jabber.org/protocol/httpbind" ack="3232148937"><message from="mopsdom.de" xmlns="jabber:client"><subject>(broadcast)</subject><body>test</body></message></body>`

>MUC Room:
`DEBUG: <body xmlns="http://jabber.org/protocol/httpbind" ack="3232148936"><message type="groupchat" from="ccc@conference.mopsdom.de" to="mops@mopsdom.de/converse.js-88192652" xmlns="jabber:client"><body>test</body></message></body>
`

In Conversations Messenger: 
If the broadcast was send to user jids, a user with the xmpp domain as jid will be shown in the roster with the broadcast as a message.
Even in a MUC the broadcast message will be shown as send by the MUC room.

ConverseJS: 
I can see that the message arrives on conversejs (debug mode). But thats all?!

Furthermore a broadcast to a user jid will notify me via desktopnotification but then its gone.

jcbrand:
Hi @mightymop, do you know in which XEP these are defined?

I would expect your first message to be of type `headline`, then it would work in Converse (in "overlayed" view mode).

mightymop:
Type Headline also has no effect on conversejs.
On Conversations the message became a private one in MUC.

`DEBUG: <body xmlns="http://jabber.org/protocol/httpbind" ack="3705315253"><message id="372c3cc4-b93d-45cb-9d7e-2720add1afe8" from="ccc@conference.mopsdom.de" type="headline" to="mops@mopsdom.de/converse.js-30275537" xmlns="jabber:client"><body>test/body></message></body>`

I send standard messages ... with type "chat" or "groupchat" and a childelement "body" with the message as text.
The XMPP packet looks like this:

<message id="UNIQUE ID" from="channel@conference.xmppdomain or xmppdomain" type="groupchat or chat" to="user@xmppdomain/ressource" xmlns="jabber:client"><body>messagetext</body></message>

jcbrand:
> Type Headline also has no effect on conversejs.

Headline messages appear in overlayed view mode, but not in fullscreen view mode.
https://conversejs.org/docs/html/configuration.html#view-mode

mightymop:
[...]

PR #1834, opened 2020-01-14
Add Broadcast-Messages (Announcements/MOTD)

ChaosKid42:
I picked up the work of worlword from https://github.com/conversejs/converse.js/pull/1639 because I believe it's a nice new feature which is quite near to being merged.

Therfore I cleaned the code a little bit and fixed the problem regarding backbone.

@jcbrand I would be happy if you could review this PR.

### DIFF
--- a/src/converse-service-administration.js
+++ b/src/converse-service-administration.js
@@ -1,0 +1,392 @@
+// Converse.js (A browser based XMPP chat client)
+// https://conversejs.org
+//
+// Copyright (c) 2012-2017, Jan-Carel Brand <jc@opkode.com>
+// Licensed under the Mozilla Public License (MPLv2)
+//
+
+import "converse-chatview";
+import "converse-profile";
+import "converse-rosterview";
+import converse from "@converse/headless/converse-core";
+import sizzle from "sizzle";
+import tpl_form_checkbox from "templates/form_checkbox.html";
+import tpl_form_input from "templates/form_input.html";
+import tpl_form_textarea from "templates/form_textarea.html";
+import tpl_service_admin_chat from "templates/service_administration_chat.html";
+import tpl_service_admin_option from "templates/service_administration_option.html";
+import tpl_service_admin_selector from "templates/service_administration_selector.html";
+
+const { $iq, Backbone, Strophe } = converse.env;
+const u = converse.env.utils;
+
+converse.plugins.add('converse-service-administration', {
+
+    /* Plugin dependencies are other plugins which might be
+     * overridden or relied upon, and therefore need to be loaded before
+     * this plugin.
+     *
+     * If the setting "strict_plugin_dependencies" is set to true,
+     * an error will be raised if the plugin is not found. By default it's
+     * false, which means these plugins are only loaded opportunistically.
+     *
+     * NB: These plugins need to have already been loaded via require.js.
+     */
+    dependencies: ["converse-chatboxes", "converse-rosterview", "converse-chatview"],
+
+    initialize () {
+        const { _converse } = this,
+            { __ } = _converse;
+
+        _converse.SERVICE_ADMIN_TYPE = 'service-administration';
+
+        _converse.ServiceAdministration = Backbone.Model.extend({
+            async discoverSupport () {
+                const supported = await _converse.api.disco.supports(Strophe.NS.COMMANDS, _converse.bare_jid);
+                if (supported) {
+                    const stanza = $iq({
+                        'from': _converse.connection.jid,
+                        'id': u.getUniqueId(),
+                        'to': _converse.domain,
+                        'type': 'get'
+                    }).c('query', {
+                        xmlns: Strophe.NS.DISCO_ITEMS,
+                        node: 'announce'
+                    }).up();
+
+                    _converse.connection.sendIQ(stanza, result => {
+                        const commands = sizzle('item[node^="http://jabber.org/protocol/admin"]', result);
+                        _converse.serviceAdminCommands = new _converse.ServiceAdminCommandCollection();
+                        commands.forEach(command => {
+                            const command_description = command.getAttribute('name');
+                            const command_node = command.getAttribute('node');
+                            const command_name = command_node.split('#').pop();
+
+                            _converse.serviceAdminCommands.add(new _converse.ServiceAdminCommandItem({
+                                command_node: command_node,
+                                description: command_description,
+                                command_name: command_name
+                            }));
+                        });
+                        this.renderControlboxElement();
+                    });
+                }
+            },
+
+            renderControlboxElement () {
+                _converse.serviceAdministrationCommandView = new _converse.ServiceAdminCommandView({collection: _converse.serviceAdminCommands});
+                const groupchat_element = _converse.chatboxviews.get('controlbox').el.querySelector('#chatrooms');
+                if (groupchat_element !== null) {
+                    groupchat_element.insertAdjacentElement('beforebegin', _converse.serviceAdministrationCommandView.render());
+                }
+            }
+        });
+
+        _converse.ServiceAdminCommandItem = Backbone.Model;
+
+        _converse.ServiceAdminCommandItemView = Backbone.NativeView.extend({
+            tagName: 'div',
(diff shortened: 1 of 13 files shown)


jcbrand:
Thanks @ChaosKid42, however this PR needs some tests before it can be merged. You can probably look at the headline tests for inspiration.

ChaosKid42:
@jcbrand Ok. Will add some tests.

ChaosKid42:
@jcbrand Finished implementing tests. Would be great if you could review. Thanks!

src/converse-service-administration.js
```diff
@@ -0,0 +1,412 @@
...
+            renderControlboxElement () {
+                _converse.serviceAdministrationCommandView = new _converse.ServiceAdminCommandView({collection: _converse.serviceAdminCommands});
+                const groupchat_element = _converse.chatboxviews.get('controlbox').el.querySelector('#chatrooms');
+                if (groupchat_element !== null) {
+                    groupchat_element.insertAdjacentElement('beforebegin', _converse.serviceAdministrationCommandView.render());
+                }
+            }
+        });
+
+        _converse.ServiceAdminCommandItem = Backbone.Model.extend({});
```
jcbrand: You don't have to call `.extend({})`, simply setting it to `Backbone.Model` is enough.

src/converse-service-administration.js
```diff
@@ -0,0 +1,412 @@
...
+                    _converse.log(error, Strophe.LogLevel.ERROR);
+                });
+            },
+
+            getAttributesFromField (field) {
+                const attribute_names = field.getAttributeNames();
+                const obj = {};
+                attribute_names.forEach(name => {
+                    obj[name] = field.getAttribute(name);
+                });
```
jcbrand: Not really important, but I like using `reduce` instead of `forEach` to create an object from an array.
You can usually save a few lines of code that way.

For example, you can replace the above 3 lines with one line:
```
const obj = attribute_names.reduce((o, n) => (o[n] = field.getAttribute(n)), {});
```

src/converse-service-administration.js
```diff
@@ -0,0 +1,412 @@
...
+                obj["value"] = this.getValueOfFieldElement(field);
+                obj["required"] = field.getElementsByTagName('required').length !== 0;
+                if (obj.var === "body" && obj.value !== null && obj.value.length === 0) {
+                    obj.value = __("-- sent from %1$s",
+                            Strophe.getBareJidFromJid(_converse.jid));
```
jcbrand: You can just use `_converse.bare_jid`

src/converse-service-administration.js
```diff
@@ -0,0 +1,412 @@
...
+
+            getValueOfFieldElement (field) {
+                const values = field.getElementsByTagName('value');
+                if (values.length > 0) {
+                    return values[0].innerHTML;
+                } else {
+                    return '';
+                }
+            },
+        });
```
jcbrand: This looks like a utility method which you can put in src/headless/utils/form.js and then call in this plugin via `u.getValueOfFieldElement`.

Make sure to import form.js in this plugin if you don't already.

src/converse-service-administration.js
```diff
@@ -0,0 +1,412 @@
...
+
+                this.render();
+            },
+
+            render () {
+                this.el.innerHTML = tpl_service_admin_chat({
+                    'service_admin_header': this.model.attributes.header_text,
+                    'info_close': __('Close'),
+                    'send_button_text': this.model.attributes.send_button_text,
+                    'placeholder_text': this.model.attributes.placeholder_text,
```
jcbrand: You can use `this.model.getJSON()`, then you don't have to type fields out that re-use the same name.

For example:
```
this.el.innerHTML = tpl_service_admin_chat(Object.assign(this.model.toJSON(), {
    'service_admin_header': this.model.get('header_text'),
    'info_close': __('Close'),
});
```

jcbrand:
@ChaosKid42: I enabled [mod_admin_adhoc](https://prosody.im/doc/modules/mod_admin_adhoc) on my local Prosody instance and then tried to test this plugin locally.

It looks like this plugin is a weird combination of trying to add generic support for ad-hoc commands and service administration and specific support for announcements and IMO it goes about it in the wrong way.

We should instead add generic support for ad-hoc commands, and then if the server supports announcements, the user can choose that from the list of supported commands and make the announcement.

Also, I don't think the UI for executing commands should be in a chat box. Instead, I think it should go into a modal, and the first step should be specifying the JID of the entity for which you want to fetch the available commands.

So, the UX would be as follows:

* Open the service admin modal (not yet sure where the button/link for this should be).
* The modal contains an input field where you need to enter the JID of the entity for which you want to query for a list of commands (perhaps it can be pre-filled with the domain of the user's JID).
* The user submits and a list of supported commands appears in the modal.
* The user chooses to execute a command and the relevant form for that command appears  in the modal.
* The user executes the command and a success/error response appears.

As far as I can tell, this workflow is much different than what this plugin tries to do currently and therefore I don't think this plugin will ever be merged.

jcbrand:
Thinking about this further... I do think that it would be nice to create an easy way for someone to write announcements without having to first search for available commands in a modal.

We already have support for `headline` messages with `converse-headlines` and AFAIK the announcements will appear in a headlines box (is that correct?).

So I think that would be the logical place to make it possible to write announcements. So I think what we can do is to add support for showing headlines in the controlbox (see #1274) and then if the user's server allows the user to make announcements, we can automatically list a headlines-box for that server.

This would be similar to the approach in this PR, but it would be using the correct and relevant chatbox for sending (and receiving) announcements.

Then besides that, we could also work on providing general support for executing ad-hoc commands via a modal.

src/converse-service-administration.js
ChaosKid42: I belive this won't work because `(o[n] = field.getAttribute(n))` returns the assigned value and not the complete object.

ChaosKid42:
> We already have support for `headline` messages with `converse-headlines` and AFAIK the announcements will appear in a headlines box (is that correct?).

Yes.

> So I think that would be the logical place to make it possible to write announcements. So I think what we can do is to add support for showing headlines in the controlbox (see #1274) and then if the user's server allows the user to make announcements, we can automatically list a headlines-box for that server.

I like your idea. I'll try to find the time to work on it on the weekend.

ChaosKid42:
@jcbrand The folllowing screenshot shows the current state of this PR. In case the user has the appropriate rights, a collapsable list of all discovered commands is inserted in the headline-panel in the controlbox, below any headline messages. Initially the list is in collapsed-state.

Sending of broadcast-messages and reception of these is happening in different chatboxes, which for me makes sense, because for sending you need to visualize the forms sent from the XMPP-server.

converse_broadcast

What do you think? How can we move forward?

jcbrand:
I think that "SERVER MESSAGES' should rather be renamed to "ANNOUNCEMENTS".

The controlbox already gets quite full, so I don't think we should use it to put commands in there. I therefore think the "Broadcast-Messages" list should go into a modal and you can open the modal by clicking a button (e.g. "cog" icon) next to the ANNOUNCEMENTS heading, similarly to how modals can be opened for CONTACTS and GROUPCHATS.

The same applies to the Service Admin box. I think that functionality should be inside the same modal that shows the broadcast messages commands. Some deployments have singleton set to `true` but they might still want to allow ad hoc commands, by moving that stuff into a modal those deployments can still use ad hoc commands.

I've currently making lots of changes, especially to modals, by replacing all the lodash templates with [lit-html](https://lit-html.polymer-project.org/). I've only replaced about half of all the templates and already the tests are running way faster. lit-html is really fast.

I want to merge these changes soonish, so please hold off on this PR for a while until those changes are merged in.

ChaosKid42:
Ok. I'll wait until you finished the switch to lit-html.

jcbrand:
@ChaosKid42: I've now added general support for XEP-0050, which AFAIK means that you can also send out announcements.

You can use ad-hoc commands by clicking on the gear icon next to your JID/name at the top left in fullscreen mode.

A modal will appear and there will be a tab called `Commands` which you can click.

ChaosKid42:
@jcbrand Agreed. Your solution works like a charm. Thanks a lot.

PR #1839, opened 2020-01-22
show headline messages in controlbox

ChaosKid42:
This PR is supposed to solve https://github.com/conversejs/converse.js/issues/1274


jcbrand:
Thanks @ChaosKid42, looks good.

Please just add one or two tests as well.

 [spec/roomslist.js](https://github.com/conversejs/converse.js/blob/master/spec/roomslist.js) has some similar tests that you can look at.

ChaosKid42:
@jcbrand Added two tests.

PR #1847, opened 2020-01-27
Fix incrementMsgCounter cleaning up after itself

laszlovl:
Previously, after increasing the msg_counter from 1 to 2, the window title would be set to "Messages (2)bla", and in the next iteration from 2 to 3 the regex wouldn't match anymore since it expects a space after the closing parentheses. This would spam up your window title with dozens of "Messages (x)" references.

Small enough that I thought it doesn't need a dedicated issue.


jcbrand:
@laszlovl Thanks!

Issue #1715, opened 2019-09-13
Converse does not parse Chat State on LMC messages


iNPUTmice:
When parsing a message correction converse does not parse the chat state from the same message.

**To Reproduce**
1. Send message
2. Send typing chat state
3. Send message that corrects message from step 1 and contains an active chat state

**Expected behavior**
Converse resets the 'Contact is typing' info upon parsing and displaying the correction.

**Environment:**
 - Firefox. Converse 5.0.1

**Additional context**
Parsing the active state from normal messages (that aren’t corrections) works fine.

PR #1893, opened 2020-02-27
Update chat state notifications when message is edited.

legastero:
Chat state notifications for a sender are cleared out once a new message is added to the collection. For corrections, however, the existing message model is updated instead of a new one added.

This adds a `change:edited` hook to the messages collection so that the chat state notifications can be updated.

Resolves issue #1715

### DIFF
--- a/src/converse-chatview.js
+++ b/src/converse-chatview.js
@@ -196,6 +196,7 @@ converse.plugins.add('converse-chatview', {
                 this.initDebounced();
 
                 this.listenTo(this.model.messages, 'add', this.onMessageAdded);
+                this.listenTo(this.model.messages, 'change:edited', this.onMessageEdited);
                 this.listenTo(this.model.messages, 'rendered', this.scrollDown);
                 this.model.messages.on('reset', () => {
                     this.content.innerHTML = '';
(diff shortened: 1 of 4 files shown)


src/converse-chatview.js
```diff
@@ -1072,6 +1082,16 @@ converse.plugins.add('converse-chatview', {
...
+            /**
+             * Remove chat state notifications for a given sender JID.
+             * @private
+             * @method _converse.ChatBoxView#clearChatStateForSender
+             * @param {string} sender - The sender of the chat state
+             */
+            clearChatStateForSender (sender) {
+                sizzle(`.chat-state-notification[data-csn="${sender}"]`, this.content).forEach(u.removeElement);
+            },
+
```
jcbrand: Can you please also call `clearChatStateForSender` in `showMessage`?

src/converse-chatview.js
```diff
@@ -196,6 +196,7 @@ converse.plugins.add('converse-chatview', {
                 this.initDebounced();
 
                 this.listenTo(this.model.messages, 'add', this.onMessageAdded);
+                this.listenTo(this.model.messages, 'change:edited', this.onMessageEdited);
```
jcbrand: This is for one-on-one chats only, so we need a similar line in the `initialize` method of `_converse.ChatRoomView`.

src/converse-chatview.js
legastero: Already handled in this commit 😁

Unless there's another `showMessage` I need to search for? I see another one in the omemo view, but it invokes the one here.

jcbrand: Sorry, I missed the relevant line.

jcbrand:
Merged manually.

Issue #1895, opened 2020-02-27
Build doesn't work on windows


Posnakidesd:
Based on Documentation provided https://conversejs.org/docs/html/index.html both Linux and Windows build should be possible.

**Describe the bug**
"webpack --config webpack.dev.js" on windows fails

**To Reproduce**
clone the repo in windows and run "npm install"

**Expected behavior**
Build should succeed similarly to Linux

**Environment (please complete the following information):**
 - Win 10
 - Converse.js version tryed both 5 and 6

**Additional context**
WARNING in ./locale/af/LC_MESSAGES/converse.po 1:0
Module parse failed: Unexpected character '#' (1:0)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
\> # Afrikaans translations for Converse.js package.
| # Copyright (C) 2013 Jan-Carel Brand
| # This file is distributed under the same license as the Converse.js package.
 @ ./locale lazy ^\.\/.*\/LC_MESSAGES\/converse\.po$ namespace object ./af/LC_MESSAGES/converse.po
 @ ./src/headless/i18n.js
 @ ./src/converse-bookmark-views.js
 @ ./src/converse.js
 @ ./src/entry.js
.....
....

ERROR in ./src/templates/error_message.html
Module not found: Error: Can't resolve 'C:projectconverse.js
ode_moduleslodashescape.js' in 'C:\project\converse.js\src\templates'
 @ ./src/templates/error_message.html 1:16-79
 @ ./src/converse-chatview.js
 @ ./src/converse.js
 @ ./src/entry.js

Posnakidesd:
It looks like the problem is in lodash-template-webpack-loader loader. This loader doesn't process the windows path separator correctly. I've replaced this loader with underscore-template-loader and it works fine now. But i haven't tested it in Linux.

jcbrand:
@Posnakidesd: Please make a pull request and I'll test it in Linux.

Posnakidesd:
@jcbrand Just want to emphasize that although the build was success I find some issue on template processing.

Posnakidesd:
I've fount the following settings working for me. 

`
...
 var webpack = require('webpack');
...
           {
                test: /\.(html|svg)$/,
                exclude: /node_modules/,
                use: [{
                    loader: "underscore-template-loader",
                    query: {
                        escape: '\\{\\{\\{(.+?)\\}\\}\\}',
[...]

PR #1903, opened 2020-03-03
1895 Build doesn't work on windows

Posnakidesd:
This should resolve issue #1895 Build doesn't work on windows


jcbrand:
Thanks for making the PR, however there are lots of test failures and this therefore doesn't look feasible.

PR #1938, opened 2020-04-01
fix #1937 - Editing a message removes the mentions highlight

### LINKED ISSUE #1937: Editing a message removes the mentions highlight
**Describe the bug**

When the cursor UP/DOWN keys are used to select and edit a previous message, the edition work fine, but it removes the highlight around and mentioned name in the edited text. However if the edit icon is used to select the message, this behavior does not occur

**Additional context**
Checking through the code, reveals that [onMessageEditButtonClicked](https://github.com/conversejs/converse.js/blob/bb9e394e93f86731061cadd36023e87ff0ae4a7a/src/converse-chatview.js#L1026) uses prefixMentions to apply mentions while [editLaterMessage](https://github.com/conversejs/converse.js/blob/bb9e394e93f86731061cadd36023e87ff0ae4a7a/src/converse-chatview.js#L1033)  and [editEarlierMessage](https://github.com/conversejs/converse.js/blob/bb9e394e93f86731061cadd36023e87ff0ae4a7a/src/converse-chatview.js#L1055) do not.

deleolajide:


jcbrand:
Thank you @deleolajide!

PR #1940, opened 2020-04-02
fix #1939 - Add smooth scroll to scrollDown

### LINKED ISSUE #1939: Add smooth scroll to scrollDown
When a chatbox is opened, the user is first presented with the chat scrolled to the top, and then suddenly brought to the very bottom in a split of a second. This effect has been described by a user as making them nauseous :-)

A simple way to remove that effect is to smoothly animate the scroll position, instead of a straight jump with the CSS scroll-behavior property.  In JavaScript, that would be adding
```
this.content.style.scrollBehavior = "smooth";
```
to the _scrollDown function

To keep the existing behavior, I propose a new boolean setting called **smooth_scroll** with default value *false*

deleolajide:


jcbrand:
As discussed in #1939, I'm closing this PR.

PR #1964, opened 2020-04-09
fix #1963 - Mentions are visually incorrect when used in message replies

### LINKED ISSUE #1963: Mentions are visually incorrect when used in message replies
**Describe the bug**
A message reply typically starts with the > character. Unfortunately, that the is the XML begin tag character. This causes filterXSS to escape it into *& lt;* This invalidates the start values of all references in the message and causes them to to be displayed incorrectly like this

image

**To Reproduce**
Type any message with one or more mentions and start it with > character

**Expected behavior**
The > character at the start of a message should be ignored and left unescaped as it cannot be part of any well-formed XML

**Solution**
Override the escapeHtml function in filterXSS like this
```
'escapeHtml': function (html) {
  if (html.startsWith('>'))
    return '>' + html.substring(1).replace(/</g, '&lt;').replace(/>/g, '&gt;');
  else
    return html.replace(/</g, '&lt;').replace(/>/g, '&gt;');
}

```

deleolajide:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,7 @@ Soon we'll deprecate the latter, so prepare now.
 - #1839: Headline messages are shown in controlbox
 - #1896: Don't send receipts for messages fetched from the archive 
 - #1937: Editing a message removes the mentions highlight
+- #1963: Mentions are visually incorrect when used in message replies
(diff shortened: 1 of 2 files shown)


jcbrand:
Closed based on discussion in #1963

PR #1966, opened 2020-04-11
Fix links

mheinzel:
There is another reference to `tests.html` in `converse-logs/README.md`, but I am missing some context there and don't know what `tests/main.js` (which doesn't seem to exist) should be. It's probably a quick fix for you.

Also, I don't think the following conditions apply for a small fix of the docs, or do they?


jcbrand:
Thanks!

PR #1968, opened 2020-04-12
fix #1963 - Mentions are visually incorrect when used in message replies

### LINKED ISSUE #1963: Mentions are visually incorrect when used in message replies

deleolajide:
Third attempt at fixing this by modifying addMentionsMarkup. I now adjust reference start and end based on the count for prior occurences of the &_lt; and &_gt; characters.

I really dislike the start/end value system used in the references instead of a simple markdown like the colon delimters used with emojis. It reminds me of fixed length records used in very old school COBOL programming.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -19,6 +19,7 @@ Soon we'll deprecate the latter, so prepare now.
 - #1839: Headline messages are shown in controlbox
 - #1896: Don't send receipts for messages fetched from the archive 
 - #1937: Editing a message removes the mentions highlight
+- #1963: Mentions are visually incorrect when used in message replies
 - Allow ignoring of bootstrap modules at build using environment variable. For xample: `export BOOTSTRAP_IGNORE_MODULES="Modal,Dropdown" && make dist`
 - Bugfix. Handle stanza that clears the MUC subject
 - New config option [modtools_disable_assign](https://conversejs.org/docs/html/configuration.html#modtools-disable-assign)
--- a/spec/muc_messages.js
+++ b/spec/muc_messages.js
@@ -900,6 +900,18 @@
                     'hello <span class="mention">z3r0</span> '+
                     '<span class="mention mention--self badge badge-info">tom</span> '+
                     '<span class="mention">mr.robot</span>, how are you?');
+
+                const quoted_msg = $msg({
+                        from: 'lounge@montague.lit/tom',
+                        id: u.getUniqueId(),
+                        to: 'romeo@montague.lit',
+                        type: 'groupchat'
+                    }).c('body').t('>how are you?\nI am fine. Thanks gibson').up()
+                        .c('reference', {'xmlns':'urn:xmpp:reference:0', 'begin':'32', 'end':'38', 'type':'mention', 'uri':'xmpp:gibson@montague.lit'}).nodeTree;
+                await view.model.queueMessage(quoted_msg);
+                const quoted_message = await u.waitUntil(() => view.el.querySelector('.chat-msg__text'));
+                expect(quoted_message.classList.length).toEqual(1);
+                expect(quoted_message.innerHTML).toBe('&gt;how are you?\nI am fine. Thanks <span class="mention">gibson</span>');
                 done();
             }));
         });
--- a/spec/muc_messages.js
+++ b/spec/muc_messages.js
@@ -900,18 +900,47 @@
                     'hello <span class="mention">z3r0</span> '+
                     '<span class="mention mention--self badge badge-info">tom</span> '+
                     '<span class="mention">mr.robot</span>, how are you?');
+                done();
+            }));
 
-                const quoted_msg = $msg({
-                        from: 'lounge@montague.lit/tom',
+            it("highlights all users mentioned via XEP-0372 references in a quoted message",
+                mock.initConverse(
+                    ['rosterGroupsFetched'], {},
+                    async function (done, _converse) {
+
+                const muc_jid = 'lounge@montague.lit';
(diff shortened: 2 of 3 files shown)


jcbrand:
I've been thinking a bit about referencing rich elements in text messages. 

XEP-394 also uses start/end indicators for rich elements, so this is a problem that needs solving generally and not just for a particular use-case.

I think this should be fixed as part of the lit-html rewrite of how messages are rendered, that the general problem of referencing positions in a message is handled and that this is done in a way so that people can add message rendering "middleware" to add new capabilities to what can be rendered in a message.

I'm laying the groundwork for using lit-html for rendering message and will hopefully be able to start working on that in the next few weeks.

Until then, I don't think it makes sense to merge this PR yet.

deleolajide:
> Until then, I don't think it makes sense to merge this PR yet.

I appreciate your efforts towards refactoring text messages for rich elements using lit-html. However, I have an immediate usability issue to address. StartinBlox users make heavy use of message replies with mentions and have made this an [**issue that has to be fixed**](https://git.startinblox.com/components/sib-chat/issues/132). They can't wait for Converse 7 for this.

I have spent a bit of time and effort trying to find a fix with mimium impact that passes all the converse tests as you requested in your first comment. 

By asking me to throw all that away, you are sending me mixed messages about how to contribute to Converse development. I can't fix this from a plugin and do not want to create or maintain a fork for StartinBlox. Please review my logic for improvment and reconsider for an immediate merge. Thank you.

jcbrand:
> StartinBlox users make heavy use of message replies with mentions and have made this an issue that has to be fixed. They can't wait for Converse 7 for this.

The reason I don't want to merge this to trunk is because I think the chance that this change will ever get released is very low.

If your users can't wait for Converse 7, then how is merging this to trunk going to solve anything for you? AFAIK your users aren't using trunk in any case, but instead use a patched version of Converse 6, right?

> I have invested a bit of time and effort trying to find a fix with mimium impact that passes all the converse tests as you requested in your first comment.

Asking that all tests pass isn't some kind of extra burden that I placed on you for a particular PR, it's pretty much the most basic requirement for *any* PR to be mergeable (unless it's clear to me that tests are failing due to some other reason not related to the PR).

While we're on the topic of tests, I've often asked other people to first write tests specific to their PR before merging and when fixing a bug it's generally expected to write a regression test to check that the bug is fixed and to ensure that it doesn't happen again.

In the interest of good engineering practice and fairness to all contributors, I'd like you to also write tests when making PRs. When someone is a new contributor who doesn't know the codebase very well, then one could perhaps let the tests requirement slide, but you're a veteran by now.

> By asking me to throw all that away, you are sending me mixed messages about how to contribute to Converse development.

I didn't ask that you throw away this PR, but I'm also not obligated to merge a PR just because it took someone a lot of effort to contribute it.

I'm sorry if you feel like you're getting mixed messages, that's not my intention. I, and others, appreciate your contributions. However, I also reasonably expect contributions to conform to a minimum level of quality (e.g. tests being green) and relevance to the roadmap (e.g. not in conflict with future planned changes).

> I can't fix this from a plugin and do not want to create or maintain a fork for StartinBlox

The 6.x.x branch (which AFAIK you're using) is already a fork and this PR will likely never be released as part of Converse 7.

So as a compromise, you can make this PR against the 6.x.x branch, add a regression test, and then I'll merge it there.

jcbrand:
> So as a compromise, you can make this PR against the 6.x.x branch, add a regression test, and then I'll merge it there.

I realise now that having this change together with a regression test in trunk will ensure that when I do eventually rewrite message rendering to use lit-html, that this bug doesn't reappear.

So if you write a regression test, then I think there is a good enough reason to have this in trunk and not just the 6.x.x branch.

So please add a regression test to this PR and I'll merge to trunk.

deleolajide:
> So please add a regression test to this PR and I'll merge to trunk.

Thank you. I have always added a test for new features, but you have a point about bug fixes that I did not consider. Adding the the mssing test that would have prevented the bug in the first place is also part of fixing the bug :-)

>I didn't ask that you throw away this PR, but I'm also not obligated to merge a PR just because it took someone a lot of effort to contribute it.

I apprecate that, but if you had pointed out in your first comment that "_don't think it makes sense to merge this PR yet_" into master, then I would have closed the PR and concentrated on a fork. 

A question for you. 

Should PRs that make changes to message rendering before you refactor with lit-html target the converse-6.xx branch or the master because of the regression tests?

jcbrand:
> Should future PRs that make changes to message rendering before you refactor with lit-html target the converse-6.xx branch or the master because of the regression tests?

The answer depends on how far trunk and 6.x.x have diverged. If they diverge too much, you'll have to make a decision about which branch you want to use.

If they haven't yet diverged much (when it comes to message rendering) then based on this discussion, if it's a bugfix with a regression test, then it should probably go into master with a cherry-pick into 6.x.x.

PR #2002, opened 2020-04-27
fix rendering of `muc_roomid_policy_hint`

ChaosKid42:
This PR fixes the rendering of `muc_roomid_policy_hint`.

### DIFF
--- a/src/templates/add_chatroom_modal.js
+++ b/src/templates/add_chatroom_modal.js
@@ -1,7 +1,8 @@
 import { html } from "lit-html";
+import { unsafeHTML } from "lit-html/directives/unsafe-html.js";
 import { __ } from '@converse/headless/i18n';
 import { modal_header_close_button } from "./buttons"
-
+import xss from "xss/dist/xss";
 
 const i18n_join = __('Join');
 const i18n_enter = __('Enter a new Groupchat');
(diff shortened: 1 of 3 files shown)


src/templates/add_chatroom_modal.js
```diff
@@ -32,7 +33,7 @@ export default (o) => html`
                         ${ (o.muc_roomid_policy_error_msg) ? html`<label class="roomid-policy-error">${o.muc_roomid_policy_error_msg}</label>` : '' }
                         <input type="text" required="required" name="chatroom" class="form-control roomjid-input" placeholder="${o.chatroom_placeholder}"/>
                     </div>
-                    ${ o.muc_roomid_policy_hint ?  html`<div class="form-group">{{o.muc_roomid_policy_hint}}</div>` : '' }
```
jcbrand, line 35: This makes me nervous :)

Please use the XSS library to filter the policy hint for malicious scripts before passing it to `unsafeHTML`.

ChaosKid42: It's already there: https://github.com/conversejs/converse.js/blob/master/src/converse-muc-views.js#L545

jcbrand: ah ok, can you please move that to the template? 

I think it makes sense to have it all in one place, then one can be sure that the HTML is being escaped.

ChaosKid42: Done

PR #2004, opened 2020-04-28
fix #1999 - Show first unread message

### LINKED ISSUE #1999: Show first unread message

When Converse displays "You have unread messages", it will nice to have a delimiter showing the beginning of unread messages like this
image

Original issues was posted on https://git.startinblox.com/components/sib-chat/issues/119

deleolajide:

**DONE**
**NOT APPLICABLE**
**Existing tests modified to additional check for first_unread message indicator**


src/headless/converse-muc.js
```diff
@@ -2361,7 +2361,14 @@ converse.plugins.add('converse-muc', {
...
                 if (u.isNewMessage(message) && this.isHidden()) {
-                    const settings = {'num_unread_general': this.get('num_unread_general') + 1};
+                    let first_unread = this.get('first_unread');
+
+                    if (this.get('num_unread_general') == 0) {
+                      if (first_unread) first_unread.set("first_unread", false);
+                      message.set("first_unread", true);
+                      first_unread = message;
+                    }
+                    const settings = {'first_unread': first_unread, 'num_unread_general': this.get('num_unread_general') + 1};
```
jcbrand, line 2371: Looks to me like you're saving a message model as the value of an attribute on a ChatBox or ChatRoom instance.

AFAIK this won't work because when the data is restored from localStorage, you won't get the original Message instance back, you'll get a map back.

Besides that, all the model's attributes gets JSON serialized and stored in localStorage, which is wasteful and problematic because localStorage is limited in size.

You'll need to use (and store) something like the `id` value of the first unread message and then use that to look up the Message instance when you need it.

deleolajide: OK. thanks for the review. I will store the message id instead as suggested.

src/converse-message-view.js
```diff
@@ -260,6 +260,7 @@ converse.plugins.add('converse-message-view', {
                         'occupant': this.model.occupant,
                         'pretty_time': time.format(api.settings.get('time_format')),
                         'retraction_text': is_retracted ? this.getRetractionText() : null,
+                        'first_unread': this.model.get('first_unread'),
```
jcbrand, line 263: This line isn't necessary. `first_unread` already gets passed in via `this.model.toJSON()` on line 251.

src/headless/converse-chat.js
```diff
@@ -1100,7 +1100,17 @@ converse.plugins.add('converse-chat', {
                     return;
                 }
                 if (utils.isNewMessage(message) && this.isHidden()) {
-                    this.save({'num_unread': this.get('num_unread') + 1});
+                    let first_unread = this.get('first_unread');
+
+                    if (this.get('num_unread') == 0) {
+                      if (first_unread) {
+                          const msg = this.messages.where({'msgid': first_unread});
```
jcbrand, line 1107: `msgid` isn't guaranteed to be unique. This is the `id` attribute that the client puts on the stanza, and older clients in particular don't maintain globally unique message ids.

You should instead try to use the XEP-359 stanza id if it's available, otherwise try the `origin_id` and lastly fall back to the `msgid`.

The stanza ID needs to be checked together with the `by` value, something similar to what's being done here:
https://github.com/conversejs/converse.js/blob/29638309168551d609a797e5e17e8e214d695955/src/headless/converse-mam.js#L41

If the stanza id isn't available, then you can probably fall back to to the `msgid`.

---

You can use `messages.findWhere` instead of `where` to stop once the first match has been found and you can use the new optional chaining syntax to shorten everything into one line.

`this.messages.findWhere(query)?.save("first_unread", false);`

Also, is there a particular reason why you're using `set` and not `save`? With `set`, the values aren't persisted to localStorage.

src/headless/converse-chat.js
```diff
@@ -1100,7 +1100,17 @@ converse.plugins.add('converse-chat', {
                     return;
                 }
                 if (utils.isNewMessage(message) && this.isHidden()) {
-                    this.save({'num_unread': this.get('num_unread') + 1});
+                    let first_unread = this.get('first_unread');
```
jcbrand, line 1103: `first_unread` doesn't give you any indication what the value is. Given that it's an ID, I think this key should be named `first_unread_id`.

src/headless/converse-muc.js
```diff
@@ -2361,7 +2361,17 @@ converse.plugins.add('converse-muc', {
...
+
+                    if (this.get('num_unread_general') == 0) {
+                      if (first_unread) {
+                          const msg = this.messages.where({'msgid': first_unread});
+                          if (msg.length > 0) msg[0].set("first_unread", false);
+                      }
+                      message.set("first_unread", true);
+                      first_unread = message.get('msgid');
+                    }
+                    const settings = {'first_unread': first_unread, 'num_unread_general': this.get('num_unread_general') + 1};
```
jcbrand, line 2374: Looks to me like this code is mostly a copy/paste of what's in the `ChatBox` model. You can probably move the code in `ChatBox` into a new method and then call it from both places, instead of duplicating it in two places.

spec/chatbox.js
```diff
@@ -1387,8 +1391,10 @@ describe("Chatboxes", function () {
             _converse.handleMessageStanza(msgFactory());
             await u.waitUntil(() => chatbox.messages.length);
             expect(chatbox.get('num_unread')).toBe(1);
+            expect(chatbox.get('first_unread')).toBeDefined();
             _converse.saveWindowState(null, 'focus');
             expect(chatbox.get('num_unread')).toBe(1);
+            expect(chatbox.get('first_unread')).toBeDefined();
```
jcbrand, line 1397: Please add tests to check that the right message gets marked as the first unread.

src/headless/converse-chat.js
deleolajide: > `msgid` isn't guaranteed to be unique.

We don't need this to be unique because it is not persistent. The requirement is for real-time messages only. When messages are re-fetched from MAM or persistent storage, their first_unread message indicator is reset to false.

src/converse-message-view.js
deleolajide: ok. I will remove it

src/headless/converse-chat.js
deleolajide: ok. It will be renamed

src/headless/converse-muc.js
deleolajide: There is slight difference in behavior. That is why you had to override the method in ChatRoom class. I will refactor into a re-usable method in ChatBox class provided the code does not get convoluted.

spec/chatbox.js
deleolajide: ok. I will modify to confirm the expected msgid is set in the first_uread_id attribute

src/headless/converse-chat.js
jcbrand: I think this information should be persisted in localStorage, otherwise when you reload the tab you'll lose the unread marker, even if you haven't looked at the messages yet.

To simplify things (if we don't need to store an attribute from the actual stanza), you can just save the `id` attribute instead of the `msgd`. That's also unique and then you can do a direct lookup instead of using `findWhere` you can just use `this.messages.get(id)`.

deleolajide: Thanks. I will  use id instead of msgid

src/headless/converse-muc.js
deleolajide: Thanks. I implemented new method in ChatBox called setFirstUnreadMsgId.

src/headless/converse-chat.js
```diff
@@ -1110,11 +1110,29 @@ converse.plugins.add('converse-chat', {
...
+            /**
+             * Sets the msgid of the first unread realtime message in a ChatBox.
+             * @param {_converse.Message} message
+             */
+            setFirstUnreadMsgId (message) {
+                if (this.get('num_unread') == 0) {
+                    const first_unread_id = this.get('first_unread_id');
+
+                    if (first_unread_id) {
+                      const msg = this.messages.get(first_unread_id);
```
jcbrand, line 1128: This won't work if you're using the `msgid` value, it must be the `id`.

CHANGES.md
```diff
@@ -21,8 +21,10 @@ Soon we'll deprecate the latter, so prepare now.
 - #1896: Don't send receipts for messages fetched from the archive
 - #1937: Editing a message removes the mentions highlight
 - #1963: Mentions are visually incorrect when used in message replies
+- #1999: Jump to first unread message
```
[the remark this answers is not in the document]
deleolajide: For my sanity and cross referencing, I have used the same issue title as the original issue. I could have changed that.  No, there is no jump. This PR for the issue adds a Converse UI chat separator before the first unread chat message.

src/headless/converse-chat.js
deleolajide: It did pass the test confirming that msgid and id are usually the same value. In any case, I have corrected it. It was a typo.

CHANGES.md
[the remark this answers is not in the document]
deleolajide: Feel free to re-title it as you see fit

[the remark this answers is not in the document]
jcbrand: There's no point in using a wrong and misleading title in the changelog. Please change to "Mark first unread message" or something like that.

src/headless/converse-chat.js
```diff
@@ -1110,11 +1110,29 @@ converse.plugins.add('converse-chat', {
...
+                      if (msg) msg.set("first_unread", false);
+                    }
+                    message.set("first_unread", true);
+                    this.set({'first_unread_id': message.get('id')});
```
jcbrand, line 1132: Instead of setting values on both the message and the chatbox, you can instead only set `first_unread_id` on the chatbox and then compare the message id with `first_unread_id` inside the template (or assign that to a variable and pass it in to the template).

That would simplify the shrink the code here.

I also think it's desirable to use `.save` instead of `.set` so that the value gets persisted across page loads.

deleolajide: @jcbrand At this point, I need to ask fro a favor from you to finish this PR on my behalf. I have run out of time to complete this. I have started a holiday and will be away for a while. By the time I get back, this PR would have expired and I might have to start all over again.

As you have indicated with an emoji reaction like you like it in principle and have a clear idea of how you want it implemented, could please kindly merge it and finish off the implementation. It would really help avoid having to get my head around this PR in addition to all the other things I would have to face when I get back. Thanks.

deleolajide: >Instead of setting values on both the message and the chatbox, you can instead only set first_unread_id on the chatbox 

Suggestions on how I would trigger onChanged  in converse-message-view.js without including first_unread in the list of properties to be monitored would be very much appreciated.

>I also think it's desirable to use .save instead of .set so that the value gets persisted across page loads.

ok.

jcbrand: > Suggestions on how I would trigger onChanged in converse-message-view.js without including first_unread in the list of properties to be monitored would be very much appreciated.

Once we render messages with lit-element, we can simply listen for `change` on the chatbox because rendering will be efficient (if nothing changes in a message's markup the DOM doesn't update).

Unfortunately we're not there yet and currently all messages will be re-rendered every time :(

Sorry, I didn't consider the performance implications of having to re-render all messages when `first_unread_id` changes on the chatbox.

So it's good like this for now.

Please just change the entry the changelog (since there's no jumping to anything) and I'll merge.

jcbrand:
Thanks! Merge manually.

PR #2006, opened 2020-04-28
make use_system_emojis==false work

ChaosKid42:
This PR fixes rendering of emojis in case `use_system_emojis == false`. Unfortunately I couldn't make it work using js-xss so I uses sanitize-html instead.

### DIFF
--- a/src/templates/emoji_picker.js
+++ b/src/templates/emoji_picker.js
@@ -10,7 +10,7 @@ const skintones = ['tone1', 'tone2', 'tone3', 'tone4', 'tone5'];
 
 
 const emoji_category = (o) => {
-    const category_emoji = unsafeHTML(xss.filterXSS(o.transformCategory(o.emoji_categories[o.category]), {'whiteList': {'img': []}}));
+    const category_emoji = unsafeHTML(xss.filterXSS(o.transformCategory(o.emoji_categories[o.category]), {'whiteList': {'img': ['class', 'draggable' ,'alt', 'src', 'title']}}));
     return html`
         <li data-category="${o.category}"
             class="emoji-category ${o.category} ${(o.current_category === o.category) ? 'picked' : ''}"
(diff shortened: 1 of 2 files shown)


src/templates/emoji_picker.js
```diff
@@ -1,16 +1,15 @@
...
 
 const i18n_search = __('Search');
 const i18n_search_results = __('Search results');
 const skintones = ['tone1', 'tone2', 'tone3', 'tone4', 'tone5'];
 
 
 const emoji_category = (o) => {
-    const category_emoji = unsafeHTML(xss.filterXSS(o.transformCategory(o.emoji_categories[o.category]), {'whitelist': {'img': []}}));
+    const category_emoji = unsafeHTML(sanitizeHtml(o.transformCategory(o.emoji_categories[o.category]), {allowedTags: ['img'], allowedAttributes: {'img': ['class', 'draggable' ,'alt', 'src', 'title']}}));
     return html`
```
jcbrand, line 13: Why couldn't you get it to work with js-xss?

I'd really like to avoid adding another 3rd party dependency and having two libraries to sanitize HTML.

Looking at the js-xss README, it appears as if you can whitelist attributes like this:

 ```
  whiteList: {
    'img': ["class", "draggable", "alt", "src"]
  }
```

I've just noticed that the docs says `whiteList` (with a capital L), and we've been using `whitelist`, all small letter, maybe this is the reason?

Comment (ChaosKid42) [reply src/templates/emoji_picker.js:13]: Unfortunately not. See the demo here: https://runkit.com/chaoskid42/5ea84fc3ac39fc0013fee804

Code:
```js
const sanitizeHtml = require("sanitize-html")
const xss = require("xss")
const dirty = '<img class="emoji" draggable="false" alt="😁" src="twemoji/72x72/1f601.png" title=":grin: 😁">'
console.log(dirty)
console.log(sanitizeHtml(dirty, {allowedTags: ['img'], allowedAttributes: {'img': ['class', 'draggable' ,'alt', 'src', 'title']}}))
console.log(xss(dirty, {whiteList: {img: ['class', 'draggable' ,'alt', 'src', 'title']}}))
```

Output:
```
"<img class=\"emoji\" draggable=\"false\" alt=\"😁\" src=\"twemoji/72x72/1f601.png\" title=\":grin: 😁\">"
"<img class=\"emoji\" draggable=\"false\" alt=\"😁\" src=\"twemoji/72x72/1f601.png\" title=\":grin: 😁\">"
"<img class=\"emoji\" draggable=\"false\" alt=\"😁\" src title=\":grin: 😁\">"
```

xss seems to remove the value of the src attribute.

ChaosKid42: I believe I found the problem: It works, if you don't use relative URLs. Pushing a new PR in a minute ...

ChaosKid42: @jcbrand Done. Now it works without adding another 3rd party dependency.

PR #2020, opened 2020-05-05
Prevent tranforming malformed urls and urls with non-approved protocols into hyperlinks

### LINKED ISSUE #2019: Insecure and malformed URLs are being parsed as correct.
**Describe the bug**
Insecure and malformed URLs are being parsed as correct.

**To Reproduce**
1. Open a chat window
2. Type wrong and unsafe URLs and see them getting parsed as correct:
- http://#$%^&*(&%$$%^&*
- file://evil.sh

Screen Shot 2020-05-04 at 8 04 14 PM

**Expected behavior**
Not parsing malformed or insecure URLs

afuggini:
Fixes #2019 

Screen Shot 2020-05-04 at 7 48 17 PM

### DIFF
--- a/src/utils/html.js
+++ b/src/utils/html.js
@@ -22,6 +22,7 @@ import tpl_video from "../templates/video.js";
 import u from "../headless/utils/core";
 
 const URL_REGEX = /\b(https?\:\/\/|www\.|https?:\/\/www\.)[^\s<>]{2,200}\b\/?/g;
+const APPROVED_URL_PROTOCOLS = ['http', 'https', 'xmpp'];
 
 function getAutoCompleteProperty (name, options) {
     return {
@@ -95,7 +96,7 @@ function renderAudioURL (_converse, uri) {
 
 function renderImageURL (_converse, uri) {
     if (!_converse.api.settings.get('show_images_inline')) {
-        return u.convertToHyperlink(uri);
+        return u.convertUriToHyperlink(uri);
     }
     const { __ } = _converse;
     return tpl_image({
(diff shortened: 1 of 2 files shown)


src/utils/html.js
```diff
@@ -403,14 +399,41 @@ u.convertToHyperlink = function (url) {
...
+        const url = new URL(urlString);
+        return !!url;
+    } catch (error) {
+        return false;
+    }
 }
 
+u.convertUrlToHyperlink = function (url) {
+    const uri = getURI(url);
+    if (uri !== null && isUrlValid(url) && isProtocolApproved(uri._parts.protocol)) {
```
jcbrand, line 420: The `uri !== null` check can go into `isUrlValid`, after all, `null` isn't a valid URL :)

Generally in the Converse codebase we write abbreviations as all-caps. So `isURLValid` instead of `isUrlValid`.

src/utils/html.js
```diff
@@ -388,13 +389,8 @@ u.addMentionsMarkup = function (text, references, chatbox) {
     return text;
 };
 
-
-u.convertToHyperlink = function (url) {
-    const uri = getURI(url);
-    if (uri === null) {
-        return url;
-    }
```
jcbrand, line 396: It's not clear to me why you removed these lines, can you please explain why?

jcbrand:
@afuggini: I see your changes have introduced a test failure. Please fix.

https://travis-ci.org/github/conversejs/converse.js/builds/683193248#L2230

src/utils/html.js
afuggini: `uri` is not a url string, but an instance of the `URI` class, whereas the purpose of `isUrlValid` is to check for valid url strings. I moved this check to here from the code that you commented on below, which was then removed.

src/utils/html.js
afuggini: This `uri === null` check is now part of line `420`.

afuggini:
Added fixes and tests

afuggini:
@jcbrand Added fixes and tests. However, I still see errors... would you please shed some light on those?

src/utils/html.js
jcbrand: Ah sorry, I see now they're two different variables.

jcbrand:
Thanks @afuggini I restarted the build on Travis and now all tests are green.

PR #2021, opened 2020-05-07
Adds mailto protocol to safe list

afuggini:


jcbrand:
@afuggini: Thanks, please also update your test to add a `mailto:` URL so that we don't accidentally remove it again in the future.

afuggini:
Done. Although it only parses when you explicitly add `mailto:` to the URL. In the future it might be nice to parse all email address automatically, though.

PR #2031, opened 2020-05-14
fix #2028 - Implement XEP-0333 displayed chat marker

### LINKED ISSUE #2028: Implement XEP-0333 displayed and acknowledged chat markers
Currently, only the **received** chat marker is implemented in Converse. However, Prosody is able to handle **displayed** and **acknowledged** chat markers in order to send accurate [room activity notifications](https://xmpp.org/extensions/xep-0437.html)  back to Converse

**Describe the solution you'd like**
Converse is tracking the total unread messages and first unread message displayed in a chatbox as well as user gestures that reset the counter. When the counter gets reset, mark the last message  as **acknowledged**. All new messages should be marked as **displayed**.

deleolajide:

**Will be done as the last commit to minimize conflict with the CHANGES.md file**

**Existing tests extended to check for outgoing chat marker stanzas**

### DIFF
--- a/src/headless/converse-chat.js
+++ b/src/headless/converse-chat.js
@@ -1114,7 +1114,9 @@ converse.plugins.add('converse-chat', {
                         this.setFirstUnreadMsgId(message);
                     }
                     this.save({'num_unread': this.get('num_unread') + 1});
-                    this.sendMarker(message.get('from'), message.get('msgid'), 'displayed');
+                    if (message.get('is_markable')) {
+                        this.sendMarker(message.get('from'), message.get('msgid'), 'displayed');
+                    }
                     _converse.incrementMsgCounter();
                 }
             },
@@ -1137,7 +1139,9 @@ converse.plugins.add('converse-chat', {
             clearUnreadMsgCounter () {
                 if (this.get('num_unread') > 0) {
                     const msg = this.messages.last();
-                    if (msg) this.sendMarker(msg.get('from'), msg.get('msgid'), 'acknowledged');
+                    if (msg && msg.get('is_markable')) {
+                        this.sendMarker(msg.get('from'), msg.get('msgid'), 'acknowledged');
+                    }
                 }
                 u.safeSave(this, {'num_unread': 0});
             },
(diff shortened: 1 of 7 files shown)


src/headless/converse-chat.js
```diff
@@ -1110,8 +1110,11 @@ converse.plugins.add('converse-chat', {
                     return;
                 }
                 if (utils.isNewMessage(message) && this.isHidden()) {
-                    this.setFirstUnreadMsgId(message);
+                    if (this.get('num_unread') == 0) {
+                        this.setFirstUnreadMsgId(message);
+                    }
                     this.save({'num_unread': this.get('num_unread') + 1});
+                    this.sendMarker(message.get('from'), message.get('msgid'), 'displayed');
```
jcbrand, line 1117: Looks like you need to first check for the `is_markable` boolean saved on the message object before sending out a marker. We don't want to send out markers to messages that didn't ask for them.

deleolajide: I have added the check for is_markable. That had an impact on the tests, so I I had to add markable to the generated text chat message from mock.createChatMessage

deleolajide:
MattJ has just discovered an issue with the sendMarker that affects this PR. The message stanza type is fixed to 'chat' instead of using the type of the received message stanza. Also, it should use the bare JID of the message sender and not the full JID. This is especially true with MUCs

Don't merge this yet. I may have to modify accordingly

deleolajide:
This PR is ready for another review. It makes the following changes
1. modifies sendMarker to use bareJID and the chat type of the receiving messages
2. sends a displayed chat marker in addition to received chat marker when the message is seen by user (ie num_unread and num_unread_general are both zero)
3. fixes the every unread messages highlighting

jcbrand:
Thank you @deleolajide!

I resolved the conflicts locally but then got so many conflicts while rebasing that I just ended up making a new commit based on the code in your branch.

deleolajide:
Sorry my PRs are a bit of a pain. I need to get this right as I have a few more. 

It might be easier to discuss a lot more of the implementation details before I raise a PR to shorten the review/merge cycle We should also decide if there should be a new test or a modification to an existing test. I suggest we do this when I raise the issue. I will can provide more implementation detail for discussion before moving forward with a PR which hopefully, should be much easier to merge :-)

jcbrand:
It's all good, thanks for making the PR.

PR #2041, opened 2020-05-26
Fixed emoji click area and emoji picker tabs

afuggini:


sass/_emoji.scss
```diff
@@ -119,18 +126,15 @@
...
-                        padding: 0.25em 0;
                         font-size: var(--font-size-huge);
                         &.picked {
                             background-color: white;
                             border: 1px var(--chat-head-color) solid;
                             border-bottom: none;
                         }
-                        &.selected a, &:hover a {
-                            background-color: var(--highlight-color-darker);
-                        }
```
jcbrand, line 131: Why did you remove this? The hover effect is now gone for the emoji categories.

sass/_emoji.scss
```diff
@@ -94,7 +97,11 @@
                         }
                         a {
                             padding: 3px;
+                            display: inline-block;
                             font-size: var(--font-size-huge);
+                            height: 30px;
+                            width: 30px;
```
jcbrand, line 103: In changing the width from 32px to 30px, the emoji is no longer centered within the `.insert-emoji` element which gets a background color upon hover. Please fix.

Screenshot from 2020-05-27 07-27-17

sass/_emoji.scss
afuggini: It seems like my cache was showing me a different version where a hover still showed, the intention was to make a visual improvement. I will put that back.

sass/_emoji.scss
afuggini: Are those image-based emojis or is that how the regular ones render in Linux?

jcbrand: These are system (regular) emojis on Linux.

afuggini: Ok check now, I tested with system emojis on Mac and also image emojis by updating the settings. Lmk what it looks like on Linux.

jcbrand: Looks worse to me now, on both Linux and Mac.

Can you please post a screenshot of what it looks like on your machine?

afuggini: I was looking at the fullscreen view, but I was missing the fact that the overlay version has different emoji sizes. So I just fixed that. Here is what overlay and fullscreen look like now

Screen Shot 2020-06-01 at 12 12 53 PM
Screen Shot 2020-06-01 at 12 13 48 PM

afuggini:
I just adjusted image based emojis' sizes too. Please review.

jcbrand:
Thanks @afuggini

PR #2052, opened 2020-05-30
oauth_providers settings should be an array

rgrunbla:
The oauth_providers settings should be an array as a forEach is used on the settings later in the script.

No changelog entry has been made in CHANGES.md as the change is minor.


jcbrand:
Thanks

PR #2061, opened 2020-06-03
Makes mentions case-insensitive, plus parsing functionality refactor

afuggini:
With this PR, nicknames submitted in whatever case are parsed and displayed in the right case after matching with existing nicknames.

It also corrects parsing in some corner cases where we are currently not parsing nicknames properly.

pr_before

pr_after

Additionally we could build additional functionality into it, either now or at a later phase, where as someone types, when spacebar is pressed, the nicknames would format automatically before submission, much like Slack does. Although this wouldn't have much effect over the final delivered message, as that is already being formatted on send.

Still needs testing. Unit testing the helper functions would be relatively easy as they are pure. Integration tests might be desirable.

It is also not working correctly yet in some cases, where the message ends up showing as a URL instead. I am working on it.


src/headless/converse-muc.js
```diff
@@ -941,74 +942,38 @@ converse.plugins.add('converse-muc', {
...
-                if (!longest_match) {
-                    return null;
-                }
-                if ((mention[longest_match.length] || '').match(/[A-Za-zäëïöüâêîôûáéíóúàèìòùÄËÏÖÜÂÊÎÔÛÁÉÍÓÚÀÈÌÒÙ0-9]/i)) {
-                    // avoid false positives, i.e. mentions that have
-                    // further alphabetical characters than our longest
-                    // match.
-                    return null;
-                }
+            getOccupantByNickname (nickname) {
```
jcbrand, line 945: This method looks up occupants by nickname or JID and not just by nickname.

I think it should therefore be named appropriately, e.g. `getOccupant (nick_or_jid) {`

src/headless/converse-muc.js
```diff
@@ -941,74 +942,38 @@ converse.plugins.add('converse-muc', {
...
+                return this.occupants.findOccupant({ nick: nickname })
+                    || u.isValidJID(nickname)
+                    && this.occupants.findOccupant({ jid: nickname });
+            },
 
-                let uri;
-                const occupant = this.occupants.findOccupant({'nick': longest_match}) ||
-                        u.isValidJID(longest_match) && this.occupants.findOccupant({'jid': longest_match});
+            parseTextForReferences (originalMessage) {
```
jcbrand, line 951: Please stick to the convention in this codebase of writing variable names all lower-case with underscores.
See here: https://conversejs.org/docs/html/style_guide.html#underscores-or-camelcase

src/headless/utils/parse-helpers.js
```diff
@@ -0,0 +1,98 @@
...
+ */
+const helpers = {
+    // Captures all mentions, but includes a space before the @
+    mentionRegex: /\s([@][\w_-]+)|^([@][\w_-]+)/ig,
+
+    filterIsNicknameMentioned (mentions) {
+        // testing against lowercase nickname to be able to
+        // produce matches that are case insensitive
+        return nickname => mentions.includes(nickname.toLowerCase());
+    },
```
jcbrand, line 15: I don't see this method being called from anywhere.

Also, it looks quite generic, so we don't have to semantically tie it to nicknames, it could be used to filter for any matching string.

afuggini: I removed the function as you are right I left it unused.

jcbrand:
@afuggini: There is a failing test which needs to be fixed:

> Chrome 83.0.4103.97 (Linux x86_64) A groupchat shown in the groupchats list has an info icon which opens a details modal when clicked FAILED
> 	Uncaught TypeError: _this$model$getOccupa.getDisplayName is not a function thrown
> Chrome 83.0.4103.97 (Linux x86_64): Executed 8 of 414 (1 FAILED) (0 secs / 1.78 secs)

What sometimes happen, is that if one test fails, a bunch of tests after that one will time out.
I'm not sure why this happens, but I think it might be because Jasmine starts subsequent tests before previous ones have finished.

In any case, it might be that when you fix the first test, the others also become green.

src/headless/utils/parse-helpers.js
```diff
@@ -0,0 +1,68 @@
...
+helpers.mapFormatMentions = mention => mention.slice(1).toLowerCase();
+
+// One of the two possible matches for `mention_regex`
+// One will always be undefined
+helpers.mapMentionsInMatches = match => match[1] || match[0];
+
+helpers.mapMentionsToTempReferences = (mention) => ({
+    mentioned_as: mention,
+    type: 'mention'
+});
```
jcbrand, line 41: Looks like most of the `map*` methods take a single item as parameter and not an array, so I think it's misleading to name the helper in the plural form.

I think `mapMentionToTempReference` would therefore be better (same applies to some other map methods).

src/headless/converse-muc.js
```diff
@@ -941,74 +942,40 @@ converse.plugins.add('converse-muc', {
...
+                    this.getOccupant.bind(this),
+                    this.get('jid')
+                )
+
+                const references = mentions
+                    .map(p.mapMentionsToTempReferences)
+                    .reduce(p.reduceReferencesWithNicknames(known_nicknames), [])
+                    .map(p.mapAddUriToReferences(make_uri_from_reference))
+                    .map(p.mapAddCoordsToReferences(indexes))
+                    .map(p.mapCleanReferences);
```
jcbrand, line 970: Instead of mapping over the array three times after `reduce` has been called, you can use the `flow` method from lodash to map only once.

For example (untested):
```
const finalizeReference = flow([
    p.mapAddUriToReferences(make_uri_from_reference),
    p.mapAddCoordsToReferences(indexes),
    p.mapCleanReferences
]);

const reducer = p.reduceReferencesWithNicknames(known_nicknames);

const references = mentions
    .map(p.mapMentionsToTempReferences)
    .reduce(reducer, [])
    .map(finalizeReference);
```

This way, we avoid 2 additional iterations through the array and we also don't repeatedly invoke `reduceReferencesWithNicknames`, `mapAddUriToReferences` and `mapAddCoordsToReferences` (some of which in turn call `map` as well).

afuggini: Please review the new code.
This is still a WIP since there are failing tests.

afuggini:
Updated the PR to accept nicknames with spaces. Made another refactor, now without functional stuff since it didn't make sense anymore. I can delete the functional helpers, unless you think they can be useful later @jcbrand

jcbrand:
> I can delete the functional helpers, unless you think they can be useful later

Please delete all functions that aren't being called from anywhere.

afuggini:
Ready.

src/headless/converse-muc.js
```diff
@@ -941,74 +942,82 @@ converse.plugins.add('converse-muc', {
                 ])].filter(n => n);
             },
 
-            getReferenceForMention (mention, index) {
-                const nicknames = this.getAllKnownNicknames();
-                const longest_match = u.getLongestSubstring(mention, nicknames);
```
jcbrand, line 946: Please remove `getLongestSubstring`, given that it's no longer being called from anywhere.

jcbrand:
@afuggini I restarted Travis twice to try and get your tests green, but there is a particular test that fails every time:

https://travis-ci.org/github/conversejs/converse.js/builds/697702114#L2831

I don't see how it's connected to this PR, but without this PR all tests are green.

Can you please investigate and see whether you find the issue?

src/headless/converse-muc.js
afuggini: As you see it is in red, it is removed already :)

afuggini:
Fixed.

src/headless/converse-muc.js
```diff
@@ -941,74 +942,83 @@ converse.plugins.add('converse-muc', {
...
+                    }
+                    return matches;
+                  }, {});
+                  return { ...matches_by_index, ...matches };
+                };
+                
+                const mentions_by_index = known_nicknames
+                    .map(makeObjectFromNick)
+                    .map(addMatchesForNick)
+                    .reduce(toIndexedNicknames, {})
```
jcbrand, line 990: I've been thinking about this PR tonight, and this code in particular makes me nervous.

In some large MUCs, you can have thousands of users, which means we would be running thousands of regexes every single time someone sends a message, which I think might become noticeable to the user as small freeze-ups. 

Additionally, I've been thinking about the fact that we add mentions for nicknames that aren't prefixed with an `@` symbol. Slack for  example doesn't highlight a nickname if there's no `@` in front of it, and I think that makes sense, because sometimes you want to refer to someone without disturbing them, and sometimes someone has a very generic nickname and you're not actually referring to them at all.

I think we could kill two birds with one stone here... if we update the code to only add mentions if the nickname is preceded by an `@` symbol, and to then to stop matching as soon as we have as many matching nicknames as `@` symbols in the message. Similarly, we can avoid running the regexes entirely if there aren't any `@` symbols in the message. That way we avoid some unnecessary regex matching.

What do you think @afuggini?

afuggini: Let me reflect on this and get back to you.

Also FYI I still need to add a test.

jcbrand:
Thanks @afuggini!

PR #2085, opened 2020-06-11
Moves click-to-minimize event listener to title container

afuggini:


jcbrand:
Thanks @afuggini, I merged this manually.

BTW, a tip for next time, you also needed to add an event handler for the `ChatBoxView`.

See here: https://github.com/conversejs/converse.js/commit/85688d1871f023021bd17abd721e66eae3b5dacf

PR #2089, opened 2020-06-18
Makes message with errors non-editable

afuggini:


src/components/message-history.js
```diff
@@ -15,7 +15,7 @@ const tpl_message = (o) => html`
         .hats=${o.hats}
         .model=${o.model}
         ?correcting=${o.correcting}
-        ?editable=${o.editable}
+        ?editable=${o.editable && !o.error}
```
jcbrand, line 18: I don't think this is a proper fix. Instead, the `editable` boolean needs to be set to `false` if the message has an error.

See here: https://github.com/conversejs/converse.js/blob/7c29ae4a8ae69e1c7ff60b0bf15dcb29fa686ead/src/headless/converse-muc.js#L629

and here: https://github.com/conversejs/converse.js/blob/7c29ae4a8ae69e1c7ff60b0bf15dcb29fa686ead/src/headless/converse-chat.js#L405

Otherwise there are other parts in the code where messages will be filtered according to the `editable` value and we will then get false positives (see for example here: https://github.com/conversejs/converse.js/blob/7c29ae4a8ae69e1c7ff60b0bf15dcb29fa686ead/src/converse-chatview.js#L945), and it's not desirable to have to know that everywhere you check for `editable` you also need to check if `error` is set.

Please also add a test for this. Specifically a test that the pencil icon doesn't appear, that you cannot edit the message by pressing the up arrow and that `editable` is `false`.

The existing tests for editing messages are here: https://github.com/conversejs/converse.js/blob/7c29ae4a8ae69e1c7ff60b0bf15dcb29fa686ead/spec/messages.js#L81 and here:  https://github.com/conversejs/converse.js/blob/master/spec/muc_messages.js#L547

You can write similar tests to those.

afuggini:
Updated, please check.

jcbrand:
Thanks!

PR #2098, opened 2020-06-24
WIP: Emoji Refactor

jcbrand:
* Fix rendering of custom emojis
* Turn the emoji picker into a web component
* Optimize the emoji picker by reducing the amount of times it renders when you open it
* Add title attribute to emojis in the chat history

I have more performance-related updates planned for the emoji picker web component, however I think they might better go into a new MR, so that we can get these changes out ASAP.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -117,7 +117,7 @@
         "no-console": "off",
         "no-catch-shadow": "off",
         "no-cond-assign": [
-            "error",
+            "off",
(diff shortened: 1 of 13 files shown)


jcbrand:
@afuggini: thanks for the feedback. A lot of this code was just copy/pasted from the view to the component. If I were to write it new, I would probably have used optional chaining as you've suggested.

PR #2100, opened 2020-06-24
Updates textarea selector when applying focus after emoji selection

afuggini:
The current selector is very broad and may have undesired effects outside Converse. Used a more specific selector instead.

Also fixes textarea not focusing in some edge cases by adding stopPropagation and preventDefault.


jcbrand:
Thanks!

PR #2115, opened 2020-07-02
CSS: Enlarge room title in 'embedded' view mode

Zash:
This makes it the same size as in the 'fullscreen' view mode.

Before:
image

With `view_mode = "fullscreen"`
image

After:
image


jcbrand:
Thanks, I've made a small fix and merged it manually.

PR #2131, opened 2020-07-09
Jcbrand/body transform refactor

jcbrand:
Refactor how the message body text is updated in order to add rich markup features.

We now have a MessageText class which subclasses String and which allows you to store lit-html TemplateResult objects as metadata on it. These TemplateResult objects represent the markup parts of the message.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -231,4 +231,4 @@ doc: node_modules docsdev apidoc
 
 PHONY: apidoc
 apidoc:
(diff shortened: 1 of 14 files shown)


afuggini:
I think the code looks good. And I like this approach with clean, short functions and classes.

I found a bug with emojis, though, the dropdown doesn't show and I see an error in the console.

Screen Shot 2020-07-09 at 11 46 16 AM

jcbrand:
Thanks @afuggini, I found and fixed the bug and merged.

PR #2173, opened 2020-07-31
WIP: Chatview plugin on its own folder (PoC)

afuggini:


src/plugins/chatview/index.js
```diff
@@ -0,0 +1,86 @@
+/**
+ * @module converse-chatview
+ * @copyright 2020, the Converse.js contributors
+ * @license Mozilla Public License (MPLv2)
+ */
+import { __ } from '@converse/headless/i18n';
+import { _converse, api, converse } from "@converse/headless/converse-core";
+
+// Plugins
```
jcbrand, line 9: I think once the plugins are all moved into the plugins folder, the nature of these imports will be clearer, and comments such as this won't be necessary. It's IMO already unnecessary for the components being imported below.

src/plugins/chatview/settings.js
```diff
@@ -0,0 +1,21 @@
+const settings = {
```
jcbrand, line 1: I'm a little ambivalent about having such small files, just for storing settings, I think an argument could be made that it makes discoverability harder because you have everything spread out into so many files that aren't all visible in your editor.

That said... if we were to go the route of putting settings into separate files, we could also make them JSON files.

src/plugins/chatview/views/ChatBoxView.js
```diff
@@ -0,0 +1,1005 @@
+/**
```
jcbrand, line 1: We don't use camel case for any file names so far, so I think we should avoid it for the views as well.

src/plugins/chatview/views/ChatBoxView.js
```diff
@@ -0,0 +1,1005 @@
...
+import tpl_chatbox_head from "../templates/chatbox_head.js";
+import tpl_chatbox_message_form from "../templates/chatbox_message_form.js";
+import tpl_toolbar from "../templates/toolbar.js";
+
+const { dayjs, utils: u } = converse.env;
+
+const ChatBoxView = View.extend({
+  length: 200,
+  className: 'chatbox hidden',
+  is_chatroom: false,  // Leaky abstraction from MUC
```
jcbrand, line 23: Looks like some lines are indented only two spaces.

src/plugins/chatview/views/UserDetailsModal.js
```diff
@@ -0,0 +1,97 @@
+import log from "@converse/headless/log";
```
jcbrand, line 1: Modals are being moved to the `src/modals` folder. AFAIK, there's nothing about the `UserDetailsModal` that makes it specific to this plugin, and I think it should therefore also go there.

This also brings up the topic of which files to keep inside plugins and which files to make more "global" in order to facilitate code reuse, and this applies to modals but also to components (and probably other files as well).

I'm concerned that if we put relatively general purpose modals and components inside plugin folders, that we're creating too tight coupling between them and the plugins. I'm also concerned that if we have a "global" `modals` or `components` folders, and also the same folders inside the plugin folders, that we make discoverability harder.

src/plugins/chatview/index.js
afuggini: I don't think these comments are necessarily keepers in the long term, just used them here as a way to make the import categories super obvious, but they don't need to make it to production.

src/plugins/chatview/settings.js
afuggini: I am not sure I fully share the concern around the number of files. I think small files with a crystal clear name, that are next to another small set of files with the same characteristics, is not a complex structure to be able to understand at first glance. Also, mentally I much prefer that approach which makes it easier for my brain to quickly capture where is what, and its hierarchy, than having to mentally split different parts of the code to understand it every time I edit one single, bigger file.
The reason why I would put settings in a separate file (not religiously, but at least when they surpass a certain number of lines) is to separate config from functionality. So that when I am working on a certain functionality, and I don't care about the specific settings, I don't have to see them.
Of course this is not an absolute truth, and neither do all plugins have to be structured in the same way.

src/plugins/chatview/views/UserDetailsModal.js
afuggini: Moreover, if we are working on a plugin-based architecture, maybe we should avoid as much as possible, if not completely, anything more global than a plugin. If the plugin architecture is strong enough, and I think it is, we should probably put any "global" code in a plugin of its own. E.g.
- `converse-modals`
- `converse-utils`
- etc

src/plugins/chatview/settings.js
jcbrand: Yeah, it is a relatively common pattern to separate config from functionality.

What do you think about making it a JSON file, to make it even clearer that this is config?

src/plugins/chatview/views/UserDetailsModal.js
jcbrand: >  Why not have a converse-core or converse-global plugin for global elements such as modals?

The purpose of plugins is to make it possible to add and/or remove functionality to Converse. In other words, for something to be a plugin, it must make sense to be able to remove that plugin from your build and still have something functional (for a particular use-case of course).

Seen in this way, it doesn't make sense to put code that is so core to Converse that it won't be able to function without it in a plugin.

That said, we can probably have a `core`, `global` or `shared` folder which contains those files that are too central to belong to any one plugin.

> Or for modals, even have its own plugin, converse-modals

We do have a `converse-modal` plugin, but it's more concerned with making it possible to create modals, and I didn't consider it as a container for all "global" modals. But now that you mention it, I guess that could work yeah.

So we can put plugin-specific modals in the plugin itself, and more general modals go into the `modal` plugin folder.

I still think we might run into situations where it's not 100% clear where something should go or where it's confusing to people that a particular file is in one place and not somewhere else, but I think generally the layout is logical.

> Moreover, if we are working on a plugin-based architecture, maybe we should avoid as much as possible, if not completely, anything more global than a plugin

I think we can get quite far with plugins, but I'd be wary of abusing the notion of a plugin in order to maintain some kind of file structure, and as mentioned above, it doesn't make semantic sense to me to put code that is so central that Converse can't ever function without it, inside a plugin.

For the headless build, that's mainly the code in `converse-core.js`, for the UI layer, we might actually be able to put everything into plugins, we'll have to see.

PR #2175, opened 2020-08-04
Adds converse-message plugin + url query params filter plugin

afuggini:
`converse-message` can serve as a utility container for messages, e.g. message filters and transformers.

### DIFF
--- a/src/converse-chatboxviews.js
+++ b/src/converse-chatboxviews.js
@@ -71,7 +71,9 @@ converse.plugins.add('converse-chatboxviews', {
                 /* Override method from backbone.js
                  * If the #conversejs element doesn't exist, create it.
                  */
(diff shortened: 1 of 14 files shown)


src/converse-chatboxviews.js
```diff
@@ -71,7 +71,9 @@ converse.plugins.add('converse-chatboxviews', {
                 /* Override method from backbone.js
                  * If the #conversejs element doesn't exist, create it.
                  */
-                if (!this.el) {
+                if (this.el) {
+                    this.setElement(result(this, 'el'), false);
+                } else {
```
jcbrand, line 76: Looks like there are changes in this diff that I made. I'm not sure why. I guess you need to rebase onto the latest master, or you worked off of a branch instead of master.

PR #2178, opened 2020-08-04
WIP: Message filters and middleware plugin

afuggini:
Includes built-in functionality to prevent double parsing / reference collision.

Former `MessageText` class is now `RichMessage` to differentiate from the raw text messages.

---

From the `README.md` file:

# MessageParser plugin

## Filters

Filter functions transform the raw message text before middleware kicks in.

They receive a string as argument and return a transformed string. 

```javascript
function exampleFilter (message_string) {
    const transformed_message_string = doSomeMagicTo(message_string);
    return transformed_message_string;
}

api.message.addFilters(exampleMiddleware);
```

## Middleware

Middleware functions take the raw message text and return an array of reference objects to be added to the RichMessage object.

```javascript
function exampleMiddleware (message_string) {
    const matching_urls = message_string.matchAll(/https?:\/\/(\w+\.com)/gi);
    const references = Array.from(matching_urls).map(url => ({
        begin: url.index,
        end: url.index + url[0].length,
        template: `<a href="${url[0]}">${url[0]}</a>`
    }));
    return references;
}

api.message.addMiddleware(exampleMiddleware);
```

## Note:

Filters update the message text, affecting XMPP reference positions. This is something to solve.

### DIFF
--- a/src/plugins/message-parser/MessageParser.js
+++ b/src/plugins/message-parser/MessageParser.js
@@ -1,0 +1,86 @@
+import { html } from 'lit-html';
+import { pipe } from '@converse/headless/utils/functional';
+import RichMessage from './RichMessage';
+import log from '@converse/headless/log';
+
+class MessageParser {
+
+    constructor () {
+        this._filters = [];
+        this._middleware = [];
+    }
+
+    get filters () {
+        return this._filters;
+    }
+
+    get middleware () {
+        return this._middleware;
+    }
+
+    static checkReferenceCollision (reference, rich_message) {
+        const collides = rich_message.references.some(msg_ref => {
+          const is_after = reference.begin > msg_ref.end;
+          const is_before = reference.end < msg_ref.begin;
+          return !is_after && !is_before;
+        });
+        return collides;
+    }
+
+    static makeTemplateResultFromString (string_or_templateresult) {
+        const is_template_result = !!string_or_templateresult.getTemplateElement;
+        return is_template_result
+          ? string_or_templateresult
+          : html`${string_or_templateresult}`;
+    }
+
+    static addSingleReferenceToMessage (reference, rich_message) {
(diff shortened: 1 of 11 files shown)


src/plugins/message-parser/MessageParser.js
```diff
@@ -0,0 +1,86 @@
...
+        this._middleware.push(...middleware);
+    }
+
+    richMessageFromText (text) {
+        return new RichMessage(text);
+    }
+
+    runFilters (text) {
+        return pipe(...this.filters)(text);
+    }
```
jcbrand, line 67: We cannot have filters that remove parts of the original message, because that would break all the  indexes in the references we have.

The original message text must be considered immutable and read-only.

That's why I created the `MessageText` class in the first place, so that we can save metadata (as `TemplateResult` instances in the `references` attribute) on the `MessageText` instance without having to mutate the original text.

Instead of having filters, I think we can achieve the same result by simply sticking to the current approach whereby we add references that replace the parts that would have been filtered out with `TemplateResult` instances that render nothing (or an empty Text node).

src/plugins/message-parser/MessageParser.js
```diff
@@ -0,0 +1,86 @@
...
+        );
+    }
+
+    addFilters (...filters) {
+        this._filters.push(...filters);
+    }
+
+    addMiddleware (...middleware) {
+        this._middleware.push(...middleware);
+    }
```
jcbrand, line 59: I like the idea of having a middleware pipeline that 3rd party plugins can use to add more rich content to messages.

The middleware will have to conform to some strict rules. For example, we can't have rich parts of the message overlap one another.

jcbrand:
@afuggini: This MR uses the new planned file structure, which we only want to implement after the Converse 7 release. I think this will therefore have to wait for that release.

jcbrand:
The message formatting code has seen significant changes now with the introduction of markdown support.

PR #2187, opened 2020-08-09
merge only relevant settings when calling api.settings.extend

XaviFP:
Conversejs is unable to start when a shadowRoot is provided as root in settings through converse.initialize. The first plugin calling settings.extend triggers an exception in https://github.com/conversejs/converse.js/blob/master/src/headless/converse-core.js#L665 because it tries to merge the whole initialization_settings defined by the keys in DEFAULT_SETTINGS. As shadowRoot elements have read-only attributes, when the merge tries to overwrite those attributes there's an exception raised .

`TypeError: Cannot assign to read only property 'mode' of object '#<ShadowRoot>'`

The proposed change here solves the issue extending and merging only the settings that  are relevant for the current plugin while keeping the overwrite from the settings passed in to Converse through converse.initialize method (if any).

Although this handles read-only attributes on start-up, if a plugin tried to overwrite a setting containing read-only attributes it would need some kind of custom merger.

### DIFF
--- a/spec/converse.js
+++ b/spec/converse.js
@@ -354,7 +354,7 @@ describe("Converse", function() {
             done();
         }));
         
-        it("extended via settings.extend works and don't override settings passed in via converse.initialize when using a shadowRoot as root specified via converse.initialize",
+        it("only overrides the passed in properties",
                 mock.initConverse([],
                 {
                     'root': document.createElement('div').attachShadow({ 'mode': 'open' }),
@@ -369,7 +369,7 @@ describe("Converse", function() {
                         'emoji_categories': { 'travel': ':motorcycle:', 'food': ':burger:' },
                     });
 
-                    expect(_converse.api.settings.get('emoji_categories')?.travel).toBe(':rocket:');
+                    expect(_converse.api.settings.get('emoji_categories').travel).toBe(':rocket:');
                     expect(_converse.api.settings.get('emoji_categories')?.food).toBe(undefined);
                     done();
                 }
--- a/spec/converse.js
+++ b/spec/converse.js
@@ -370,7 +370,7 @@ describe("Converse", function() {
                     });
 
                     expect(_converse.api.settings.get('emoji_categories').travel).toBe(':rocket:');
-                    expect(_converse.api.settings.get('emoji_categories')?.food).toBe(undefined);
+                    expect(_converse.api.settings.get('emoji_categories').food).toBe(undefined);
                     done();
                 }
             )
--- a/spec/converse.js
+++ b/spec/converse.js
@@ -353,6 +353,28 @@ describe("Converse", function() {
             expect(_converse.api.settings.get('emoji_categories')?.food).toBe(undefined);
             done();
         }));
+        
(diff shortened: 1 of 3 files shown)


jcbrand:
Hi @XaviFP, thanks for the PR and the explanation.

Looks good AFAICT. Can you please also add a test for this?

There are already two tests related to extending settings, you can add another one here:
https://github.com/conversejs/converse.js/blob/696c8e35220a99270e05fa470db38a7a7668b55a/spec/converse.js#L326

The 2nd parameter sent to `mock.initConverse` is an object containing the default settings. And then in the test you can call `api.settings.extend` again.

To create a shadowRoot, you can create a div via `document.createElement` and then call [attachShadow](https://developer.mozilla.org/en-US/docs/Web/API/Element/attachShadow).

spec/converse.js
```diff
@@ -353,6 +353,28 @@ describe("Converse", function() {
             expect(_converse.api.settings.get('emoji_categories')?.food).toBe(undefined);
             done();
         }));
+        
+        it("extended via settings.extend works and don't override settings passed in via converse.initialize when using a shadowRoot as root specified via converse.initialize",
```
jcbrand, line 357: LMAO, can you make this a little bit shorter?

What about `it("only overrides the passed in properties"` ?

jcbrand: Otherwise, looks good to be merged :)

spec/converse.js
```diff
@@ -353,6 +353,28 @@ describe("Converse", function() {
...
+                (done, _converse) => {
+                    expect(_converse.api.settings.get('emoji_categories')?.travel).toBe(':rocket:');
+
+                    // Test that the extend command doesn't override user-provided site
+                    // settings (i.e. settings passed in via converse.initialize).
+                    _converse.api.settings.extend({
+                        'emoji_categories': { 'travel': ':motorcycle:', 'food': ':burger:' },
+                    });
+
+                    expect(_converse.api.settings.get('emoji_categories')?.travel).toBe(':rocket:');
```
jcbrand, line 372: I don't think you need the optional chaining here, since 'emoji_categories' *will* be defined.

XaviFP:
:rofl: Sure no problem :+1:

spec/converse.js
```diff
@@ -369,7 +369,7 @@ describe("Converse", function() {
                         'emoji_categories': { 'travel': ':motorcycle:', 'food': ':burger:' },
                     });
 
-                    expect(_converse.api.settings.get('emoji_categories')?.travel).toBe(':rocket:');
+                    expect(_converse.api.settings.get('emoji_categories').travel).toBe(':rocket:');
                     expect(_converse.api.settings.get('emoji_categories')?.food).toBe(undefined);
```
jcbrand, line 373: Hey, not that it's super important, but there is one more places where optional chaining is used and not necessary.

XaviFP: Yup, no worries :-) Thank you!

jcbrand:
Thanks!

PR #2188, opened 2020-08-10
WIP: Adds ability to ban users as a moderator

afuggini:


src/headless/converse-muc.js
```diff
@@ -2436,6 +2436,12 @@ converse.plugins.add('converse-muc', {
                     'num_unread': 0,
                     'num_unread_general': 0
                 });
+            },
+
+            async onBanUser (message, reason) {
+                const jid = message.get('from');
+                const room = await api.rooms.get(''); //@TODO get proper room
+                room.sendAffiliationIQ('outcast', { jid, reason });
```
jcbrand, line 2444: You created the `onBanUser` method on the ChatRoom model, so there's no need to get the room via the api, you can just call `this.sendAffiliationIQ('outcast', { jid, reason })`.

src/components/message-actions.js
```diff
@@ -44,8 +44,25 @@ class MessageActions extends CustomElement {
         this.chatview.onMessageRetractButtonClicked(this.model);
     }
 
+    async onBanUserButtonClicked (ev) {
+        ev.preventDefault();
+        const reason = await api.prompt(__('Reason'), __('Type a reason for the ban'));
+        this.chatview.onBanUser(this.model, reason);
```
jcbrand, line 50: Did you test this code?

You're calling `onBanUser` on the view, but you created `onBanUser` on the model (which is IMO the right place). So you'll get an AttributeError when this code runs.

Every view has a `model` attribute which references the model it applies to, so you could call `this.chatview.model.onBanUser(this.model, reason);` (in case it's not clear, `this.chatview.model` is the `ChatBox` instance and `this.model` is the `Message` instance).

Also, for an event handler, I think using the `on` prefix is fine, but it doesn't make sense to me for general methods. So I think `onBanUser` should be renamed to just `banUser`.

That said, it's not clear to me that we even need a `banUser` method, I think you could probably just call `this.chatview.model.sendAffiliationIQ` from here.

jcbrand:
I'm working on this feature in the jcbrand/quick-ban branch.

PR #2199, opened 2020-08-20
Fix BOSH session restore

XaviFP:
After refactor in https://github.com/conversejs/converse.js/commit/16ca8044f814299b1b39d682678978f05d92b975#diff-ed7b4b012c33f7e3c09eaaf7df1bd15a function call fell out of sync.


jcbrand:
Thanks!

PR #2201, opened 2020-08-24
add html to converse.env

ChaosKid42:
I'd like to have `html` from `lit-element` in `converse.env`. Otherwise it does not seem to be possible to add toolbar-buttons from plugins. (See here: https://github.com/Polymer/lit-element/issues/390)


jcbrand:
Thanks, good to see you're still active :)

PR #2202, opened 2020-08-25
add Overview to converse.env

ChaosKid42:
Another proposedm addition to converse.env:

It would be nice to be able to access `Overview` from a plugin.


jcbrand:
@ChaosKid42 What are you using Overview for? I'm not against making it available, but be aware that it's deprecated and we're planning to replace all Views and Overviews with LitElement components.

ChaosKid42:
@jcbrand It's used in the download dialog plugin (https://github.com/conversejs/community-plugins/pull/44/). But as I'm currently working on it, I can as well try to replace Overview with LitElement.

PR #2203, opened 2020-08-25
Refactors notification function

afuggini:
The original intention was to remove the notification by default when a user is mentioned (captured by RegExp).
However, ideally I would have this function return `true` only in the required cases, vs always returning `true` with exceptions.
Is there any other case where this function should return `true`, besides when `notify_all` is true or has a room jid?

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -1348,6 +1348,13 @@ notification_icon
 This option specifies which icon is shown in HTML5 notifications, as provided
 by the ``src/converse-notification.js`` plugin.
 
+notify_nicknames_without_references
+-----------------------------------
+
+* Default: ``false``
+
+Enables desktop notification when user nickname is mentioned without the ``@`` sign.
+
 oauth_providers
 ---------------
 
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -1353,7 +1353,10 @@ notify_nicknames_without_references
 
 * Default: ``false``
 
-Enables desktop notification when user nickname is mentioned without the ``@`` sign.
+Enables notifications for nicknames in messages that don't have associated
+XEP-0372 references linking them to the JID of the person being mentioned.
+
+In Converse, these would be nicknames that weren't mentioned via the ``@`` sign.
 
 oauth_providers
 ---------------
--- a/src/converse-notification.js
+++ b/src/converse-notification.js
@@ -36,7 +36,7 @@ converse.plugins.add('converse-notification', {
             notify_nicknames_without_references: false
         });
 
-        _converse.shouldNotifyOfGroupMessage = function (message) {
+        _converse.shouldNotifyOfGroupMessage = function (message, data) {
             /* Is this a group message worthy of notification?
              */
             const jid = message.getAttribute('from');
@@ -46,15 +46,18 @@ converse.plugins.add('converse-notification', {
             const resource = Strophe.getResourceFromJid(jid);
             const sender = resource && Strophe.unescapeNode(resource) || '';
             let is_mentioned = false;
-            
+            const nick = room.get('nick');
+
             if (api.settings.get('notify_nicknames_without_references')) {
                 const body = message.querySelector('body');
-                is_mentioned = (new RegExp(`\\b${room.get('nick')}\\b`)).test(body.textContent);
+                is_mentioned = (new RegExp(`\\b${nick}\\b`)).test(body.textContent);
             }
-            
-            const is_not_mine = sender !== room.get('nick');
+
+            const is_referenced = data.attrs.references.map(r => r.value).includes(nick);
+            const is_not_mine = sender !== nick;
             const should_notify_user = notify_all === true
                 || (Array.isArray(notify_all) && notify_all.includes(room_jid))
+                || is_referenced
                 || is_mentioned;
             return is_not_mine && !!should_notify_user;
         };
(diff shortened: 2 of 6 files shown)


jcbrand:
@afuggini Looks good. You just need to update/fix the failing tests as well, then I'll merge.

afuggini:
@jcbrand I removed the tests that expected a mention without `@` to show a notification, and also made notifications work on mentions with `@`.

What kind of setting do you mean we should add?

jcbrand:
@afuggini:

A settings such as `notify_nicknames_without_references` which if set to `true` keeps the old behaviour.

Not all XMPP clients support XEP-372 references and I believe some people will still expect to be notified if their nickname is mentioned in a message, without references.

afuggini:
@jcbrand added a setting for this purpose, and tests.

src/converse-notification.js
```diff
@@ -32,7 +32,8 @@ converse.plugins.add('converse-notification', {
             play_sounds: true,
             sounds_path: api.settings.get("assets_path")+'/sounds/',
             notification_icon: 'logo/conversejs-filled.svg',
-            notification_delay: 5000
+            notification_delay: 5000,
+            notify_nicknames_without_references: false
```
jcbrand, line 36: When adding a new configuration option, please also document it here:

Also, please update one of your tests to also check this configuration setting.

src/converse-notification.js
```diff
@@ -32,32 +32,28 @@ converse.plugins.add('converse-notification', {
...
+                const room = _converse.chatboxes.get(room_jid);
+                const body = message.querySelector('body');
+                is_mentioned = (new RegExp(`\\b${room.get('nick')}\\b`)).test(body.textContent);
             }
-            return true;
+            
+            const result = notify_all === true
+                || (Array.isArray(notify_all) && notify_all.includes(room_jid))
+                || is_mentioned;
+            return !!result;
```
jcbrand, line 56: Looks like you removed this line:
```
            if (sender === room.get('nick') || (!notify_all && !mentioned)) {	
                return false;	
            }
```

which means that the user will also be notified for messages he sends himself, which is not what we want.
This would have been easily caught with a test if we had one for it. Please update the tests to check that users aren't notified for messages they send themselves.

I think you can just put this line near the beginning of the method to fix this:
```
            if (sender === room.get('nick')) {
                return false;	
            }
```

jcbrand (changes requested):
There's a failing test on Travis which appears to be related to the changes you made here:
https://travis-ci.org/github/conversejs/converse.js/builds/721844049#L2026

afuggini:
There seems to be an error here which I can't relate to the changes I made:
https://travis-ci.org/github/conversejs/converse.js/builds/724296150#L1204
@jcbrand can you spot the problem?

jcbrand:
I tested your branch locally, and now `@` notifications don't work.

You'll need to do some refactoring in order to fix this.

`shouldNotifyOfMessage` gets called by an event handler of the `message` event.

Firstly, you need to find every place where the `message` event is triggered (looks like `src/headless/converse-chat.js` and `src/headless/converse-muc.js`) and update it so that the parsed attributes are also included in the triggered event data.

In other words, replace `api.trigger('message', { stanza })` to `api.trigger('message', { stanza, attrs });`

Then you need to make sure that `shouldNotifyOfMessage` uses the `attrs` object to get the references array, which is an array of nicknames mentioned via the `@` prefix and then if the user's nickname is in that array, they should be notified.

docs/source/configuration.rst
```diff
@@ -1348,6 +1348,13 @@ notification_icon
 This option specifies which icon is shown in HTML5 notifications, as provided
 by the ``src/converse-notification.js`` plugin.
 
+notify_nicknames_without_references
+-----------------------------------
+
+* Default: ``false``
+
+Enables desktop notification when user nickname is mentioned without the ``@`` sign.
```
jcbrand, line 1356: This text isn't precise enough.

 It's not really about the `@` sign. It's about whether an incoming stanza includes [XEP-372](https://xmpp.org/extensions/xep-0372.html) references that maps JIDs to nicknames.

Converse allows the user to use the `@` sign add references, but other XMPP clients don't, and the `@` sign isn't event sent over the wire.

Alternative suggestion:

```
Enables notifications for nicknames in messages that don't have associated XEP-0372 references linking them to the JID of the person being mentioned. In Converse, these would be nicknames that weren't mentioned via the ``@`` sign.
```

jcbrand:
@afuggini I still don't get notifications when I'm mentioned in a MUC that's not currently visible. I get the impression you're not really testing your changes. This also looks like something a test could be written for.

src/converse-notification.js
```diff
@@ -32,32 +32,34 @@ converse.plugins.add('converse-notification', {
...
+            let is_mentioned = false;
+            const nick = room.get('nick');
+            
+            if (api.settings.get('notify_nicknames_without_references')) {
+                const body = message.querySelector('body');
+                is_mentioned = (new RegExp(`\\b${nick}\\b`)).test(body.textContent);
             }
-            return true;
+            
+            const is_referenced = data.attrs.references.map(r => r.value).includes(`@${nick}`);
```
jcbrand, line 56: This line doesn't work because we don't include the `@` in the reference data.

Please add a test for the case where a message with references is received, to check that a user is properly notified.

afuggini: I added a test that is passing. The references array contains object with a key `value` which has the `@` sign, at least that is what the test environment showed.

afuggini:
@jcbrand why was this closed?

jcbrand:
@afuggini It was merged manually.

PR #2209, opened 2020-08-27
adds setting to set non-image-extension image urls

afuggini:
Example RegExp:

`/^https?:\/\/(?:www.)?(?:imgur\.com\/\w{7}|tradingview.com\/\w\/\w{8})\/?$/i`

Parsing function unit-tested here:
https://codepen.io/arielfuggini/pen/YzqVjyg

### DIFF
--- a/src/converse-chatview.js
+++ b/src/converse-chatview.js
@@ -67,6 +67,7 @@ converse.plugins.add('converse-chatview', {
                 'emoji': true,
                 'spoiler': true
             },
+            'image_urls_regex': null
         });
 
 
--- a/src/converse-chatview.js
+++ b/src/converse-chatview.js
@@ -52,6 +52,7 @@ converse.plugins.add('converse-chatview', {
         api.settings.extend({
             'auto_focus': true,
             'debounced_content_rendering': true,
+            'image_urls_regex': null,
             'message_limit': 0,
             'muc_hats_from_vcard': false,
             'show_images_inline': true,
@@ -67,7 +68,6 @@ converse.plugins.add('converse-chatview', {
                 'emoji': true,
                 'spoiler': true
             },
-            'image_urls_regex': null
         });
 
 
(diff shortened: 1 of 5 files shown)


src/converse-chatview.js
```diff
@@ -67,6 +67,7 @@ converse.plugins.add('converse-chatview', {
                 'emoji': true,
                 'spoiler': true
             },
+            'image_urls_regex': null
```
jcbrand, line 70: When adding a new configuration option, please also document it here:

Options are in alphabetical order.

Also, please provide a test for this. In the test you can pass in a regex for `image_urls_regex` and test that images render.

Your test will be something similar to this:
https://github.com/conversejs/converse.js/blob/823134d7629bad2b82875964169340609c0ce2c9/spec/messages.js#L937

src/utils/html.js
```diff
@@ -75,8 +75,12 @@ function checkFileTypes (types, url) {
 }
 
 u.isAudioURL = url => checkFileTypes(['.ogg', '.mp3', '.m4a'], url);
-u.isImageURL = url => checkFileTypes(['.jpg', '.jpeg', '.png', '.gif', '.bmp', '.tiff', '.svg'], url);
 u.isVideoURL = url => checkFileTypes(['.mp4', '.webm'], url);
+u.isImageURL = url => checkFileTypes(['.jpg', '.jpeg', '.png', '.gif', '.bmp', '.tiff', '.svg'], url);
+u.isWhitelistedImageURL = url => {
+    const regex = _converse.api.settings.get('image_urls_regex');
+    return regex ? regex.test(url) : false;
+}
```
jcbrand, line 83: I don't see any place where `isImageURL` is called without also having to call `isWhitelistedImageURL`, so it seems unnecessary to me to have two methods.
Instead, `isImageURL` can also do the regex test.

If it ever were necessary to not check for the regex, we could disable that by passing in a boolean to `isImageURL`.

jcbrand (changes requested):

jcbrand:
@afuggini There are failing tests due to changes you made in this PR. Please fix.

src/utils/html.js
afuggini: Fixed.

src/converse-chatview.js
afuggini: Done.

jcbrand:
Thanks!

PR #2210, opened 2020-08-27
Removes query params from urls as set in config

afuggini:


src/converse-chatview.js
```diff
@@ -67,6 +67,7 @@ converse.plugins.add('converse-chatview', {
                 'emoji': true,
                 'spoiler': true
             },
+            'filter_url_query_params': []
```
jcbrand, line 70: When adding a new configuration option, please also document it here:

Options are added in alphabetical order.

jcbrand (changes requested):
Looks like there are failing tests due to this PR. Please fix.

spec/messages.js
```diff
@@ -896,6 +896,29 @@ describe("A Chat Message", function () {
...
+        mock.initConverse(
+            ['rosterGroupsFetched', 'chatBoxesFetched'], {},
+            async function (done, _converse) {
+
+        await mock.waitForRoster(_converse, 'current');
+        await mock.openControlBox(_converse);
+        const contact_jid = mock.cur_names[0].replace(/ /g,'.').toLowerCase() + '@montague.lit';
+        await mock.openChatBoxFor(_converse, contact_jid);
+        const view = _converse.api.chatviews.get(contact_jid);
+        _converse.api.settings.set('filter_url_query_params', ['utm_medium', 'utm_content', 's']);
```
jcbrand, line 909: Instead of manually calling `settings.set` here, you can just pass in default configuration settings to `mock.initConverse`. The 2nd parameter takes configuration settings.

jcbrand (commented):
Thanks, I've merged manually.

PR #2211, opened 2020-08-27
adds option to whitelist image domains

afuggini:
This requires changes in https://github.com/conversejs/converse.js/pull/2209

### DIFF
--- a/spec/messages.js
+++ b/spec/messages.js
@@ -974,6 +974,7 @@ describe("A Chat Message", function () {
         expect(msg.textContent.trim()).toEqual('hello world');
         expect(msg.querySelectorAll('img').length).toEqual(2);
 
+        // @XXX This test isn't really testing anything - needs revisiting
         // Non-https images aren't rendered
         message = base_url+"/logo/conversejs-filled.svg";
         expect(view.content.querySelectorAll('img').length).toBe(4);
@@ -985,7 +986,31 @@ describe("A Chat Message", function () {
         message = 'http://imgur.com/xxxxxxx';
         mock.sendMessage(view, message);
         await u.waitUntil(() => view.el.querySelectorAll('.chat-content .chat-image').length === 5, 1000);
-        expect(view.content.querySelectorAll('img').length).toBe(5);
+        expect(view.content.querySelectorAll('.chat-content .chat-image').length).toBe(5);
+
+        done();
+    }));
+
+    it("will render images from approved URLs only",
+        mock.initConverse(
+            ['rosterGroupsFetched', 'chatBoxesFetched'], {},
+            async function (done, _converse) {
+
+        await mock.waitForRoster(_converse, 'current');
+        const base_url = 'https://conversejs.org';
+        let message = 'http://wrongdomain.com/xxxxxxx.png';
+        const contact_jid = mock.cur_names[0].replace(/ /g,'.').toLowerCase() + '@montague.lit';
+        await mock.openChatBoxFor(_converse, contact_jid);
+        _converse.api.settings.set('show_images_inline', ['conversejs.org']);
+        const view = _converse.api.chatviews.get(contact_jid);
+        spyOn(view.model, 'sendMessage').and.callThrough();
+        mock.sendMessage(view, message);
+        message = base_url+"/logo/conversejs-filled.svg";
+        mock.sendMessage(view, message);
+        await u.waitUntil(() => view.el.querySelectorAll('.chat-content .chat-msg').length === 2, 1000);
+        await u.waitUntil(() => view.el.querySelectorAll('.chat-content .chat-image').length === 1, 1000)
+        expect(view.content.querySelectorAll('.chat-content .chat-image').length).toBe(1);
+
         done();
     }));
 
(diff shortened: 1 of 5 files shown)


src/templates/directives/body.js
```diff
@@ -120,9 +120,11 @@ function addHyperlinks (text, onImgLoad, onImgClick) {
...
             url_obj.start,
             url_obj.end,
-            show_images && u.isImageURL(url_text) ?
-                u.convertToImageTag(url_text, onImgLoad, onImgClick) :
-                u.convertUrlToHyperlink(url_text),
+            show_images
+                && (u.isImageURL(url_text) || u.isWhitelistedImageURL(url_text))
+                && u.isImageDomainAllowed(url_text)
+                    ? u.convertToImageTag(url_text, onImgLoad, onImgClick)
+                    : u.convertUrlToHyperlink(url_text),
```
jcbrand, line 127: `convertToImageTag` still has code that's no longer relevant due to the changes you've made here and in #2209.

Specifically the call to `u.isImageURL` and the check for `image_url_without_ext` are no longer relevant and can be removed.

src/converse-chatview.js
```diff
@@ -67,6 +67,8 @@ converse.plugins.add('converse-chatview', {
                 'emoji': true,
                 'spoiler': true
             },
+            'image_urls_regex': null,
+            'image_domains_whitelist': null
```
jcbrand, line 71: Instead of adding a new configuration setting, we can re-use [show_images_inline](https://github.com/conversejs/converse.js/blob/823134d7629bad2b82875964169340609c0ce2c9/docs/source/configuration.rst#L1681).

If `show_images_inline` is set to `true`, then all domains are whitelisted. If an array is passed in, then only those domains are whitelisted.

Please remember to update the configuration documentation: https://github.com/conversejs/converse.js/blob/master/docs/source/configuration.rst

And please also add a test that checks image domain whitelisting.

jcbrand (changes requested):

spec/messages.js
```diff
@@ -974,11 +974,43 @@ describe("A Chat Message", function () {
         expect(msg.textContent.trim()).toEqual('hello world');
         expect(msg.querySelectorAll('img').length).toEqual(2);
 
+        // @XXX This test isn't really testing anything - needs revisiting
         // Non-https images aren't rendered
         message = base_url+"/logo/conversejs-filled.svg";
         expect(view.content.querySelectorAll('img').length).toBe(4);
         mock.sendMessage(view, message);
         expect(view.content.querySelectorAll('img').length).toBe(4);
+
```
afuggini, line 983: The test block above doesn't test anything and needs to be revisited.
Simply put anything in the URL, https or not, and the test will still pass.
I tried to solve it but I'm not exactly sure what setting it is testing.

jcbrand: Looks like it also tests for something which isn't implemented anymore. I can't find the ticket where it was asked/decided that images without https shouldn't be rendered, so I've decided to just remove it for now.

spec/messages.js
```diff
@@ -974,11 +974,43 @@ describe("A Chat Message", function () {
...
+        _converse.api.settings.set('show_images_inline', ['conversejs.org']);
+        const view = _converse.api.chatviews.get(contact_jid);
+        spyOn(view.model, 'sendMessage').and.callThrough();
+        mock.sendMessage(view, message);
+        message = base_url+"/logo/conversejs-filled.svg";
+        mock.sendMessage(view, message);
+        await u.waitUntil(() => view.el.querySelectorAll('.chat-content .chat-msg').length === 2, 1000);
+        await u.waitUntil(() => view.el.querySelectorAll('.chat-content .chat-image').length === 1, 1000)
+        expect(view.content.querySelectorAll('.chat-content .chat-image').length).toBe(1);
+
```
jcbrand, line 1013: This test doesn't work correctly and in fact Converse does try to render an image for 'http://wrongdomain.com/xxxxxxx.png' because the default value for `show_images_inline` is `true` which effectively whitelists any domain.

One solution is to set `show_images_inline` to `['conversejs.org']` from the start.

Tests like this are hard to get right because of timing issues. In this case, you might have caught it if you waited until there is a message in the DOM from the first URL, before sending the second one.

I added a fix here:
https://github.com/conversejs/converse.js/commit/d83d0158e20ac1df7fac57634d3445cd15a0c3ed

src/utils/html.js
```diff
@@ -75,8 +75,25 @@ function checkFileTypes (types, url) {
...
+        ? regex.test(url)
+        : checkFileTypes(['.jpg', '.jpeg', '.png', '.gif', '.bmp', '.tiff', '.svg'], url);
+}
+u.isImageDomainAllowed = url => {
+    try {
+        const show_images_inline = api.settings.get('show_images_inline');
+        const is_domains_array = Array.isArray(show_images_inline);
+        if (!is_domains_array) return true;
+        const image_domain = getURI(url).domain();
+        return show_images_inline.includes(image_domain);
```
jcbrand, line 91: When putting code in a try/catch clause, please put as few lines as possible (ideally only one) there. It should only be the line that you expect to trigger exceptions under certain circumstances.

When there are many lines in a try/catch clause, then other errors in the code might get caught, which can have unexpected results and make debugging harder. It's also harder to see which line is the one where one should expect a potential exception to be raised.

I updated the code here:  https://github.com/conversejs/converse.js/commit/eb016a2aa4f5e9a323afa9e4039bc51c72087452

If this is wrong because the error is expected in a different line, please let me know.

jcbrand (commented):
Thanks @afuggini.

I've merged the branch manually because there were conflicts. Please see the last review comments I left.

PR #2213, opened 2020-08-30
add CustomElement to env

ChaosKid42:
I'd like to add `CustomElement` to `converse.env`.


pslxx:
image

only use headless build error

PR #2220, opened 2020-09-04
fix emojis in case use_system_emojis == false

ChaosKid42:
I found that emojis weren't rendered in case `use_system_emojis == false`. This PR fixes it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -30,6 +30,7 @@ Soon we'll deprecate the latter, so prepare now.
 - #2199: Fix BOSH session restore.
 - #2201: added html to converse.env
 - #2213: added CustomElement to converse.env
+- #2220: fix rendering of emojis in case `use_system_emojis == false` (again).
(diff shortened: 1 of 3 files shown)


jcbrand:
Thanks @ChaosKid42, looks good.

Could you please add a regressoin test to `spec/emojis.js` where Converse is configured with `use_system_emojis` set to `false`?

Similar to this one: https://github.com/conversejs/converse.js/blob/0594c480ffdeecc8baa0b36054e860654282920d/spec/emojis.js#L259

ChaosKid42:
> Thanks @ChaosKid42, looks good.
> 
> Could you please add a regressoin test to `spec/emojis.js` where Converse is configured with `use_system_emojis` set to `false`?
> 
> Similar to this one:
> 
> https://github.com/conversejs/converse.js/blob/0594c480ffdeecc8baa0b36054e860654282920d/spec/emojis.js#L259

@jcbrand : Done.

jcbrand:
Thanks!

PR #2222, opened 2020-09-04
Updates `onOwnPresence` method

afuggini:


jcbrand:
@afuggini Your commit message contains no explanation why this change was made or necessary and there are multiple tests that are failing which need to be fixed as well.

PR #2224, opened 2020-09-05
Prevent message actions dropdown from being closed after event retargeting when bubbl…

XaviFP:
…ing out of a shadowRoot.

When the click event bubbles out of the shadowRoot, the event gets re-targeted to the shadowRoot host. The event listener is expecting the same target so when the target changes the dropdown gets closed prematurely.
This change uses the original target instead of the re-targeted one allowing the dropdown to work as expected even when using converse in a shadowRoot.

### DIFF
--- a/src/components/dropdown.js
+++ b/src/components/dropdown.js
@@ -15,7 +15,7 @@ export class BaseDropdown extends CustomElement {
         this.button = this.dropdown.querySelector('button');
         this.dropdown.addEventListener('click', ev => this.toggleMenu(ev));
         this.dropdown.addEventListener('keyup', ev => this.handleKeyUp(ev));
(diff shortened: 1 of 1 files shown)


jcbrand:
Looks like [originalTarget](https://developer.mozilla.org/en-US/docs/Web/API/Event/originalTarget) is Mozilla specific and not standardized, so this doesn't look like a viable solution.

XaviFP:
Was looking for that, thank you!

I'm a bit concerned about if Converse was to be used inside a closed shadowRoot according to this: https://developer.mozilla.org/en-US/docs/Web/API/Event/composedPath
Seems in that case in won't propagate the event inside the shadow DOM and would start propagating just from outside above.
What are your thoughts?

PR #2234, opened 2020-09-17
Fix #2231 add sort by query

### LINKED ISSUE #2231: [autocomplete] prioritize the user for which the nickname STARTS with the text typed
**Describe the bug**
If you start typing the at-character ("@") followed by letters, the Converse UI will automatically try to auto-complete with the nickname of an occupant of the room.
For example, when I type "@ ber" I get this:

converse-bug

One thing that would be an improvement, is if Converse would prioritize the user for which the nickname STARTS with the text that I typed
You see in the example above that both Bernard's name as well as Viktor's match
If I would have pressed 'tab' after typing "@ ber" then it would have auto-completed with Viktor's name, instead of Bernard's
I think 99% of the times, a user intends to select the user who's nickname _starts_ with whatever is typed.

**To Reproduce**
type a text starting with "@"

**Expected behavior**
prioritize the user for which the nickname STARTS with the text that is typed

 - Desktop
 - Browser [Chrome]
 - Converse.js version [6.0.1]

bernard-ng:
- [*] Add a changelog entry for your change in `CHANGES.md`


guusdk:
This PR has lots of unrelated commits. Please remove them (rebase / squash), as they add nothing but noise (even if the grand total of all commits isn't adding any changes).

src/converse-autocomplete.js
```diff
@@ -22,12 +22,13 @@ export const FILTER_STARTSWITH = function (text, input) {
 };
 
 
-const SORT_BYLENGTH = function (a, b) {
-    if (a.length !== b.length) {
-        return a.length - b.length;
+const SORT_BY_QUERY = function (a, b) {
+    const query = a.query;
+    if (a.label.indexOf(query) === b.label.indexOf(query)) {
+        return 0
```
jcbrand, line 28: Instead of returning `0`, you can return `SORT_BY_LENGTH(a, b)` so that the longest relative match is shown higher.

src/converse-autocomplete.js
```diff
@@ -147,7 +148,7 @@ export class AutoComplete {
             'auto_first': false, // Should the first element be automatically selected?
             'data': a => a,
             'filter': FILTER_CONTAINS,
-            'sort': config.sort === false ? false : SORT_BYLENGTH,
+            'sort': config.sort === false ? false : SORT_BY_QUERY,
```
jcbrand, line 151: `SORT_BY_QUERY` is not a very descriptive name  for what the function is doing.

I instead suggest `SORT_BY_QUERY_POSITION`

jcbrand (changes requested):

spec/muc.js
```diff
@@ -2018,7 +2018,7 @@ describe("Groupchats", function () {
             spyOn(view.model, 'directInvite').and.callThrough();
 
             const input = modal.el.querySelector('#invitee_jids');
-            input.value = "Balt";
+            input.value = "Balth";
```
jcbrand, line 2021: @bernard-ng Why did you add an `h` here?

jcbrand (changes requested):
@bernard-ng Do you know how to squash your commits so that you have only one commit in the PR?

Can you also please add a test case for this improvement?

src/converse-autocomplete.js
```diff
@@ -22,13 +22,24 @@ export const FILTER_STARTSWITH = function (text, input) {
...
 
+const SORT_BY_QUERY_POSITION = function (a, b) {
+    const query = a.query;
+    const x = a.label.toLowerCase().indexOf(query);
+    const y = b.label.toLowerCase().indexOf(query);
+
+    if (x === y) {
+        return SORT_BY_LENGTH(a, b);
+    }
+    return (x === -1 ? Infinity : x) < (y === -1 ? Infinity : y) ? -1 : 1
```
jcbrand, line 40: Shouldn't `x` and `y` never be `-1`?

If this sorting function is called for `x` and `y`, then it means they already match the query string, so `indexOf` should never return `-1`. Or is there something I'm missing?

src/converse-autocomplete.js
```diff
@@ -22,13 +22,24 @@ export const FILTER_STARTSWITH = function (text, input) {
...
     if (a.length !== b.length) {
         return a.length - b.length;
     }
     return a < b? -1 : 1;
 };
 
+const SORT_BY_QUERY_POSITION = function (a, b) {
+    const query = a.query;
+    const x = a.label.toLowerCase().indexOf(query);
+    const y = b.label.toLowerCase().indexOf(query);
```
jcbrand, line 35: Perhaps necessary to add `toLowerCase` to `query` as well?

jcbrand (changes requested):

src/converse-autocomplete.js
bernard-ng: Sometimes this happens when a "list" is directly passed to an instance of autocomplete as in the following case: 

https://github.com/conversejs/converse.js/blob/5ea2ccfc17ebb0360dcdde80321cb2e1a8b1fc20/src/modals/muc-invite.js#L31-L35

I don't know why, but checking that the index is different to -1, allowed me to correct this.

jcbrand: Ok, then I think your change makes sense 👍

jcbrand:
Thanks @bernard-ng!

bernard-ng:
> Thanks @bernard-ng!

Anytime

PR #2237, opened 2020-09-18
fix i18n of label group name

ChaosKid42:
The label group names in the roster are not translated. This PR aims to fix that.


jcbrand:
Thanks!

PR #2248, opened 2020-09-25
Bugfix. Promise.resolve is not a constructor

Zash:
Fixes

> FATAL: error: undefined .../converse.min.js:241 - TypeError: Promise.resolve is not a constructor

as noticed in the console

Small fix, changelog entry and test needed?

### DIFF
--- a/src/headless/converse-core.js
+++ b/src/headless/converse-core.js
@@ -942,7 +942,7 @@ export const api = _converse.api = {
             }
         } else {
             _converse.connection.sendIQ(stanza);
-            promise = new Promise.resolve();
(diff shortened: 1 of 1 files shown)


guusdk:
Oh, I'm wondering what functional issue this fixes! If I understand javascript exception handling, the promise was never returned, and the api 'send' trigger was not either.

Zash:
It fixes that there's an error in the console, no idea what else this affects.

jcbrand:
Thanks!

@Guus this was a syntactic error, and not really a logical error.

Issue #2168, opened 2020-07-30
FIle upload images show up twice


codecleric:
**Describe the bug**
When uploading and image using the toolbar paper-clip icon, the image shows up in the chat twice. 
The first one is a <converse-chat-message-body> tag, and the second is <div class="chat-msg__media">, both container img tags pointing to the newly uploaded image.

**To Reproduce**
1. Go to a private chat or group chat
2. Click on the toolbar paper-clip icon and select an image file.
3. Select 'open' on file selection dialog.
4. See image show up twice.

**Expected behavior**
I expect only one image to show up in the chat.

**Screenshots**
![converse_file-upload-error](https://user-images.githubusercontent.com/14131975/88898566-24f89c00-d20a-11ea-9fe6-10ff86790c5e.png)

**Environment (please complete the following information):**
 - Prosody 0.11.4
 - Firefox 78.0.2 (64-bit)
 - Converse.js version 6.0.1dev

**Additional context**
This is version of converse.js that I build and loaded the assets into the docker container.

jcbrand:
Fixed in #2254

PR #2254, opened 2020-09-29
Fix issue #2168 (Images appear twice in chat)

ChaosKid42:
This fixes https://github.com/conversejs/converse.js/issues/2168

The clients I use (converse.js, Conversations) transmit the Image file URL twice: once inside the message text and once inside the OOB tag. Therefore it is required to no render the image URL as an img in the OOB tag.


fabiomontefuscolo (approved):
## Before

Screenshot from 2020-10-01 14-50-45

## After

Screenshot from 2020-10-01 14-57-43

jcbrand:
Thanks, merged manually.

PR #2259, opened 2020-10-01
Update info messages setting

XaviFP:

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -1379,13 +1379,53 @@ Example:
 
     muc_roomid_policy_hint: '<br><b>Policy for groupchat id:</b><br>- between 5 and 40 characters,<br>- lowercase from a to z (no special characters) or<br>- digits or<br>- dots (.) or<br>- underlines (_) or<br>- hyphens (-),<br>- no spaces<br>',
 
-muc_show_join_leave
--------------------
+muc_show_info_messages
+----------------------
 
-* Default; ``true``
+* Default: List composed of MUC status codes, role changes, join and leave events
+and affiliation changes. The values of converse.MUC_INFO_CODES below are joined to 
+build the default list:
 
-Determines whether Converse will show info messages inside a chatroom
-whenever a user joins or leaves it.
+.. code-block:: javascript
+    converse.MUC_AFFILIATION_CHANGES_LIST = ['owner', 'admin', 'member', 'exowner', 'exadmin', 'exmember', 'exoutcast']
+    converse.MUC_ROLE_CHANGES_LIST = ['op', 'deop', 'voice', 'mute'];
+    converse.MUC_TRAFFIC_STATES_LIST = ['entered', 'exited'];
+
+    converse.MUC_INFO_CODES = {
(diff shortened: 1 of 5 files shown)


src/headless/converse-muc.js
```diff
@@ -130,8 +139,16 @@ converse.plugins.add('converse-muc', {
...
-            'muc_show_join_leave': true,
-            'muc_show_logs_before_join': false
+            // 'muc_show_join_leave': true,
+            'info_messages': [
+                100, 102, 103, 104, 110, 170,
+                171, 172, 173, 174, 201, 210,
+                301, 303, 307, 321, 322, 332, 333,
+                'affiliation_changes',
+                'join_leave_events',
+                'role_changes',
```
jcbrand, line 149: Having a mixture of strings and codes seems weird to me.

I think we can store only codes and then use const/variables to improve DX.

For example, what about something like this:

```
info_messages: [
    ... MUC_INFO_CODES.affiliation_changes,
    ... MUC_INFO_CODES.visibility_changes,
    // etc.
]
```

We can then assign `MUC_INFO_CODES` to the `converse` object so that when users call `converse.initialize` they can use the user-friendly names/categories instead of passing in the codes themselves.

src/headless/converse-muc.js
```diff
@@ -25,14 +25,17 @@ converse.MUC_ROLE_CHANGES = ['op', 'deop', 'voice', 'mute'];
...
     'visibility_changes': [100, 102, 103, 172, 173, 174],
     'self': [110],
     'non_privacy_changes': [104, 201],
     'muc_logging_changes': [170, 171],
     'nickname_changes': [210, 303],
     'disconnect_messages': [301, 307, 321, 322, 332, 333],
-}
+    'affiliation_changes': 'affiliation_changes',
+    'join_leave_events': 'join_leave_events',
+    'role_changes': 'role_changes',
```
jcbrand, line 37: I was more thinking along the lines of:

`'role_changes': [...converse.MUC_ROLE_CHANGES]`

instead of:

`'role_changes': 'role_changes'`

What do you think of that?

XaviFP: @jcbrand I went ahead with your suggestion and added values for the three categories. Now visibility can be controlled case by case :)

jcbrand:
@XaviFP: Can you please fix the conflicts and squash your commits? Then I'll merge this PR

docs/source/configuration.rst
```diff
@@ -1318,13 +1318,53 @@ Example:
 
     muc_roomid_policy_hint: '<br><b>Policy for groupchat id:</b><br>- between 5 and 40 characters,<br>- lowercase from a to z (no special characters) or<br>- digits or<br>- dots (.) or<br>- underlines (_) or<br>- hyphens (-),<br>- no spaces<br>',
 
-muc_show_join_leave
+muc_show_info_messages
 -------------------
```
jcbrand, line 1323: The line under the heading must be the same length as the heading itself.

XaviFP:
@jcbrand, it's ready :)

Issue #1829, opened 2020-01-10
OMEMO lock icon - change colour if disabled


licaon-kter:
...as the difference is just some pixels, while making it grey or red makes it clear that encryption is not enabled.

licaon-kter:
With `concord` theme the private MUC icons (smiley/lock/etc) are grey.

jcbrand:
Fixed in #2260

PR #2260, opened 2020-10-01
#1829 add color for different omemo state

bernard-ng:
**active** : #3DAC44
**inactive** : #F75681

conversejs-screenshots (1)

### DIFF
--- a/src/converse-omemo.js
+++ b/src/converse-omemo.js
@@ -502,6 +502,7 @@ function getOMEMOToolbarButton (toolbar_el, buttons) {
                 @click=${toggleOMEMO}>
         <converse-icon class="fa ${model.get('omemo_active') ? `fa-lock` : `fa-unlock`}"
                     path-prefix="${api.settings.get('assets_path')}" size="1em"
+                    color="${model.get('omemo_active') ? `#3DAC44` : `#F75681`}"
         ></converse-icon>
         </button>`
     );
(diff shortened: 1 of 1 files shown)


src/converse-omemo.js
```diff
@@ -502,6 +502,7 @@ function getOMEMOToolbarButton (toolbar_el, buttons) {
                 @click=${toggleOMEMO}>
         <converse-icon class="fa ${model.get('omemo_active') ? `fa-lock` : `fa-unlock`}"
                     path-prefix="${api.settings.get('assets_path')}" size="1em"
+                    color="${model.get('omemo_active') ? `#3DAC44` : `#F75681`}"
```
jcbrand, line 505: Can we instead use CSS classes and custom CSS properties `--var()` to set colors, instead of hardcoding them like this?

src/converse-omemo.js
```diff
@@ -502,6 +502,7 @@ function getOMEMOToolbarButton (toolbar_el, buttons) {
                 @click=${toggleOMEMO}>
         <converse-icon class="fa ${model.get('omemo_active') ? `fa-lock` : `fa-unlock`}"
                     path-prefix="${api.settings.get('assets_path')}" size="1em"
+                    color="${model.get('omemo_active') ? `var(--info-color)` : `var(--error-color)`}"
```
jcbrand, line 505: @bernard-ng Did you check whether the color updates dynamically? I.e. when you click the button, does it switch from green to red?

bernard-ng: Yes it does

jcbrand:
Thanks @bernard-ng

PR #2266, opened 2020-10-05
check against own jid when detecting forged mam messages

ChaosKid42:
I see many `Invalid Stanza: Forged MAM message from ...` error messages on the console when fetching messages from the archive (MAM). Fetched messages are then discarded.

This PR fixes it.


jcbrand:
Thank you @ChaosKid42

Could you also please write a test for this? To make sure we avoid a similar regression in the future.

ChaosKid42:
> Thank you @ChaosKid42
> 
> Could you also please write a test for this? To make sure we avoid a similar regression in the future.

@jcbrand Done.

spec/mam.js
```diff
@@ -265,6 +265,60 @@ describe("Message Archive Management", function () {
                 done();
             }));
 
+            it("is not discarded if it comes from the right sender",
+                mock.initConverse(
```
jcbrand, line 269: Can you please add a forged message to this test-case and then check that it gets rejected?

That way we cover both cases.

I also think the test can then be renamed to `is discarded if it's forged`, since that's the exception and being not discarded is the norm.

ChaosKid42: The other case is already covered [here](https://github.com/conversejs/converse.js/blob/master/spec/mam.js#L213). That's why I didn't add it.

PR #2267, opened 2020-10-06
Allow mention autocomplete to trigger right after a new line

XaviFP:

### DIFF
--- a/src/headless/utils/core.js
+++ b/src/headless/utils/core.js
@@ -412,7 +412,7 @@ u.getCurrentWord = function (input, index, delineator) {
     if (!index) {
         index = input.selectionEnd || undefined;
     }
-    let [word] = input.value.slice(0, index).split('\n').slice(-1)[0].split(' ').slice(-1);
+    let [word] = input.value.slice(0, index).split(/\s/).slice(-1);
     if (delineator) {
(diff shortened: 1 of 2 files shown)


src/headless/utils/core.js
```diff
@@ -412,7 +412,7 @@ u.getCurrentWord = function (input, index, delineator) {
     if (!index) {
         index = input.selectionEnd || undefined;
     }
-    let [word] = input.value.slice(0, index).split(' ').slice(-1);
+    let [word] = input.value.slice(0, index).split('\n').slice(-1)[0].split(' ').slice(-1);
```
jcbrand, line 415: [According to the docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split), you can pass a regex to String.prototype.split, which means that you can just replace `'\n'` with the regex `/\s/`, instead of calling `split` (and `slice`) twice.

`\s` is metacharacter that matches whitespace, which can be:

- A space character
- A tab character
- A carriage return character
- A new line character
- A vertical tab character
- A form feed character

So it has the advantage of matching other whitespace characters as well.

XaviFP: Oh! Thanks for that one, didn't know :-) Just changed it :+1:

PR #2275, opened 2020-10-07
Allow selected characters to precede a mention

XaviFP:

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 configuration settings should now be accessed via `_converse.api.settings.get` and not directly on the `_converse` object.
 Soon we'll deprecate the latter, so prepare now.
 
+- #2275: Allow selected characters to precede a mention
 - #2231: add sort_by_query and remove sort_by_length
 - #1313: Stylistic improvements to the send button
 - #1481: MUC OMEMO: Error No record for device
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,8 +6,7 @@
 configuration settings should now be accessed via `_converse.api.settings.get` and not directly on the `_converse` object.
 Soon we'll deprecate the latter, so prepare now.
 
-- #2275: Allow selected characters to precede a mention
-- #2231: add sort_by_query and remove sort_by_length
+- #515 Add support for XEP-0050 Ad-Hoc commands
 - #1313: Stylistic improvements to the send button
 - #1481: MUC OMEMO: Error No record for device
 - #1490: Busy-loop when fetching registration form fails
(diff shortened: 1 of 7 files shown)


CHANGES.md
```diff
@@ -6,6 +6,7 @@
 configuration settings should now be accessed via `_converse.api.settings.get` and not directly on the `_converse` object.
 Soon we'll deprecate the latter, so prepare now.
 
+- #2275: Allow selected characters to precede a mention
 - #2231: add sort_by_query and remove sort_by_length
```
jcbrand, line 10: Please add tickets in the right order in the changelog (I know not all contributors always do that).

src/headless/converse-muc.js
```diff
@@ -973,7 +973,7 @@ converse.plugins.add('converse-muc', {
             getAllKnownNicknamesRegex () {
                 const longNickString = this.getAllKnownNicknames().join('|');
                 const escapedLongNickString = p.escapeRegexString(longNickString)
-                return RegExp(`(?:\\s|^)@(${escapedLongNickString})(?![\\w@-])`, 'ig');
+                return RegExp(`(?:\\s|^).?@(${escapedLongNickString})(?![\\w@-])`, 'ig');
```
jcbrand, line 976: Instead of matching `.?` shouldn't it be matching `opening_mention_characters`?

docs/source/configuration.rst
```diff
@@ -1416,6 +1416,23 @@ omemo_default
...
+the specified characters is preceding the ``@`` in a mention.
+
+If no character should be allowed before the ``@`` in a mention other than
+whitespace characters, this setting should be set to empty list: ``[]``
+
+Example:
+
+Typing ``(@`` triggers the mention autocompletion component but typing ``>@`` doesn't.
+
+
```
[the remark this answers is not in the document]
XaviFP: I did it this way to allow more flexibility configuring Converse. There might be groups of people who use some character to add meaning to a name, or maybe those people use different characters than the ones we use for example these [characters](https://en.wikipedia.org/wiki/List_of_Japanese_typographic_symbols#Brackets_and_quotation_marks), but mainly it was for the sake of flexibility.
I understand your point of view regarding the presence of too many moving parts and the burden it carries.
If we remove the setting as such and leave it as an internal set of values, there are changes to apply as you already mention below in another comment.
As for the boundaries, right now you can type any character after a mention already, so the issue as I see it is to be able to trigger and interact with the mention autocompletion component even if the mention is preceded by one of these selected characters.
But yes, I do like `mention_boundaries` and think it matches closely with what can be achieved with the functionality.

docs/source/configuration.rst
```diff
@@ -1416,6 +1416,23 @@ omemo_default
 
 Use OMEMO encryption by default when the chat supports it.
 
+opening_mention_characters
+__________________________
+
+* Default: ``['"', '(', '<', '#', '!', '\\', '/', '+', '~', '[', '{', '^']``
```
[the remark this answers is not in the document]
XaviFP: It's not that I deliberately left it out, I just added some characters that could potentially make sense to add.
Do you have more characters in mind that could be added? And related to the variable name above, would the closing counterparts of the current characters be added as well?

src/converse-autocomplete.js
```diff
@@ -365,7 +366,9 @@ export class AutoComplete {
             this.auto_completing = true;
         } else if (ev.key === "Backspace") {
             const word = u.getCurrentWord(ev.target, ev.target.selectionEnd-1);
-            if (this.ac_triggers.includes(word[0])) {
+            const opening_mention_chars = api.settings.get('opening_mention_characters');
+            if (this.ac_triggers.includes(word[0]) ||
+            (opening_mention_chars.includes(word[0]) && this.ac_triggers.includes(word[1]))) {
```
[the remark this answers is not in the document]
XaviFP: I fail to remember the reason for this but I recall not wanting to modify `getCurrentWord`. I'll update it to leave out the initial mention boundary if any.

CHANGES.md
XaviFP: Yes, no worries. I saw the flow of the document from new to old and instinctively pushed that on the top ^^'

src/headless/converse-muc.js
XaviFP: Definitely. Used that as a test and forgot to update afterwards. Thanks!

src/headless/utils/parse-helpers.js
```diff
@@ -7,7 +7,7 @@
 const helpers = {};
 
 // Captures all mentions, but includes a space before the @
-helpers.mention_regex = /(?:\s|^)([@][\w_-]+(?:\.\w+)*)/ig;
+helpers.mention_regex = /(?:\s|^)(.?[@][\w_-]+(?:\.\w+)*)/gi;
```
jcbrand, line 10: Looks like this `.?` should also be removed, right?

docs/source/configuration.rst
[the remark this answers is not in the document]
jcbrand: > There might be groups of people who use some character to add meaning to a name, or maybe those people use different characters than the ones we use

Then they can create a plugin that provides the behavior that they need, or they can create a ticket/MR for the changes they need.

Right now this is all hypothetical, and I'd rather avoid adding new configuration settings for speculative use-cases.

Please remove the configuration setting.

jcbrand (changes requested):

XaviFP:
Just updated with the changes :)

jcbrand:
@XaviFP: Is this PR still WIP? Otherwise you can remove that from the title please.

XaviFP:
@jcbrand Done :) Thanks!

jcbrand:
Hey @XaviFP 

I held out on merging this PR because I wanted to propose a slightly different way of solving the problem.

Instead of using a whitelist of allowable characters, we can use a regular expression with unicode categories to allow a mention if (for example) it's preceded by punctuation.

Here's an article explaining how to use unicode properties in a regular expression:
https://javascript.info/regexp-unicode#unicode-properties-p

By for example using something like `/\p{P}/u` instead of the whitelist, we can match *any* punctuation.

What do you think?

XaviFP:
@jcbrand I like that as an extension to a list of specific characters as it allows for the inclusion of whole sets of characters in a simple way, but wouldn't it complicate the general case? For example if one wants to specify just a couple of Unicode characters from one or two of the punctuation sub-categories, and some more characters from other categories or sub-categories.

For me it just depends on the method that simplifies the most based on the default set of characters Converse uses. If it was for a setting, I would go for the regular list as the set of designated characters is up for whoever is overriding the default to select. But not being the case I would choose whatever is simpler or more coherent with the rest of the code-base.

jcbrand:
Merged manually and then updated to use a regex.

Issue #2274, opened 2020-10-06
Concord theme: Invisible send button


Zash:
**Describe the bug**
When using the 'concord' theme, the button to send messages is a white icon on white background.

**To Reproduce**
1. Configure concord theme
2. Open a chat
3. Look for send button

**Expected behavior**
Button should match the colors of the theme, ~~i.e. gray on white~~, gray and white (which should be which of these left to the concord theme designer). Compare the `<<` icon in attached screenshots.

**Screenshots**

![image](https://user-images.githubusercontent.com/197474/95246869-66fb1b00-0815-11eb-84c1-fb2a28ab77b6.png)

![image](https://user-images.githubusercontent.com/197474/95246901-72e6dd00-0815-11eb-8bb5-c7e216008fa8.png)

**Environment (please complete the following information):**
 - Debian 10 desktop
 - Firefox ?? ESR
 - Converse.js version almost 7 f68eb65cd79b4bd5edd67d0d87c5f01e22f653df (might be a few commits earlier actually)

**Additional context**
Spotted a `color: #fff !important` in the browser DOM inspector tool which appears to come from:

https://github.com/conversejs/converse.js/blob/f68eb65cd79b4bd5edd67d0d87c5f01e22f653df/sass/_core.scss#L392-L396

PR #2279, opened 2020-10-08
#2274 add background to send-button

bernard-ng:
conversejs-screenshots (3)

### DIFF
--- a/sass/_toolbar.scss
+++ b/sass/_toolbar.scss
@@ -14,7 +14,7 @@
 
     .chatroom {
         .send-button {
-            background-color: var(--chatroom-head-bg-color);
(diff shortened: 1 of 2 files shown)


sass/_toolbar.scss
```diff
@@ -14,7 +14,7 @@
 
     .chatroom {
         .send-button {
-            background-color: var(--chatroom-head-bg-color);
+            background-color: var(--chatroom-send-button-bg);
```
jcbrand, line 17: I see there is already a CSS propery `--muc-toolbar-btn-color`.

Looks like you can just use that, instead of adding a new one.

PR #2286, opened 2020-10-12
WIP: XEP-0393 Message Styling

XaviFP:

### DIFF
--- a/src/headless/utils/parse-helpers.js
+++ b/src/headless/utils/parse-helpers.js
@@ -40,27 +40,8 @@ const reduceReferences = ([text, refs], ref, index) => {
 
 helpers.reduceTextFromReferences = (text, refs) => refs.reduce(reduceReferences, [text, []]);
 
-helpers.getLineSpans = (line, offset = 0, start = 0) => {
-    const spans = { open: [], closed: [] };
-    let char;
-    for (let index = start; index < line.length; index++) {
-        char = line[index];
-        if (isStylingDirective(char)) {
-            if (!hasOpenedSpan(char, spans)) {
-                if (isOpeningDirective(char, index, line, start)) {
-                    openSpan(char, index, spans);
-                }
-            } else {
-                if (isClosingDirective(index, line)) {
-                    closeSpan(char, index, offset, spans);
-                }
-            }
-        }
-    }
-    return [...spans['closed']];
-};
 
-helpers.getStylingReferences = (message) => {
+export function getStylingReferences (message) {
     const line_space_offsets = {};
     const line_offsets = {};
     const lines = message.split("\n");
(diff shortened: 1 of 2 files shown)


licaon-kter:
Pics?

jcbrand (commented):
Thanks @XaviFP 

I prefer to avoid implementing and using nested references for the `MessageText` type since this creates unnecessary extra complexity IMO.

Instead, the message text can be parsed and transformed into HTML markup text (i.e. without interleaving it with lit-html templates).

That HTML text can then be rendered in a single lit-html template via the `unsafeHTML` directive.

I've created a branch and WIP PR that implements this idea.

Take a look here:
https://github.com/conversejs/converse.js/pull/2296

It's not completely done, but it shows how the idea works and there are some initial tests already.

XaviFP:
@jcbrand That was my first approach. It's more or less doable for one-line styling, but it gets too complex when blocks come in.
Styling spans are parsed horizontally whereas blocks need to be parsed vertically, and, in the case of a quote block its content needs to be parsed and so on.
If it was not done recursively it would make the task of parsing way more complex. The number of nesting levels and the content of each block is arbitrary, so the logic to keep the state increases substantially. 
In this case, recursion allows to traverse each character of the message just once (as noted in `Block` section of [glossary](https://xmpp.org/extensions/xep-0393.html#glossary) and the [implementation notes](https://xmpp.org/extensions/xep-0393.html#impl)).

When I started, I have to admit that I had second thoughts about what the XEP says and recommends. 
I took a look at what some other clients were doing and realized they do it linearly, but none of them implements multi-level quoting or blocks whatsoever.
If you have in mind something simple for one-line styling or just one level of quoting it is doable with a linear approach, but it gets too tricky for blocks.

Also it makes more sense logically to have an element recursively nested inside another as it represents it's real position.
I am personally no fan of recursion, but I must admit that in this case it just makes things easier.
If the issue is to render nested TemplateResult objects, I can add a flattening function and substitute the styling templates for text(as you mention), and then render the flat list of text and TemplateResult objects with the unsafeHTML directive.

I've taken this screenshot with the current code and some CSS just so that you can get an image of what I mean(don't mind the CSS):
Converse_message_styling

XaviFP:
I've created an example branch with the flattening as an example [here](https://github.com/conversejs/converse.js/pull/2314/files#diff-985102a70641f47f5ebe4b59f9edc8f7267381ccbeab7ab1dce73f0fd4d2fac6R122)

I don't have an alternative in mind different than that one right now. 
Both solutions allow merging elements of different types when they are nested and the flattening example allows for flattening to a single array prior to rendering. 
I think if we did it in several goes, searching what to replace in an already and partially rendered message would make things more complex. 
I would still go for this current PR instead of the flattening one. But I'm of course up for alternatives and the styling parsing machinery just returns a flat list of styling references, so it's just a matter of how to store them.

jcbrand:
> @jcbrand That was my first approach. It's more or less doable for one-line styling, but it gets too complex when blocks come in.
> Styling spans are parsed horizontally whereas blocks need to be parsed vertically, and, in the case of a quote block its content needs to be parsed and so on.

I think the distinction between parsing vertically and horizontally is unnecessary and likely due to the fact that you're parsing line-by-line instead of character by character.

When you consider that a line break is just another character, then you can greatly simplify the parsing by simply doing it per character, which is what I'm doing in my branch.

> If it was not done recursively it would make the task of parsing way more complex.

The parsing in my branch *is* done recursively. I just don't create a recursive data structure for the references, which isn't necessary and which creates unnecessary complexity.

I'm fairly sure that arbitrarily nested block elements are not a problem for the approach that I'm advocating and that the resulting code will still be less and simpler.

Have you looked at the branch I worked in? Do you see the fundamental difference between what's being done there and what you did in your branch? 

I keep the recursion in one relatively small function: https://github.com/conversejs/converse.js/pull/2296/files#diff-00d9b44530a52203588f4488ae86920c6a8a15b7adb13bfb0297260298dd507cR66

And there's no need for the complexity of creating and maintaining a nested data-structure for rendering rich messages.

jcbrand:
Closed in favour of #2296

PR #2297, opened 2020-10-16
Muppdev

muppeth:
Change Aspect ratio to slice instead of meet. This fills viewbox so all avatars are same size.


jcbrand:
Thanks for the PR.

However, I'm not convinced that this is better, since oddly shaped avatars might have parts cropped off.

muppeth:
At this moment most of avatars (unless perfectly square) have either white lines around, or in case of round avatars, no full circle. 
I don't know if this is the best approach to the problem but so far it has been the only way to make the avatars appear as same size for all. It does mean some of them are cropped, but overall it gives (imo) more esthetic view as all avatars appears to be same size.
Screenshot from 2020-11-30 00-57-35
Screenshot from 2020-11-30 00-57-58

PR #2306, opened 2020-10-20
fix #2304 - Custom emojis (stickers) images not shown

### LINKED ISSUE #2304: custom emojis (stickers) images not shown
**Describe the bug**
I am in the process of upgrading a few plugins from 6.x to 7.x and while working on the [stickers plugin](https://github.com/conversejs/community-plugins/tree/converse.6.x/packages/stickers), discovered that the images for custom emojis are not showing because [getShortnameReferences](https://github.com/conversejs/converse.js/blob/c9ba42d0a02d0b1c81ae42becc4f2e5588c01a9a/src/headless/converse-emoji.js#L150) uses a scoped variable shortnames_regex that can not be accessed by a plugin.

**To Reproduce**
image

**Expected behavior**
I am expecting compatible behavior with 6.x. By using a function instead of a scoped variable, I am amble to make it work

image

**Additional context**
I refactored the getShortnameReferences function and added a getShortnameRegex

```
function getShortnameRegex()
{
  const getShortNames = () => converse.emojis.shortnames.map(s => s.replace(/[+]/g, "\\$&")).join('|');
  return new RegExp(getShortNames(), "gi");
}

function getShortnameReferences(text) {
  if (!converse.emojis.initialized) {
    throw new Error('getShortnameReferences called before emojis are initialized. ' + 'To avoid this problem, first await the converse.emojis.initilaized_promise.');
  }

  const shortnames_regex = getShortnameRegex();
  const references = [...text.matchAll(shortnames_regex)].filter(ref => ref[0].length > 0);
  return references.map(ref => {
    const cp = converse.emojis.by_sn[ref[0]].cp;
    return {
      cp,
      'begin': ref.index,
      'end': ref.index + ref[0].length,
      'shortname': ref[0],
      'emoji': cp ? converse_emoji_convert(cp) : null
    };
  });
}
```

deleolajide:
This fix can only be tested in a plugin. The plugin and PR is currently running at https://conversejs.github.io/community-plugins

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,6 +36,7 @@ Soon we'll deprecate the latter, so prepare now.
 - #2220: fix rendering of emojis in case `use_system_emojis == false` (again).
 - #2092: fixes room list update loop when having the `locked_muc_domain` truthy or `'hidden'`
 - #2285: Rename config option `muc_hats_from_vcard` to [muc_hats](https://conversejs.org/docs/html/configuration.html#muc-hats). Now accepts a list instead of a boolean and allows for more flexible choices regarding user badges.
+- #2304: Custom emojis (stickers) images not shown 
 - The `trusted` configuration setting has been removed in favor of two new settings:
     [allow_user_trust_override](https://conversejs.org/docs/html/configuration.html#allow-user-trust-override)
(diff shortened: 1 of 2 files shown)


jcbrand:
Thanks @deleolajide, nice to see you working again.

deleolajide:
Thank you @jcbrand for all the well wishes. It is good to be back :-)

Issue #2277, opened 2020-10-08
Anonymous instance: no error when joining a MUC fails


ge0rg:
Scenario: I'm using an anonymous converse.js instance with an auto-join room.

When I open the converse page, it greets me with "Please choose your nickname".

Then I enter a nickname, that is already used in the room and press "Enter groupchat"

The screen flickers for a second, and I'm back at "Please choose your nickname".

There is nothing in the JS debug console.

**Expected result:** Converse.js showing me an approrpiate error message.

jcbrand:
Do you have any idea why joining the MUC failed? e.g. server logs or anything?

ge0rg:
Yes, there is already a user in that room with that nickname.

Zash:
The error stanza should look like [this example in XEP-0045 7.2.8 Nickname Conflict](https://xmpp.org/extensions/xep-0045.html#enter-conflict).

jcbrand:
@bernard-ng Looks like there already is a test for this use-case: https://github.com/conversejs/converse.js/blob/5341a1ea498dd870a29235bc05a2915d4e77a6c8/spec/muc.js#L4170

The test however doesn't check that an error message is displayed to the user, to inform them that the nickname is already taken.

You can therefore update the test and the code accordingly.

jcbrand:
@bernard-ng: Here's an example of a test that checks that an error message displays in the nickname, you can probably learn/copy from that in order to test this bug:
https://github.com/conversejs/converse.js/blob/5341a1ea498dd870a29235bc05a2915d4e77a6c8/spec/muc.js#L4040

bernard-ng:
@jcbrand I think I'm stuck now, I can't return the validation error to the user, should I send a "DisconnectionMessage" instead?

I noticed that all the errors are in this object, for the nickname conflict, it's error 409, should I add it too? 

https://github.com/conversejs/converse.js/blob/646542602fbe65819a4e933744e28f1ae4f9b9a4/src/headless/converse-muc.js#L144-L195

bernard-ng:
by analyzing the code that is responsible for displaying the password field

https://github.com/conversejs/converse.js/blob/646542602fbe65819a4e933744e28f1ae4f9b9a4/src/converse-muc-views.js#L1168-L1187

I realize that the validation error is passed to the form and is displayed in the following template

[...]

PR #2312, opened 2020-10-23
WIP #2277 show error in nickname template

bernard-ng:
as you can see the error message is not sent to the view here : 
https://github.com/conversejs/converse.js/blob/646542602fbe65819a4e933744e28f1ae4f9b9a4/src/converse-muc-views.js#L1105-L1124

as here : 

https://github.com/conversejs/converse.js/blob/646542602fbe65819a4e933744e28f1ae4f9b9a4/src/converse-muc-views.js#L1168-L1187

### DIFF
--- a/src/templates/muc_nickname_form.js
+++ b/src/templates/muc_nickname_form.js
@@ -15,8 +15,13 @@ export default (o) => {
             <form class="converse-form chatroom-form converse-centered-form">
                 <fieldset class="form-group">
                     <label>${i18n_heading}</label>
-                    <input type="text" required="required" name="nick" value="${o.nick || ''}"
-                        class="form-control" placeholder="${i18n_nickname}"/>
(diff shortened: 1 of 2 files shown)


src/templates/muc_nickname_form.js
```diff
@@ -15,8 +15,13 @@ export default (o) => {
             <form class="converse-form chatroom-form converse-centered-form">
                 <fieldset class="form-group">
                     <label>${i18n_heading}</label>
-                    <input type="text" required="required" name="nick" value="${o.nick || ''}"
-                        class="form-control" placeholder="${i18n_nickname}"/>
+                    <p class="validation-message">${o.validation_message}</p>
```
jcbrand, line 18: The validation message gets saved as `nickname_validation_message`, not `validation_message`.

jcbrand (changes requested):

PR #2322, opened 2020-10-30
fix #2321

### LINKED ISSUE #2321: When Converse runs inside a browser extension, enable browser sync and local storage for persistent storage
Browser extensions have access to chrome.storage which offers local and sync drivers that offer more features. chrome.storage.local can perform as fast as localStorage and configured with no memory limits, while chrome.storage.sync has more limitations and restrictions, but offers storage that is synchronized across multiple devices and O/S platforms.

When converse runs inside a browser extension, we should use the [localforage browser extension storage driver](https://github.com/esphen/localforage-webExtensionStorage-driver) to provide additional persistent storage drivers

deleolajide:
This fix does not a new test, however id does need to pass a regression test of all existing tests


jcbrand:
Thanks!

PR #2323, opened 2020-10-30
fix #1250

### LINKED ISSUE #1250: Change favicon color when unread messages
In fullscreen mode, it would be good to be able to change the color of the favicon to red  when you have unread messages. This would let people know just looking at the converse tab favicon that they have unread messages if they missed the notification.

If changing favicon color is not possible, maybe changing the title of the page would be a good alternative (but in this case, when the tab is pinned in Chrome, we won't be able to see it).

deleolajide:
Moving feature form plugin to converse core code

image

### DIFF
--- a/src/headless/converse-chatboxes.js
+++ b/src/headless/converse-chatboxes.js
@@ -32,7 +32,8 @@ converse.plugins.add('converse-chatboxes', {
             'privateChatsAutoJoined'
         ]);
 
-        let msg_counter = 0, favicon = new Favico({type : 'circle', position: 'up', animation: 'none'});
+        let msg_counter = 0;
+        const favicon = new Favico({type : 'circle', position: 'up', animation: 'none'});
 
         _converse.incrementMsgCounter = function () {
             msg_counter += 1;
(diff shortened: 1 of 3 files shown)


src/headless/converse-chatboxes.js
```diff
@@ -31,10 +32,11 @@ converse.plugins.add('converse-chatboxes', {
             'privateChatsAutoJoined'
         ]);
 
-        let msg_counter = 0;
+        let msg_counter = 0, favicon = new Favico({type : 'circle', position: 'up', animation: 'none'});
```
jcbrand, line 35: The tests are failing because you're using `let` to instantiate an unchanging variable `favicon`.

Please use `const` instead.

src/headless/converse-chatboxes.js
```diff
@@ -7,6 +7,7 @@ import "./converse-emoji";
 import { Collection } from "@converse/skeletor/src/collection";
 import { _converse, api, converse } from "./converse-core";
 import log from "./log";
+import Favico from 'favico.js-slevomat';
```
jcbrand, line 10: Why are you using the forked version of favico.js ?

deleolajide: latest version 0.3.11 from this commit in 2018
https://github.com/slevomat/favico.js/commit/c0f5ea878e7a4d1ae2b18cdc6d30a093e7961d40

The original project is at version 0.3.10 and has not be touched since 2016

jcbrand:
Thanks!

PR #2356, opened 2020-12-04
Add in_memory storage option

XaviFP:
Complementary work for https://github.com/conversejs/skeletor/pull/6
Changed git hash to not forget about updating it when it gets merged.

### DIFF
--- a/src/headless/core.js
+++ b/src/headless/core.js
@@ -998,6 +998,9 @@ function initPersistentStorage () {
     } else if (_converse.api.settings.get("persistent_store") === 'IndexedDB') {
         config['description'] = 'indexedDB instance';
         config['driver'] = [Storage.localForage.INDEXEDDB];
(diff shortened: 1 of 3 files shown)


src/headless/core.js
```diff
@@ -998,6 +998,9 @@ function initPersistentStorage () {
     } else if (_converse.api.settings.get("persistent_store") === 'IndexedDB') {
         config['description'] = 'indexedDB instance';
         config['driver'] = [Storage.localForage.INDEXEDDB];
+    } else if (_converse.api.settings.get("persistent_store") === 'in_memory') {
```
jcbrand, line 1001: I prefer `none` to `in_memory` for the config setting value.

It's more semantic because strictly speaking, there is no persistent storage in this case.

XaviFP: Makes sense :)

jcbrand:
Thanks @XaviFP.

The tests are failing and need to be fixed first.

PR #2400, opened 2021-02-08
Fixes infinite loop bug when appending .png to allowed image urls

afuggini:


jcbrand:
Thanks! Merged manually

PR #2410, opened 2021-02-17
fix #2409 - Integrate App Badging API for unread messages

### LINKED ISSUE #2409: Integrate App Badging API for unread messages
If would be nice to have integrated support for the App Badging API so when converse is installed as a PWA a badge can be shown to show the number of unread messages.

More info: https://web.dev/badging-api/

deleolajide:
I have just made this minor change in Pade and decided to raise a PR to do the same in Converse as I had done it before on the browser tab icon with favico . It does not need a test


jcbrand:
Thanks @deleolajide, merged manually.

PR #2439, opened 2021-03-24
Move the httpToGeoUri util into the core

ubermanu:
The "sendMessage" function of chatboxes throw an error in the headless build

> converse.env.utils.httpToGeoUri does not exist

Usage:
* src/headless/plugins/chat/model.js:797
* src/headless/plugins/muc/muc.js:946


jcbrand:
Thanks!

PR #2458, opened 2021-04-13
Trigger event when connection status changes

afuggini:
Triggers api event `connectionStatusChange` when connection status changes.

### DIFF
--- a/src/headless/shared/connection.js
+++ b/src/headless/shared/connection.js
@@ -247,6 +247,12 @@ export class Connection extends Strophe.Connection {
      * @param { String } message
      */
     onConnectStatusChanged (status, message) {
+        /**
+         * Synchronous event triggered on connection status change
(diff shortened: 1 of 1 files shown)


src/headless/shared/connection.js
```diff
@@ -247,6 +247,12 @@ export class Connection extends Strophe.Connection {
      * @param { String } message
      */
     onConnectStatusChanged (status, message) {
+        /**
+         * Synchronous event triggered on connection status change
+         * @event _converse#onConnectStatusChanged
+         */
+        api.trigger('connectionStatusChange', { status, message });
```
jcbrand, line 254: Thanks, but the JSDoc description is wrong, this is not a synchronous event and the event name in the JSDoc is `onConnectStatusChanged` while the actual event name is `connectionStatusChange`.

PR #2505, opened 2021-05-21
Move decodeHTMLEntities util into headless

ubermanu:
Fixes the unknown 'shared/utils' path

On a side note these dependencies are missing in headless package.json:

* dayjs
* lit
* xss


jcbrand:
Thanks!

PR #2529, opened 2021-06-25
Fix stall document events

ubermanu:
When openening back and forth the same 2 MUCs, JS heap size goes up without being ever collected by the GC (even forced)

Browsers tested (archlinux):

- Chrome: 90.0.4430.93
- Firefox: 89.0.1

Those event listeners, if not removed properly, keep references to the chatboxes models everytime a room is open (in fullscreen mode)


jcbrand:
Thanks!

PR #2593, opened 2021-08-05
compress new avatar images

afriedmanGlacier:
Many images that might be used for avatars have a large file size, and we've seen this hit the 10 second timer on sendIQ (possibly because it is not compressed and the packets are not broken into chunks). This compresses avatars before sending them to the server.

Currently sets target size to less than 100KB and max width/height of 256.

This implementation requires new dependency client-compress (added to package.json and package-lock.json).


jcbrand:
Thanks for the pull request. Looks like a useful addition, but unfortunately the compression library you're using breaks transparent PNGs.

For that reason, I think we'd have to add an option (checkbox) to the user to turn off compression.

jcbrand:
Merged manually, thanks for the PR!

PR #2635, opened 2021-09-06
Fixes for Safari 10 and iOS

jcbrand:


src/headless/plugins/chat/model.js
```diff
@@ -160,7 +160,7 @@ const ChatBox = ModelWithContact.extend({
     async handleErrorMessageStanza (stanza) {
         const { __ } = _converse;
         const attrs = await parseMessage(stanza, _converse);
-        if (!await this.shouldShowErrorMessage(attrs)) {
+        if (await !this.shouldShowErrorMessage(attrs)) {
             return;
```
jcbrand, line 164: @afuggini This breaks the logic of the expression, because the `!` operator is evaluated before `await`.

See here:

```javascript
await !(new Promise(resolve => setTimeout(() => resolve(false), 500)))
false
!await (new Promise(resolve => setTimeout(() => resolve(false), 500)))
true
```

PR #2636, opened 2021-09-08
Ensure that the converse session is actually present

TheStranjer:
Sometimes, when registering, and when it tries to log you in, it tries to `get` something from the `shared_converse.session` object without it actually being there:

image

I noticed the expectation that I write a test, but I'm not 100% certain how to replicate it and therefore don't know how to test it. I doubt this works on all servers, or with all settings combinations, otherwise it would be noticed sooner. However, this is such a small change that hopefully the lack of a test can be allowed to slide.

Thanks!


jcbrand:
Thanks!

PR #2640, opened 2021-09-09
Adds hook to fetchLoginCredentials function

afuggini:
Adds hook to fetchLoginCredentials function to be able to modify the request, add headers, et. before sent.

### DIFF
--- a/src/headless/utils/init.js
+++ b/src/headless/utils/init.js
@@ -248,6 +248,10 @@ function fetchLoginCredentials (wait=0) {
                 }
             };
             xhr.onerror = reject;
+            /**
+             * *Hook* which allows modifying the server request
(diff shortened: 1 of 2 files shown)


src/headless/utils/init.js
```diff
@@ -248,6 +248,7 @@ function fetchLoginCredentials (wait=0) {
                 }
             };
             xhr.onerror = reject;
+            xhr = await _converse.api.hook('beforeFetchLoginCredentials', this, xhr);
```
jcbrand, line 251: Please also add a JSDoc string to document this hook. If you look at other hook calls in the code you'll see examples.

afuggini: Done

jcbrand:
Thanks!

PR #2755, opened 2021-12-15
fix #2751 - Media not rendered when Converse runs in a browser extension

### LINKED ISSUE #2751: Media not rendered when Converse runs in a browser extension
**Describe the bug**
When Converse is run inside a Chrome or Edge extension, **window.location.protocol** return "chrome-extension:". This causes the checkFileTypes  method to return false.
See https://github.com/conversejs/converse.js/blob/6cff47c6d8ff028c31b650f6c8dbf1ef85a1f194/src/headless/utils/url.js#L24

**To Reproduce**
1. Run ConverseJs in a Chrome extension and upload an image.
2. Image is not rendered. Instead the image URL is turned into a clickable URL

**Expected behavior**
Image to be rendered

**Screenshots**

image

 - Desktop 
 - Browser Chrome and Edge browser extensions
 - Converse.js version 9.0.0

**Additional context**
The screen shot was produced after hacking checkFileTypes and adding a test for 'chrome-extension:'
```
function checkFileTypes (types, url) {
    const uri = getURI(url);
    if (!window.location.protocol === 'chrome-extension:' && (uri === null || !checkTLS(uri))) {
        return false;
    }
    const filename = uri.filename().toLowerCase();
    return !!types.filter(ext => filename.endsWith(ext)).length;
}
```

deleolajide:

- [x ] Add a changelog entry for your change in `CHANGES.md`

### DIFF
--- a/src/headless/utils/url.js
+++ b/src/headless/utils/url.js
@@ -21,7 +21,7 @@ export function getURI (url) {
 
 function checkFileTypes (types, url) {
     const uri = getURI(url);
-    if (uri === null || !checkTLS(uri)) {
+    if (window.location.protocol !== 'chrome-extension:' && (uri === null || !checkTLS(uri))) {
         return false;
     }
     const filename = uri.filename().toLowerCase();
--- a/src/headless/utils/url.js
+++ b/src/headless/utils/url.js
@@ -21,7 +21,8 @@ export function getURI (url) {
 
 function checkFileTypes (types, url) {
     const uri = getURI(url);
-    if (window.location.protocol !== 'chrome-extension:' && (uri === null || !checkTLS(uri))) {
+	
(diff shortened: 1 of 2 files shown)


src/headless/utils/url.js
```diff
@@ -21,7 +21,7 @@ export function getURI (url) {
 
 function checkFileTypes (types, url) {
     const uri = getURI(url);
-    if (uri === null || !checkTLS(uri)) {
+    if (window.location.protocol !== 'chrome-extension:' && (uri === null || !checkTLS(uri))) {
```
jcbrand, line 24: Looking at this again, I think this needs to be changed to:

`(uri === null || (window.location.protocol !== 'chrome-extension:' && !checkTLS(uri)))`

Otherwise, in a chrome extension, if `uri` is `null` a `TypeError` will be thrown on line 27.

jcbrand:
Thanks!

PR #2841, opened 2022-03-08
Fixes #2754

### LINKED ISSUE #2754: Navigation keys doesn't work anymore into chat input message
**Describe the bug**
Sometimes, but often after trying to edit a previous message, navigation keys into input message do not work anymore.
It is no more possible to use left key, for example, to position the cursor one letter before.

**To Reproduce**
It is been a while that I try to identify a reproductible case, but without success.
Has anyone ever got this behavior?

**Expected behavior**
Navigation keys always usable.

 - Desktop
 - Browser all common WEB browsers
 - Since Converse.js v7 (I don't know for previous versions)

**Additional context**
ConverseJS is used in overlayed mode.
<details>

```
 converse.initialize({
          'view_mode' : 'overlayed',
          'i18n' : 'a language',
          'play_sounds' : false,
          'allow_logout' : false,
          'auto_login' : true,
          'auto_reconnect' : true,
          'jid' : 'a jid',
          'password' : 'a password',
          'locked_muc_nickname' : true,
          'nickname' : 'a nickname',
          'roster_groups' : false,
          'allow_registration' : false,
        });
```

</details>

SilverYoCha:
About https://github.com/conversejs/converse.js/issues/2754
That's a try.
Could maybe help to think about a better fix.

### DIFF
--- a/src/shared/chat/emoji-dropdown.js
+++ b/src/shared/chat/emoji-dropdown.js
@@ -95,6 +95,12 @@ export default class EmojiDropdown extends DropdownBase {
         super.showMenu();
         setTimeout(() => this.querySelector('.emoji-search')?.focus());
     }
+
+    hideMenu () {
+        this.chatview.querySelector('converse-emoji-picker')?.disableArrowNavigation();
+        super.hideMenu();
+        setTimeout(() => this.chatview.querySelector('.chat-textarea')?.focus());
+    }
 }
 
 api.elements.define('converse-emoji-dropdown', EmojiDropdown);
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,7 @@
 - #2733: Fix OMEMO race condition related to automatic reconnection and SMACKS
 - #2733: Wait for decrypted/parsed message before queuing to UI
 - #2751: Media not rendered when Converse runs in a browser extension
+- #2754: Navigation keys doesn't work anymore into chat input message
 - #2781: Flashing error when connecting to a room
 - #2786: Fix webpack configuration not working on Windows OS
 - #2788: `TypeError` when trying to use `@converse/headless`
--- a/src/shared/chat/emoji-picker.js
+++ b/src/shared/chat/emoji-picker.js
@@ -189,7 +189,6 @@ export default class EmojiPicker extends CustomElement {
             this.onEnterPressed(ev);
         } else if (ev.keyCode === converse.keycodes.ESCAPE) {
             u.ancestor(this, 'converse-emoji-dropdown').hideMenu();
-            this.chatview.el.querySelector('.chat-textarea').focus();
             ev.stopPropagation();
             ev.preventDefault();
         } else if (


src/shared/chat/emoji-dropdown.js
```diff
@@ -95,6 +95,12 @@ export default class EmojiDropdown extends DropdownBase {
         super.showMenu();
         setTimeout(() => this.querySelector('.emoji-search')?.focus());
     }
+
+    hideMenu () {
+        this.chatview.querySelector('converse-emoji-picker')?.disableArrowNavigation();
```
jcbrand, line 100: In `src/shared/chat/emoji-picker.js` there is a function `_onGlobalKeyDown`. I think it would be better to listen for the escape key and then to call `disableArrowNavigation` there, instead of doing it here.

src/shared/chat/emoji-dropdown.js
```diff
@@ -95,6 +95,12 @@ export default class EmojiDropdown extends DropdownBase {
...
+        super.hideMenu();
+        setTimeout(() => this.chatview.querySelector('.chat-textarea')?.focus());
```
jcbrand, line 102: Why did you move this here? I think it's better to leave this line where it was (including the `setTimeout` if that is somehow necessary to fix a race condition).

This, coupled with the above change would mean that it's no longer necessary to overwrite `hideMenu` in this component.

jcbrand:
Thanks for the PR @SilverYoCha :)

I've left two review comments.

jcbrand (changes requested):

src/shared/chat/emoji-dropdown.js
SilverYoCha: Thank you @jcbrand for the consideration of my PR.

There are several use cases with the emojis drop down concerning its closing. All these use cases do not result in calling same code when closing the drop down.
For example, closing the drop down just after its opening by using ESCAPE key, calls `onKeyDown` method in `emoji-picker.js`. But after having navigated into emoji list, closing it by using ESCAPE key or by clicking outside the emoji drop down doesn't call the `onKeyDown` method.
My idea was to find a method which was performed whatever the closing case. It seems to be the case of `hideMenu` method of `emojis-dropdown.js` (by overriding). I found that it was not so bad to centralize the closing stuffs into this method, and so I overrided it. 

That being said, I can try your suggestions.

SilverYoCha: It seems to be same problematics with `_onGlobalKeyDown` method. It is not called whatever the context into which the emoji drop down is closed.

jcbrand: Ok. The reason I made my suggestions, is because I'd like to reduce the usage of imperative code where for example we call `querySelector` to call methods on components and because ideally we should decouple components from one another and have the components be as declarative as possible (i.e. more like React and less like jQuery).

In other words, I think it's an anti-pattern to continuously use things like `this.chatview`, because it tightly couples this component to the component referenced by `this.chatview`. The reason why we still use `this.chatview` in various places is because it was the quickest way to turn all the backbone views into web components. But ideally we should try and get the components as independent of one another as possible and remove the usage of `this.chatview` where ever it's still used.

I think eventually we probably will move the emoji picker component into a separate NPM package (and perhaps a separate Github repo), and in that case we won't be able to rely on `this.chatview`.

SilverYoCha: I understand your aim and I agree with it.
Unfortunately, my knowing of the entire of the project, and more generally of JavaScript framework universe, are not enough to propose better for now.
Don't hesitate to close my PR without merging it.
I tried :-)
Thank you @jcbrand.

jcbrand: No worries, I've merged your commit and then did some refactoring to bring things more in line with the goals stated above.

Can you please check whether things still work for you as expected?

deleolajide:
>Can you please check whether things still work for you as expected?

I can't reproduce the issue with this commit, but my testing is not exhaustive. I will watch out for it over the next few days as it used to occur enough to notice for me

src/shared/chat/emoji-dropdown.js
SilverYoCha: Things are still working as expected :-)
:+1:

PR #2862, opened 2022-03-29
Fix endless loop of adding .png to image URL.

kicken:
If an image fails to load, converse-desktop ends up in an endless loop of adding .png to the image url and trying again.  Adding the file protocol as an exception resolves the problem.

### DIFF
--- a/src/headless/utils/url.js
+++ b/src/headless/utils/url.js
@@ -21,7 +21,7 @@ export function getURI (url) {
 
 function checkFileTypes (types, url) {
     const uri = getURI(url);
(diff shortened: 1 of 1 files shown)


jcbrand:
Thanks, I've merged your commit manually.

Looks to me like the issue was that `isURLWithImageExtension` was always returning `false`, causing the loop.

I've done some further refactoring so that `checkFileTypes` only checks whether the URL points to a file with a particular extension and the other code which checks the protocol I moved to a new function.

This makes things clearer.

I think the bug you uncovered should still be fixed, could you please double-check?

kicken:
> I think the bug you uncovered should still be fixed, could you please double-check?

Seems to be fixed still.  Thanks.

PR #2863, opened 2022-03-30
ESM compat for @converse/headless

ubermanu:
Add missing source files to NPM so it can be imported as an ES module

### DIFF
--- a/src/headless/package.json
+++ b/src/headless/package.json
@@ -18,7 +18,8 @@
   ],
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/conversejs/converse.js.git"
+    "url": "git+https://github.com/conversejs/converse.js.git",
+    "directory": "src/headless"
   },
   "scripts": {
     "test": "echo \"Error: run tests from root\" && exit 1"
--- a/src/headless/package.json
+++ b/src/headless/package.json
@@ -6,15 +6,6 @@
   "homepage": "https://conversejs.org",
   "license": "MPL-2.0",
   "main": "dist/converse-headless.min.js",
-  "files": [
-    "*.js",
-    "*.json",
-    "dist/",
-    "README.md",
-    "utils/*.js",
-    "templates/*.html",
-    "templates/*.svg"
-  ],
   "keywords": [
     "converse.js",
     "XMPP",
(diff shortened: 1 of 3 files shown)


ubermanu:
Still `i18n` is missing (since it's supposed to be in the root package)

- src/headless/plugins/bookmarks/collection.js:4
- src/headless/plugins/roster/contacts.js:6
- src/headless/plugins/roster/index.js:10

should these just point to src/headless/shared/i18n.js ?

ubermanu:
I've changed

```js
import { __ } from 'i18n';
```

into 

```js
import i18n from '../../shared/i18n.js';
const { __ } = i18n;
```

Also I've changed `converse/headless` paths to their relative counterpart

Those changes are in another branch and tests are running fine, if you want to check it out: https://github.com/ubermanu/converse.js/tree/headless_esm_2

jcbrand:
Hi @ubermanu 

Thanks for this 👍 

Concerning use of `__` in headless, it shouldn't be imported because it's not available from inside headless.
Instead it should be taken at runtime from the `converse` object. When running @converse/headless by itself, it will then be a no-op (or an integrator could add their own translation engine to `__`).

src/headless/package.json
```diff
@@ -35,11 +28,13 @@
...
   },
   "gitHead": "9641dcdc820e029b05930479c242d2b707bbe8e2",
-  "devDependencies": {},
   "dependencies": {
+    "@converse/openpromise": "^0.0.1",
     "@converse/skeletor": "conversejs/skeletor#c7e76b023337a8ac201c74507f7a1209de1c6d18",
     "dayjs": "1.10.6",
+    "dompurify": "^2.3.1",
     "filesize": "^7.0.0",
+    "lit": "^2.0.0-rc.2",
```
jcbrand, line 37: I don't think we should add `lit` as a dependency since it's a UI framework and `@converse/headless` is supposed to be UI agnostic.

I see that we're only using `lit-tml` and only in one file (src/headless/plugins//emojis/index.js), so at most we should add `lit-html` as dependency and then import from that, but I think what would be best would be to refactor the emojis plugin so that it doesn't require `lit-html` in `@converse/headless` and instead move code that relies on `lit-html` to the UI part of Converse.

jcbrand: I've now refactored @converse/headless to remove the dependency on `lit-html`

https://github.com/conversejs/converse.js/commit/9d4382c7540b8fc635652ba32a4a7b7ed6671cfb

jcbrand:
@ubermanu I've manually merged your commits and modified your last commit to remove the explicit dependency on `lit`.

PR #2873, opened 2022-04-07
Fixing empty chat box on a re-connection when 'clear_messages_on_reconnection' parameter is enabled

SilverYoCha:


jcbrand:
Thanks @SilverYoCha.

I've made a different fix for this because the problem was actually deeper in the stack.

I've fixed it in these two commits:
https://github.com/conversejs/mergebounce/commit/1a8a8b26326c8d2f12ca3c85dfe29ff977b36755
https://github.com/conversejs/converse.js/commit/0c049dd3649c29eb9d29d009d72b6a8bb02f1628

SilverYoCha:
Thank you!

Issue #1841, opened 2020-01-24
Include list of public chats from search.jabber.network


ericschdt:
It's so comfortable to have a list of well known public MUCs like in Conversations to choose from, that I really miss having feature in Converse.js :)

You may also go to the website, but 1) you have to know about it, 2) it's still at least 5 clicks more. ;)

deleolajide:
This feature request has also been mention [here](https://github.com/conversejs/community-plugins/issues/41)

PawBud:
> This feature request has also been mention [here](https://github.com/conversejs/community-plugins/issues/41)

Thanks for bringing it to my attention, I believe @Echolon can decide which issue to close

Echolon:
@PawBud happy to support your preparations!

PawBud:
> @PawBud happy to support your preparations!

Thanks, btw, I believe the [HTTP API](https://search.jabber.network/docs/api) shall suffice, let me try to implement it.

PawBud:
@Echolon 
<img width="504" alt="Screen Shot 2022-04-15 at 6 34 56 PM" src="https://user-images.githubusercontent.com/48093317/163574116-e4cd6681-62ad-48fe-80eb-a395fd9f79c2.png">
I feel like the menu would be displayed when the `show popular public groupchats` button is clicked

jcbrand:
@PawBud: Thanks for looking into this. I think search.jabber.network should be used to provide autocomplete suggestions when you join a MUC via the add-chatroom modal, not via a button in the list-chatrooms modal that you show above.

Converse has a component `converse-autocomplete` that can be used to create inputs with auto-complete suggestions.

See here:
https://github.com/conversejs/converse.js/blob/7cb86638b5e2e7fd3832ceecf7b1a8fefeba51ef/src/shared/autocomplete/component.js#L7

You pass it a `getAutoCompleteList` function which returns the autocomplete suggestions.

Here's an example of where the `converse-autocomplete` component is being used:
https://github.com/conversejs/converse.js/blob/7cb86638b5e2e7fd3832ceecf7b1a8fefeba51ef/src/plugins/muc-views/templates/ad-hoc.js#L22

The goal for this ticket would be to replace the `input` in the add-modal template, with a `converse-autocomplete` component.
https://github.com/conversejs/converse.js/blob/7cb86638b5e2e7fd3832ceecf7b1a8fefeba51ef/src/plugins/muc-views/templates/add-muc.js#L37

[...]

PR #2881, opened 2022-04-16
AutoComplete: the groups that are bookmarked but not joined will be s…

PawBud:
Resolves: #1841 

So, the issue describes having XMPP muc's fetched from [search.jabber.network](https://search.jabber.network/rooms/1). While implementing the API, I faced an error:- 
```
Access to fetch at '[https://search.jabber.network/api/1.0/rooms' from origin 'http://127.0.0.1:8000' has been blocked by CORS policy: No 'Access-Control-Allow-Origin]
```

Given that it's a problem from jabber.search.network, @jcbrand suggested that we instead add autocomplete using bookmarked group chats.

[Click here to see the working of unjoined bookmarks](https://github.com/conversejs/converse.js/files/8504991/2022-04-18.15-33-13.mp4.zip)

Working after 38b0033f0c342a7ba36e5c90f8e6ef18a315be99:

https://user-images.githubusercontent.com/48093317/164402865-46d567ba-7f3b-4163-b50a-bd6d7d590488.mp4

### DIFF
--- a/src/plugins/muc-views/templates/add-muc.js
+++ b/src/plugins/muc-views/templates/add-muc.js
@@ -4,6 +4,7 @@ import { api } from '@converse/headless/core.js';
 import { html } from "lit";
 import { modal_header_close_button } from "plugins/modal/templates/buttons.js"
 import { unsafeHTML } from "lit/directives/unsafe-html.js";
+import { getAutoCompleteList } from "../utils.js";
 
 
 const nickname_input = (o) => {
@@ -34,7 +35,10 @@ export default (o) => {
                         <div class="form-group">
                             <label for="chatroom">${o.label_room_address}:</label>
                             ${ (o.muc_roomid_policy_error_msg) ? html`<label class="roomid-policy-error">${o.muc_roomid_policy_error_msg}</label>` : '' }
-                            <input type="text" required="required" name="chatroom" class="form-control roomjid-input" placeholder="${o.chatroom_placeholder}"/>
+                            <converse-autocomplete
+                            .getAutoCompleteList="${getAutoCompleteList}"
+                            placeholder="${o.chatroom_placeholder}"
+                            name="chatroom"/>
                         </div>
                         ${ o.muc_roomid_policy_hint ?  html`<div class="form-group">${unsafeHTML(DOMPurify.sanitize(o.muc_roomid_policy_hint, {'ALLOWED_TAGS': ['b', 'br', 'em']}))}</div>` : '' }
                         ${ !api.settings.get('locked_muc_nickname') ? nickname_input(o) : '' }
--- a/src/plugins/muc-views/utils.js
+++ b/src/plugins/muc-views/utils.js
@@ -127,9 +127,27 @@ export function getAutoCompleteListItem (text, input) {
     return element;
 }
 
-export async function getAutoCompleteList () {
-    const models = [...(await api.rooms.get()), ...(await api.contacts.get())];
-    const jids = [...new Set(models.map(o => Strophe.getDomainFromJid(o.get('jid'))))];
+// models array is globalized to introduce caching of the list from https://search.jabber.network/rooms/1.
+let jids;
+let timestamp = (new Date()).toISOString();
+let firstTimeCall = true;
+
+async function fetchListOfRooms() {
+    const response = await fetch('https://search.jabber.network/api/1.0/rooms');
+    const data = await response.json();
+    const popular_mucs = [];
+    for (var x = 0; x < data.items.length; x++) {
+        popular_mucs.push(data.items[x]["address"]);
+    }
+    jids = [...new Set(popular_mucs)];
+}
+
+export function getAutoCompleteList() {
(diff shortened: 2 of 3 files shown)


jcbrand:
Thanks @PawBud.

There are lots of breaking tests though, I see errors like this for example:
`ERROR: TypeError: Cannot read properties of null (reading 'placeholder')`

src/plugins/muc-views/utils.js
```diff
@@ -127,9 +127,24 @@ export function getAutoCompleteListItem (text, input) {
     return element;
 }
 
-export async function getAutoCompleteList () {
-    const models = [...(await api.rooms.get()), ...(await api.contacts.get())];
-    const jids = [...new Set(models.map(o => Strophe.getDomainFromJid(o.get('jid'))))];
+// models array is globalized to introduce caching of the list from https://search.jabber.network/rooms/1.
+const models = [];
```
jcbrand, line 131: Let's not call this `models`. The word `model` has a very specific meaning in Converse, it refers to instances of the Backbone/Skeletor [Model](https://github.com/conversejs/skeletor/blob/af329a269841d434f1445e35d67a3381a9186b11/src/model.js#L41) class.

search.jabber.network returns the most popular MUCs right? So perhaps call it `popular_mucs`.

src/plugins/muc-views/utils.js
```diff
@@ -127,9 +127,24 @@ export function getAutoCompleteListItem (text, input) {
...
+    const data = await response.json();
+    for (var x = 0; x < data.items.length; x++) {
+        models.push(data.items[x]["address"]);
+    }
+}
+
+export function getAutoCompleteList() {
+    if (callFetchListOfRooms === false) {
+        fetchListOfRooms();
+        callFetchListOfRooms = true;
```
jcbrand, line 145: Instead of setting a boolean, you can set a timestamp and then use DayJS to check whether the current time minus the timestamp is above a certain threshold, (e.g. 4 hours).

src/plugins/muc-views/utils.js
```diff
@@ -127,9 +127,24 @@ export function getAutoCompleteListItem (text, input) {
...
+    }
+    const jids = [...new Set(models)];
```
jcbrand, line 147: Instead of always making sure that we return unique jids, we can rather just do that once in `fetchListOfRooms` when we populate `models`.

src/plugins/muc-views/templates/add-muc.js
```diff
@@ -21,6 +22,7 @@ const nickname_input = (o) => {
 export default (o) => {
     const i18n_join = __('Join');
     const i18n_enter = __('Enter a new Groupchat');
+    const i18n_jid_placeholder = __('Enter Server Name');
```
jcbrand, line 25: There was already a placeholder being used `o.chatroom_placeholder`. I think you should just continue using that instead of creating your own one (and leaving the original one as dead code not being used).

jcbrand (changes requested):

src/plugins/muc-views/utils.js
PawBud: I would be ok with it, but is it required? the above implementation simply fetches once per session, and it's simple to understand. Do we still need a timestamp and need to fetch the list in every interval? I feel that it is unnecessary, what's your opinion @jcbrand?

jcbrand: Using a timestamp is trivial as well. 

Instead of `if (callFetchListOfRooms === false) {` it becomes something like `if (dayjs().isAfter(time_fetched, 'day')) }`

Web sessions are quite short, but as more people start using the Electron version, a session can last many days or weeks.

src/shared/autocomplete/autocomplete.js
```diff
@@ -266,6 +266,9 @@ export class AutoComplete {
         }
 
         const list = typeof this._list === "function" ? await this._list() : this._list;
+        if (typeof (list) === 'undefined') {
+            return 0;
+        }
```
PawBud, line 271: @jcbrand, so given that the [jids](https://github.com/conversejs/converse.js/pull/2881/files#diff-10ea71a0f5e70bf68e685adee44d6718e8e6f408175f67a59be213eaf7b5f7deR131) is undefined, we were accessing its length, which threw an error, so I simply handled it via returning a 0.

src/plugins/muc-views/templates/add-muc.js
PawBud: Sure, will fix it.

src/plugins/muc-views/utils.js
```diff
@@ -127,9 +127,28 @@ export function getAutoCompleteListItem (text, input) {
     return element;
 }
 
-export async function getAutoCompleteList () {
-    const models = [...(await api.rooms.get()), ...(await api.contacts.get())];
-    const jids = [...new Set(models.map(o => Strophe.getDomainFromJid(o.get('jid'))))];
+// models array is globalized to introduce caching of the list from https://search.jabber.network/rooms/1.
+let jids;
+// let callFetchListOfRooms = false;
```
PawBud, line 132: To be removed

PR #2934, opened 2022-05-26
add-muc: added the locked_muc_domain condition

PawBud:
we need to not use the `converse-autocomplete` component when `locked_muc_domain` is `true`. A simple input as previously defined is ample in this case.

[locked-muc-domain](https://conversejs.org/docs/html/configuration.html#locked-muc-domain)


jcbrand:
Thanks!

PR #2944, opened 2022-06-06
Fixes #2943 (suggestion)

### LINKED ISSUE #2943: With a simple chat box, newest messages are not always fetched when setting 'clear_messages_on_reconnection' parameter to true.
With a simple chat box, newest messages are not always fetched when setting 'clear_messages_on_reconnection' parameter to true.
When closing and opening again a chat box during a same session, blank space is displayed instead of messages.

**To reproduce**
On my environment, this scenario is reproducible :
1. Open ConverseJs with a user account
2. Open a chat box and converse into it
3. Close the chat box
4. Open again the chat box
No message is displayed.

**Expected behaviour**
Messages (those into history and the newest) should be displayed.

**Environment:**
 - Desktop / Mobile
 - All browsers
 - Converse.js (master branch and 9.1.1, not tried on other versions)

**Additional context**
Functional ejabberd server.
<details>

```
 converse.initialize({
          'view_mode' : 'overlayed',
          'i18n' : 'a language',
          'play_sounds' : false,
          'allow_logout' : false,
          'auto_login' : true,
          'auto_reconnect' : true,
          'jid' : 'a jid',
          'clear_messages_on_reconnection' : true,
          'password' : 'a password',
          'locked_muc_nickname' : true,
          'nickname' : 'a nickname',
          'roster_groups' : false,
          'allow_registration' : false,
        });
```

</details>

SilverYoCha:
Listen also to 'chatBoxInitialized' event in order to fetch newset messages.

Adding robustness when closing a chat box.

Removing console logs into control box template.


src/headless/plugins/mam/index.js
```diff
@@ -47,12 +47,14 @@ converse.plugins.add('converse-mam', {
...
-        api.listen.on('chatReconnected', chat => {
+        const __fetchNewestMessages = chat => {
             // XXX: For MUCs, we listen to enteredNewRoom instead
             if (chat.get('type') === _converse.PRIVATE_CHAT_TYPE) {
                 fetchNewestMessages(chat);
             }
-        });
+        };
+        api.listen.on('chatReconnected', __fetchNewestMessages);
+        api.listen.on('chatBoxInitialized', __fetchNewestMessages);
```
jcbrand, line 58: The `afterMessagesFetched` event is already used for newly created chats and AFAIK it should fire when you close a chat and reopen it. So the `chatBoxInitialized` event here is not the right one to listen to and we already listen to the right event below.

I suspect the messages aren't fetched because of ` && !_converse.connection.restored`.

Perhaps you can check?

Are you using BOSH or websocket?

`_converse.connection.restored` is usually only relevant if you're using BOSH and it will be true if you reloaded the page after logging in.

So if this is the cause, then you should also not be able to reproduce the bug without first reloading the page after logging in. Can you please check?

jcbrand (changes requested):

src/headless/plugins/chat/utils.js
```diff
@@ -28,7 +28,7 @@ async function handleErrorMessage (stanza) {
         return;
     }
     const chatbox = await api.chatboxes.get(from_jid);
-    if (chatbox.get('type') === _converse.PRIVATE_CHAT_TYPE) {
+    if (chatbox && chatbox.get('type') === _converse.PRIVATE_CHAT_TYPE) {
```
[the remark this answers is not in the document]
SilverYoCha: Yes indeed!

src/headless/plugins/mam/index.js
SilverYoCha: > I suspect the messages aren't fetched because of ` && !_converse.connection.restored`.
> 
> Perhaps you can check?

Yes, you are right.



> Are you using BOSH or websocket?

I am using websocket.



> `_converse.connection.restored` is usually only relevant if you're using BOSH and it will be true if you reloaded the page after logging in.
> 
> So if this is the cause, then you should also not be able to reproduce the bug without first reloading the page after logging in. Can you please check?

Yes, you are entirely right.

SilverYoCha: I have modified my PR.
I propose to take into account 'clear_messages_on_reconnection' setting into 'afterMessagesFetched' event listener.

jcbrand:
@SilverYoCha I recently removed the check `!_converse.connection.restored` because it would cause chats opened after refresh to be empty :grimacing: 

I think in doing that I also fixed the issue you had, and so this MR is no longer needed. Thanks for your effort!

SilverYoCha:
@jcbrand yes indeed, I tried without `!_converse.connection.restored` and it seems to work.

PR #2945, opened 2022-06-08
Documentation updates

guusdk:


docs/source/api/index.rst
```diff
@@ -1,6 +1,6 @@
 .. raw:: html
 
-    <div id="banner"><a href="https://github.com/jcbrand/converse.js/blob/master/docs/source/theming.rst">Edit me on GitHub</a></div>
+    <div id="banner"><a href="https://github.com/jcbrand/converse.js/blob/master/docs/source/index.rst">Edit me on GitHub</a></div>
```
jcbrand, line 3: Looks like `api` is missing. Should be `docs/source/api/index.rst`

docs/source/features.rst
```diff
@@ -101,6 +107,12 @@ articles:
 OMEMO in Multi-user chats (MUC)
 -------------------------------
 
+.. note::
+    For end-to-end encryption via OMEMO, you'll need to load `libsignal-protocol.js
+    <https://github.com/signalapp/libsignal-protocol-javascript>`_ separately in
+    your page. Take a look at the section on :ref:`libsignal <dependency-libsignal>` and the
+    :ref:`security considerations around OMEMO <feature-omemo>`.
+
```
jcbrand, line 115: You already have the note in the same page above, perhaps a bit excessive to have it twice?

guusdk: I did that because it's a section that can be linked to with an anchor (making it easy to miss the note in the section above). I'm happy to remove it if you think that it's overkill.

docs/source/api/index.rst
guusdk: will fix / force push.

docs/source/features.rst
jcbrand: Yeah, I can see your point but I still think it's a bit much. I'd prefer it if you'd remove it, thanks.

guusdk: image

PR #2946, opened 2022-06-10
Add ejabberd and add XMPP servers homepages

badlop:
This is just a small documentation update: add ejabberd to the list, and also add home pages of all the three servers.


jcbrand:
Thanks!

PR #2947, opened 2022-06-13
fix vcard image hash error when image is string 'undefined'.

chenmj201601:


licaon-kter:
Why are there 31 style changes and one fixup?

PR #2956, opened 2022-06-20
WIP: Audio video capability

### LINKED ISSUE #447: [feature] Audio/video chat integration
Hello, I just found an old draft for an Audio/Video support request I was planning to submit here. @jcbrand please let me know if this is something you'd like to have in converse.js. I did a quick reseach on possible implentations. Pasting the contents below

---------------------------------------------

A quick search for FOSS a/v web chat solutions returned https://www.jsxc.org/ (XMPP + voice/video) and https://github.com/Rantanen/mumble-web (mumble). I didn't find the sources for [Firefox Hello](https://www.mozilla.org/en-US/firefox/hello/) (client and server), which works.

The go-to solution for audio/video over XMPP seems to be jingle. Much related to https://github.com/jcbrand/converse.js/issues/161

Relevant links I found:

> [XEP-0166: Jingle](https://xmpp.org/extensions/xep-0166.html) - This specification defines an XMPP protocol extension for initiating and managing peer-to-peer media sessions between two XMPP entities in a way that is interoperable with existing Internet standards. The protocol provides a pluggable model that enables the core session management semantics (compatible with SIP) to be used for a wide variety of application types (e.g., voice chat, video chat, file transfer) and with a wide variety of transport methods (e.g., TCP, UDP, ICE, application-specific transports).

>[XEP-0266: Codecs for Jingle Audio](https://xmpp.org/extensions/xep-0266.html) - This document describes implementation considerations related to audio codecs for use in Jingle RTP sessions.

>[XMPP Technologies: Jingle – The XMPP Standards Foundation](https://xmpp.org/about-xmpp/technology-overview/jingle/)

>[XEP-0167: Jingle RTP Sessions](https://xmpp.org/extensions/xep-0167.html) - This specification defines a Jingle application type for negotiating one or more sessions that use the Real-time Transport Protocol (RTP) to exchange media such as voice or video. The application type includes a straightforward mapping to Session Description Protocol (SDP) for interworking with SIP media endpoints.

>[Combining SIP and XMPP | OnSIP](http://www.onsip.com/about-voip/sip/combining-sip-and-xmpp) -> [sixpac Discussion Archive - Date Index](https://www.ietf.org/mail-archive/web/sixpac/current/maillist.html)

OTR does not seem to be able to encrypt audio/video streams, but there's **[ZRTP](https://en.wikipedia.org/wiki/ZRTP)**: [XEP-0262: Use of ZRTP in Jingle RTP Sessions](https://xmpp.org/extensions/xep-0262.html). Also:

> [[cryptography] Jingle and Otr](http://lists.randombit.net/pipermail/cryptography/2013-August/005064.html) - If you know the face of the person you are talking to, you can surely tell if the right person is speaking the right SAS, which makes the  methods used by OTR overkill for video.

-------------------------

PawBud:


src/plugins/chatview/index.js
```diff
@@ -51,7 +51,7 @@ converse.plugins.add('converse-chatview', {
             'time_format': 'HH:mm',
             'use_system_emojis': true,
             'visible_toolbar_buttons': {
-                'call': false,
+                'call': true,
```
jcbrand, line 54: This needs to be set back to `false`. This button is not related to Jingle. I'll probably just remove this button soon, but for now you can ignore it.

src/shared/chat/toolbar.js
```diff
@@ -54,16 +54,6 @@ export class ChatToolbar extends CustomElement {
...
-        if (this.show_call_button) {
-            const color = this.is_groupchat ? '--muc-toolbar-btn-color' : '--chat-toolbar-btn-color';
-            const i18n_start_call = __('Start a call');
-            buttons.push(html`
-                <button class="toggle-call" @click=${this.toggleCall} title="${i18n_start_call}">
-                    <converse-icon color="var(${color})" class="fa fa-phone" size="1em"></converse-icon>
-                </button>`
-            );
-        }
-
```
jcbrand, line 66: Please put back this code to avoid merge conflicts when I remove it in a separate commit and you then rebase.

jcbrand (changes requested):

src/plugins/chatview/index.js
deleolajide: >I'll probably just remove this button soon

Could you please keep this button. It has always been part of converse to enable click-to-call integration with external telephone/conferencing systems. I have one or two plugins that will break if this button is removed.

jcbrand: @deleolajide: There is now an easy way via hooks to add buttons to the toolbar, so I don't really think this button should be in core.

We can move the code to a community plugin, then you can include that package and continue using it as before.

src/plugins/jingle/chat-header-notification.js
```diff
@@ -0,0 +1,45 @@
...
+        this.listenTo(this.model, 'change:jingle_status', () => this.requestUpdate());
+    }
+    
+    render() {
+        return tpl_header_button(this);
+    }
+
+    endCall() {
+        const jingle_status = this.model.get('jingle_status');
+        if ( jingle_status === JINGLE_CALL_STATUS.OUTGOING_PENDING || jingle_status === JINGLE_CALL_STATUS.ACTIVE) {
```
jcbrand, line 28: The stanza with a `retract` element is only sent when you want to cancel an outgoing pending call, not an active call.

To stop an active call, a different stanza is sent.
https://xmpp.org/extensions/xep-0353.html#finish

src/plugins/jingle/chat-header-notification.js
```diff
@@ -0,0 +1,45 @@
...
+            this.model.save('jingle_status', JINGLE_CALL_STATUS.ENDED);
+            const stanza = $msg({
+                'from': _converse.bare_jid,
+                'to': this.jid,
+                'type': 'chat'
+            }).c('retract', {'xmlns': Strophe.NS.JINGLEMESSAGE, 'id': this.getAttribute('id')})
```
jcbrand, line 34: The `id` attribute of the `retract` element in this stanza must match the `id` attribute of the `propose` element that was originally sent.

Otherwise there is no way for a recipient of this stanza to know which proposed call it relates to (they might have received more than one proposed call from multiple senders).

Currently when you send the `propose` stanza, you're not saving its data anywhere for the sender.

So you need to make sure that a message object is created and stored for the sender when the `propose` stanza is sent out.

There are multiple potential ways in which this could potentially be done. For now, you can try to just call `model.messages.create(attrs)` (where `model` is the chatbox) when you send out the `propose` stanza. That might be the easiest to do for now.

Then, when you send out the `retract` stanza, you need to get the latest `propose` stanza (e.g. via `model.messages.find`) and then get the value of the `id` that you used for the `propose` element, and then use that on the `retract` element here.

jcbrand (changes requested):

src/plugins/jingle/tests/message-initiation.js
```diff
@@ -0,0 +1,58 @@
...
+    const view = _converse.chatboxviews.get(contact_jid);
+    const call_button = view.querySelector('converse-jingle-toolbar-button button');
+    call_button.click();
+    const sent_stanzas = _converse.connection.sent_stanzas;
+    const stanza = await u.waitUntil(() => sent_stanzas.filter(s => sizzle(`propose[xmlns='${Strophe.NS.JINGLEMESSAGE}']`, s).length).pop());
+    expect(Strophe.serialize(stanza)).toBe(
+        `<message from='${_converse.jid}'
+            to='${contact_jid}'
+            type='chat'>`+
+                `<propose xmlns='${Strophe.NS.JINGLEMESSAGE}' id='${stanza.getAttribute('id')}'>`+
```
jcbrand, line 24: `stanza.getAttribute('id')` will get the `id` attribute from the root of the stanza, which is the `message` element. You need to get it from the `propose` element.

jcbrand (changes requested):

src/plugins/jingle/chat-header-notification.js
PawBud: Thanks for pointing it out, corrected it

src/plugins/jingle/tests/message-initiation.js
PawBud: @jcbrand yeah about that, how do I get the `propose_id` object here https://github.com/conversejs/converse.js/pull/2956/files#diff-abe3b677d9605e0278786d222bb105d87006e60b51e0b531f6b82cd2e5cfa00cR34 ?

jcbrand: The `propose` element is part of the `stanza` that you're testing here. So you can use `stanza.querySelector` to get the `propose` element and then get the id from it.

PawBud: > The `propose` element is part of the `stanza` that you're testing here. So you can use `stanza.querySelector` to get the `propose` element and then get the id from it.

that makes sense, thanks a lot

src/plugins/jingle/chat-header-notification.js
```diff
@@ -0,0 +1,60 @@
...
+    
+    render() {
+        return tpl_header_button(this);
+    }
+
+    endCall() {
+        const jingle_status = this.model.get('jingle_status');
+        if ( jingle_status === JINGLE_CALL_STATUS.OUTGOING_PENDING ) {
+            this.model.save('jingle_status', JINGLE_CALL_STATUS.ENDED);
+            const initiator_stanza = this.model.messages.findWhere({ 'media': 'audio' });
```
PawBud, line 30: @jcbrand am getting `initiator_stanza: 1`

jcbrand: Your message is quite cryptic, I don't understand. Please provide more info.

PawBud: > You're getting a `1` back from `findWhere`? That doesn't seem likely.

that is correct

PawBud: Screen Shot 2022-07-18 at 7 58 55 PM
@jcbrand

src/headless/plugins/chat/model.js
```diff
@@ -228,6 +228,7 @@ const ChatBox = ModelWithContact.extend({
                 !this.handleChatMarker(attrs) &&
                 !(await this.handleRetraction(attrs))
         ) {
+            const message_retraction = await api.hook('onMessage', this, { retract: false });
```
PawBud, line 231: @jcbrand so this hook verifies the fact that we don't need to create a new message, right?

src/plugins/jingle/utils.js
```diff
@@ -0,0 +1,119 @@
...
+
+export function jingleCallInitialized() {
+    JingleCallModal;
+}
+
+/**
+ * This function simply sends the retraction stanza and modifies the attributes
+ */
+export function retractCall(context) {
+    const initiator_message = context.model.messages.findWhere({ 'media': 'audio' });
```
PawBud, line 43: @jcbrand I believe findWhere isn't working, I tested it several times, and yeah, if we use where, I am getting an array of 150 + to propose messages, which shouldn't happen I think?

src/plugins/jingle/utils.js
```diff
@@ -0,0 +1,137 @@
...
+    return el?.getAttribute('media');
+}
+
+function getJingleRetractionId(stanza){
+    const el = sizzle(`retract[xmlns="${Strophe.NS.JINGLEMESSAGE}"]`, stanza).pop();
+    return el?.getAttribute('id');
+}
+
+export function getJingleTemplate(model) {
+    return html`<converse-jingle-message jid='${model.get('jingle_status')}'></converse-jingle-message>`;
```
jcbrand, line 61: You're not passing in the `jid` of the contact here, you're passing in `model.get('jingle_status')`.

That's most likely why you're not getting the model in the component.

hessamhoseinkhani:
@PawBud why you removed your assignee ?

PawBud:
> @PawBud why you removed your assignee ?

I can't work on this issue anytime soon given my commitments, so for now anyone who wants to pick it up, feel free to do so!

hessamhoseinkhani:
@PawBud @jcbrand can you clarify this that do we need an explicit WebRTC library to make the call working ?

Neustradamus:
@PawBud: Can you look to finish your PR 2 years after?

Thanks in advance.

PR #2957, opened 2022-06-20
Fix enable_smacks doc

joudinet:
Fix documentation about the enable_smacks option, which is set to true by default since version 6.0.0.


jcbrand:
Thanks!

PR #3057, opened 2022-10-23
Save delay for future push.

kicken:
GCE comes before the frame data.  Save the delay to use in a future call to pushFrame.

Final frame needs to be pushed when EOF is encountered.  Intermediate frames are pushed when a new GCE is encountered.

### DIFF
--- a/src/shared/gif/index.js
+++ b/src/shared/gif/index.js
@@ -61,6 +61,7 @@ export default class ConverseGif {
         this.load_error = null;
         this.playing = this.options.autoplay;
         this.transparency = null;
+        this.frame_delay = null;
 
         this.frame_idx = 0;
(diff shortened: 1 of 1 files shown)


src/shared/gif/index.js
```diff
@@ -296,7 +299,7 @@ export default class ConverseGif {
         if (!this.frame) return;
         this.frames.push({
             data: this.frame.getImageData(0, 0, this.hdr.width, this.hdr.height),
-            delay,
+            delay: this.frame_delay
```
jcbrand, line 302: This function is still accepting a `delay` parameter which is no longer being used.

Otherwise, looks good, thanks!

kicken: I've removed the unused parameter.

Issue #3061, opened 2022-10-29
Add a way to add a MUC occupant to my roster


sloonz:
Right now, the only way to add a MUC occupant to my roster is to view its profile, copy its JID, close the profile, click on "Add Contact", paste the JID and add the contact. IMO, this is not very user-friendly. A better way would be (for example) to add a "Add to Contacts" button to the profile of the occupant.

I made an initial implementation here to open the discussion : https://github.com/sloonz/converse.js/commit/c2f5aa1d1adeb165d095dd18c76130b09c1a0135

jcbrand:
Hi @sloonz, it's always nice to see people getting their hands dirty and adding the changes they want in the source code.

I think it's a useful feature, so please make a PR with your changes.

Some initial thoughts on your your implementation:

1. This will only work for MUCs where real JIDs are exposed, so a check needs to be added for that. In the context of the `OccupantModal` which you've already modified, you can do something like this:
```javascript
const muc = this.model.collection.chatbox;
if (muc.features.get('nonanonymous')) {
   // Real JIDs are exposed, so you can add occupants to your roster
}
```
2. Currently when you add a roster contact, you get presented a modal with the option to choose a `name` and `group`.  We'd want to keep that ability when adding from a MUC as well, so when `Add to contacts` is clicked, the `Add a contact` modal should be shown (with the `XMPP Address` and `Name` inputs pre-filled) so that the user can still set their own values for `Name` and `Group`.

Opening a new modal is fairly simple, when the `Add to contacts` button is clicked, you can open the next modal like this:
```javascript
api.modal.show('converse-add-contact-modal', {'model': new Model()}, ev);
```

sloonz:
Done. I also hide the button for the user (so you can't add yourself). The button is also hidden for moderators / owners in anonymous rooms even tho they can see the JID of the occupant, not sure it’s the correct choice.

deleolajide:
@sloonz Could you please consider add another button to start a chat with contact. Sometime, I just want to chat and don't need to add the contact to my roster.

I currently do that in Pade direct from the participants list and the additional button in the dialog will be consistent with that. 

![image](https://user-images.githubusercontent.com/110731/199216964-117ac6af-4c47-4f70-8116-818c043aef00.png)

jcbrand:
Closing this ticket now since it's been implemented. @deleolajide You're asking for a new feature, so it should please go into a new ticket.

deleolajide:
[...]

PR #3064, opened 2022-11-01
muc: add a add to contacts button

sloonz:
Pull request for #3061

### DIFF
--- a/src/plugins/muc-views/modals/occupant.js
+++ b/src/plugins/muc-views/modals/occupant.js
@@ -1,7 +1,7 @@
 import BaseModal from "plugins/modal/modal.js";
 import tpl_occupant_modal from "./templates/occupant.js";
 import { _converse, api } from "@converse/headless/core";
-
+import { Model } from '@converse/skeletor/src/model.js';
 
 export default class OccupantModal extends BaseModal {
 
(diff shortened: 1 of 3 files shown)


src/plugins/muc-views/modals/occupant.js
```diff
@@ -36,13 +36,23 @@ export default class OccupantModal extends BaseModal {
         const role = this.model?.get('role');
         const affiliation = this.model?.get('affiliation');
         const hats = this.model?.get('hats')?.length ? this.model.get('hats') : null;
-        return tpl_occupant_modal({ jid, vcard, nick, occupant_id, role, affiliation, hats });
+        const muc = this.model.collection.chatroom;
+        const addToContacts = muc.features.get('nonanonymous') && jid && jid != _converse.bare_jid ? this.addToContacts.bind(this) : null;
```
jcbrand, line 40: This doesn't check whether the user isn't already in your roster.

You can check like this:
```javascript
const contact = await api.contacts.get(jid);
```

sloonz: is renderModal allowed to be async ?

jcbrand: Good question. The answer is "no" :)

What you can do however is pass a promise and use the `until` directive to render nothing until the promise has resolved.

https://lit.dev/docs/templates/directives/#until

jcbrand:
Thanks :)

PR #3065, opened 2022-11-01
Show nickname in status if fullname is not provided

sloonz:
Currently the logic of what’s show in the status is : 

1. vCard fullname if there is one
2. fall back to JID

This patch adds nickname between the two, so if the user has a vCard nickname but no fullname, it is shown instead of the JID : 

1. vCard fullname if there is one
2. vCard nickname if there is one
3. fall back to JID


jcbrand:
Thanks!

PR #3070, opened 2022-11-04
fixes #2925

### LINKED ISSUE #2925: File upload is not always enabled
The file download is not always activated after the user logs in.
Sometimes the download button appears and sometimes not.

**To reproduce**
On my environment, this scenario seems reproducible (but not guaranteed):
1. Open ConverseJs with a user account
2. Perform page reloads (F5) until the button is no longer displayed (skip this step if the button is already not displayed).
3. Perform page reloads and check that the button is not displayed.
4. Close the browser, saving the tab context
5. Restart the browser
6. The download button appears

**Expected behaviour**
The download button should always be displayed.

**Environment:**
 - Desktop
 - All browsers
 - Converse.js (master branch and 9.1.1, not tried on other versions)

**Additional context**
Functional ejabberd server.
<details>

```
 converse.initialize({
          'view_mode' : 'overlayed',
          'i18n' : 'a language',
          'play_sounds' : false,
          'allow_logout' : false,
          'auto_login' : true,
          'auto_reconnect' : true,
          'jid' : 'a jid',
          'clear_messages_on_reconnection' : true,
          'password' : 'a password',
          'locked_muc_nickname' : true,
          'nickname' : 'a nickname',
          'roster_groups' : false,
          'allow_registration' : false,
        });
```

</details>

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://app.bountysource.com/issues/114981626-file-upload-is-not-always-enabled?utm_campaign=plugin&utm_content=tracker%2F194169&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://app.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F194169&utm_medium=issues&utm_source=github).
</bountysource-plugin>

SilverYoCha:
fixes #2925

There is the assumption in the Converse disco code that a model can be shared between collections.

In order to avoid to break this assumption, the item collection of `entity.js` is synchronized with the one of utils.js at initialization:

https://github.com/conversejs/converse.js/blob/659a69e7b750656852e69d2fc94f75d167714239/src/headless/plugins/disco/utils.js#L65-L67

### DIFF
--- a/src/headless/plugins/disco/entity.js
+++ b/src/headless/plugins/disco/entity.js
@@ -40,6 +40,13 @@ const DiscoEntity = Model.extend({
         id = `converse.disco-items-${this.get('jid')}`;
         this.items.browserStorage = _converse.createStore(id, 'session');
         await new Promise(f => this.items.fetch({'success': f, 'error': f}));
+        _converse.disco_entities
+            .filter(e => {
+                // XXX: Ignore nodes for now.
+                // See: https://xmpp.org/extensions/xep-0030.html#items-nodes
+                return typeof e.get('name') !== 'undefined';
+            })
+            .forEach(e => this.items.create(e));
 
         this.identities = new Collection();
         id = `converse.identities-${this.get('jid')}`;
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - Add a "Add to Contacts" button in MUC occupant modals
+- #2925: Fix missing disco-items in browser storage (after CTRL+F5).
 
 ## 10.0.0 (2022-10-30)
 


jcbrand:
Hi @SilverYoCha, there are some failing tests.

For example this one:
https://github.com/conversejs/converse.js/actions/runs/3394684814/jobs/5643557257#step:4:421

```
Chrome 107.0.0.0 (Linux x86_64) Service Discovery Whenever converse.js queries a server for its features stores the features it receives FAILED
	Expected 5 to be 3.
	    at <Jasmine>
	    at src/headless/plugins/disco/tests/disco.js:166:41
	    at async UserContext.<anonymous> (src/shared/tests/mock.js:29:13)
Chrome 107.0.0.0 (Linux x86_64): Executed 5 of 510 (1 FAILED) (0 secs / 0.498 secs)
Chrome 107.0.0.0 (Linux x86_64) Service Discovery Whenever converse.js queries a server for its features stores the features it receives FAILED
	Expected 5 to be 3.
	    at <Jasmine>
	    at src/headless/plugins/disco/tests/disco.js:166:41
	    at async UserContext.<anonymous> (src/shared/tests/mock.js:29:13)
```

SilverYoCha:
Hi @jcbrand, I have finally found a solution that works in my environment and does not break the current tests.

src/headless/plugins/disco/entity.js
```diff
@@ -40,6 +40,13 @@ const DiscoEntity = Model.extend({
         id = `converse.disco-items-${this.get('jid')}`;
         this.items.browserStorage = _converse.createStore(id, 'session');
         await new Promise(f => this.items.fetch({'success': f, 'error': f}));
+        _converse.disco_entities
+            .filter(e => {
+                // XXX: Ignore nodes for now.
+                // See: https://xmpp.org/extensions/xep-0030.html#items-nodes
+                return typeof e.get('name') !== 'undefined';
+            })
+            .forEach(e => this.items.create(e));
```
jcbrand, line 49: @SilverYoCha I applaud your tenacity and dedication to finding this bug :) :clap: 

That said, I don't think this change itself is appropriate, it's a plaster over an underlying issue.

Your change here gave me a hunch as to what the issue is, and so I did some digging.

The fact that the icon consistently showed for me on Prosody but not for you on Ejabberd is to me an indication that this was due to some kind of race condition.

The change you made above avoided the race condition by immediately creating all global entities inside this entity. This is BTW wrong, since you're now letting this entity state that is has as `#items` all the entities of the global `_converse.disco_entities` collection and by extension that it supports the features of all those entities.
You would have to modify the code to first check whether the entity you're adding belongs to this current `DiscoEntity` as a sub-item and AFAIK there's no way currently to do that.

In any case, I managed to construct a test that reproduced the toolbar button not showing, even though the disco response indicated that it's supported.

Here's the fix: https://github.com/conversejs/converse.js/commit/a251608fc50390c7f46ae6ffbfbe9cd9774c8687

I'm hopeful that fix will now finally put this issue to bed.

Can you please check?

SilverYoCha: Thank you @jcbrand for your comments and for taking my suggestion into consideration.

My correction was indeed more of a band-aid than anything else. I lack a lot of knowledge in the project to be effective.

I'm glad I could put you on a serious track. Unfortunately, your commit did not solve the problem in my environment. Right after login, the download button is displayed as expected (and as it was already), but after a CTRL+F5, it is not displayed anymore.

I noticed something strange from the beginning: is it normal that I never find an entry in the session storage with the prefix id `converse.disco-items-`?

Thank you again.

jcbrand: Hi @SilverYoCha 

You have already been very effective and helped move this issue forwards.

It's taken me a long time to get an idea of what's going on because it's a complex/tricky bug, I cannot reproduce the issue with my XMPP server,  and I haven't had time to sit down and spend a long time to focus on figuring it out.

Based on what you've said before and your original MR, my current theory is that `this.items` is always empty because the entities are shared between two collections and for some reason are then persisted only to the one collection (`_converse.disco_items`) and not the other one (e.g. the MUC's entity).

I tried to verify this now, but with Prosody, the `Strophe.NS.HTTPUPLOAD` feature is on the MUC entity itself, not on one of the sub-items. I *think*, but haven't verified, that with Ejabberd, the feature is on a sub-item, and that this is the fundamental difference that is causing the bug to happen with Ejabberd and not Prosody. So with Prosody, it doesn't matter that `this.items` is always empty after a page reload because with Prosody the feature is on the MUC entity itself, not on a sub-entity.

I'm partly writing this down to also help myself remember. My day is full and I usually have 30 minutes to look at something like this before I have to do something else.

If I could get an Ejabberd account somewhere, I could verify this theory. I've asked in the Converse MUC if someone can help. I don't have time to set everything up myself.

> I noticed something strange from the beginning: is it normal that I never find an entry in the session storage with the prefix id converse.disco-items-?

I just checked, and I do see them in sessionStorage. They look like this: `converse-session/converse.disco-entities-XXXX`

You should be having entries like that as well.

Screenshot from 2022-11-10 08-33-02

SilverYoCha: If it helps, I have created an account in my environment for you @jcbrand.
I will email you the conversejs initialization settings that I use along with your credentials.

Issue #3156, opened 2023-02-27
dragresize over iframe stutters


drlight17:
**Describe the bug**
When dragresize is enabled and conversejs is working in overlayed mode there are stuttering effects while mouse dragging over the page with iframes. Before the 6.x.x version of conversejs there was no such problem.

**To Reproduce**
1. Simple conversejs initialization with embedded mode on the page with iframe(s)
2. Try to resize configbox, muc-box or chat-box and see stutters when mouse is under iframe and goes beyond the limits of conversejs div.

**Expected behavior**
No stutters

**Video**
[Recording_1677507172161.webm](https://user-images.githubusercontent.com/37434652/221586695-e858ddea-5332-455c-be30-8035a291ba39.webm)

**Environment (please complete the following information):**
 - Desktop
 - Chrome and Firefox browsers of the current last version
 - Converse.js version > 5.x.x

drlight17:
I have some fixes using jquery in roundcube-plugin on the initialization page.
See [this](https://github.com/conversejs/roundcube-plugin/commit/4224bc51ec10a8652e15173fc9a03c82ef32857f#diff-54daa1c14cadea75fd7a348bd8415daa2fbd7715a3765867f283f9721c23c108R87-R96) and [this](https://github.com/conversejs/roundcube-plugin/commit/4224bc51ec10a8652e15173fc9a03c82ef32857f#diff-54daa1c14cadea75fd7a348bd8415daa2fbd7715a3765867f283f9721c23c108R7-R16)
Maybe there are some abilities to fix this in a way more appropiate way or use some of my thoughts to fix this?

PR #3157, opened 2023-02-28
dragresize over iframe fix function concept

drlight17:
This PR fixes stuttering of conversejs overlayed dragresize UI elements over the iframes on the parent page I've mentioned in issue #3156

### DIFF
--- a/src/plugins/dragresize/index.js
+++ b/src/plugins/dragresize/index.js
@@ -67,26 +67,31 @@ converse.plugins.add('converse-dragresize', {
             document.removeEventListener('mouseup', onMouseUp);
         }
 
+        /**
+         * This function registers mousedown and mouseup events hadlers to 
+         * all iframes in the DOM when converse UI resizing events are called
+         * to prevent mouse drag stutter effect which is bad user experience.
+         * @function dragresizeOverIframeHandler
+         * @param {Object} e - dragging node element.
+         */
         function dragresizeOverIframeHandler (e) {
-            e.onmousedown = function () {
-                document.querySelector('iframe').style.pointerEvents  = 'none';
-            };
-            e.onmouseup = function () {
-                document.querySelector('iframe').style.pointerEvents  = 'initial';
-            };
+          const iframes = document.getElementsByTagName('iframe');
+          for (let iframe of iframes) {
+            e.addEventListener('mousedown', () => {
+                iframe.style.pointerEvents  = 'none';
+            }, { once: true });
+            
+            e.addEventListener('mouseup', () => {
+                iframe.style.pointerEvents  = 'initial';
+            }, { once: true });
+          }
         }
         
         api.listen.on('registeredGlobalEventHandlers', registerGlobalEventHandlers);
         api.listen.on('unregisteredGlobalEventHandlers', unregisterGlobalEventHandlers);
         api.listen.on('beforeShowingChatView', view => view.initDragResize().setDimensions());
-        api.listen.on('startDiagonalResize', function (e) {
-            dragresizeOverIframeHandler(e);
-        });
-        api.listen.on('startHorizontalResize', function (e) {
-            dragresizeOverIframeHandler(e);
-        });
-        api.listen.on('startVerticalResize', function (e) {
-            dragresizeOverIframeHandler(e);
-        });
+        api.listen.on('startDiagonalResize', dragresizeOverIframeHandler);
+        api.listen.on('startHorizontalResize', dragresizeOverIframeHandler);
+        api.listen.on('startVerticalResize', dragresizeOverIframeHandler);
     }
 });
--- a/src/plugins/dragresize/index.js
+++ b/src/plugins/dragresize/index.js
@@ -67,8 +67,26 @@ converse.plugins.add('converse-dragresize', {
             document.removeEventListener('mouseup', onMouseUp);
         }
 
+        function dragresizeOverIframeHandler (e) {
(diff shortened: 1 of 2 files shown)


src/plugins/dragresize/index.js
```diff
@@ -67,8 +67,26 @@ converse.plugins.add('converse-dragresize', {
...
+                document.querySelector('iframe').style.pointerEvents  = 'initial';
+            };
+        }
+        
         api.listen.on('registeredGlobalEventHandlers', registerGlobalEventHandlers);
         api.listen.on('unregisteredGlobalEventHandlers', unregisterGlobalEventHandlers);
         api.listen.on('beforeShowingChatView', view => view.initDragResize().setDimensions());
+        api.listen.on('startDiagonalResize', function (e) {
+            dragresizeOverIframeHandler(e);
+        });
```
jcbrand, line 84: Given that `dragresizeOverIframeHandler` doesn't reference `this`, there's no need to wrap it inside an anonymous function.

You can just do this: `api.listen.on('startDiagonalResize', dragresizeOverIframeHandler);`

src/plugins/dragresize/index.js
```diff
@@ -67,8 +67,26 @@ converse.plugins.add('converse-dragresize', {
             document.removeEventListener('mouseup', onMouseUp);
         }
 
+        function dragresizeOverIframeHandler (e) {
+            e.onmousedown = function () {
+                document.querySelector('iframe').style.pointerEvents  = 'none';
+            };
+            e.onmouseup = function () {
+                document.querySelector('iframe').style.pointerEvents  = 'initial';
+            };
```
jcbrand, line 76: Every time someone drags a chatbox, new `mousedown` and `mouseup` events are being registered without removing the older events. Eventually you might have dozens of such events registered.

If you use `addEventListener` you can pass `once` to register only a single event handler.

Instead of using `document.querySelector`, it's also better to use the `ancestor` utility method from Converse.
That way, you'll get the containing `iframe` of the element, and not potentially some other unrelated iframe.

And you also need to take into account the case where there is no iframe at all.

So more like this:
```
import { ancestor } from 'utils/html.js';

function dragresizeOverIframeHandler (e) {
    e.addEventListener('mousedown', () => {
        const iframe = ancestor(el, 'iframe');
        if (iframe) iframe.style.pointerEvents  = 'none';
    }, { once: true });
    
    // Other one comes here...
}
```

jcbrand: Can you also add a JSDoc comment to the `dragresizeOverIframeHandler` function to explain what it's doing and why it's necessary?

jcbrand (changes requested):
Thanks @drlight17, it's always nice when someone provides a fix for their own bug report!

src/plugins/dragresize/index.js
drlight17: That makes sense, thanks. I'll fix it.

src/plugins/dragresize/index.js
drlight17: > Can you also add a JSDoc comment to the `dragresizeOverIframeHandler` function to explain what it's doing and why it's necessary?
Yes, sure. I will a little later.

drlight17: > Instead of using `document.querySelector`, it's also better to use the `ancestor` utility method from Converse.
> That way, you'll get the containing `iframe` of the element, and not potentially some other unrelated iframe.

Hmm... I don't really understand how to use your suggested ancestor function in my case. I need to get all iframe nodes that is in the document DOM to turn their pointerEvents off. Ancestor only returns parent of the given element as I see. How is it better?

drlight17: For now I see the new function like this:
```
    function dragresizeOverIframeHandler (e) {
      var iframes = document.getElementsByTagName('iframe');
      for (let iframe of iframes) {
        e.addEventListener('mousedown', () => {
            if (iframe) iframe.style.pointerEvents  = 'none';
        }, { once: true });
        
        e.addEventListener('mouseup', () => {
            if (iframe) iframe.style.pointerEvents  = 'initial';
        }, { once: true });
      }
    }
```

jcbrand: I assumed you only cared about the containing iframe. Are you sure you need *all* iframes? If so, then what you wrote looks OK, although you don't need the `if (iframe)` checks, since in your `for` loop `iframe` will be defined.

drlight17: > I assumed you only cared about the containing iframe. Are you sure you need _all_ iframes? If so, then what you wrote looks OK, although you don't need the `if (iframe)` checks, since in your `for` loop `iframe` will be defined.

Yes, stuttering is taking place if mouse dragging is over any iframe on the parent page (tested in Roundcube and Bitrix CMS based site), not only the containing iframe. And yes, I've already removed iframe checks, thanks.
So, I will make commit soon.

drlight17:
Added c76f3bc commit

jcbrand:
@drlight17: Looks good, one last thing please. Can you add a line to `CHANGES.md`? Then it's ready to be merged.

drlight17:
> @drlight17: Looks good, one last thing please. Can you add a line to `CHANGES.md`? Then it's ready to be merged.

[Added ](https://github.com/drlight17/converse.js/commit/369f186fd7ca28b5eb46f259d13a373c6129b319)

jcbrand:
Thanks!

PR #3164, opened 2023-05-21
WIP: Block users

udanieli:
Hello @jcbrand, I was working on this feature and saw your branch. I did a little fix/refactoring on it, if you are interested here it is.

* fixed buttons aesthetics and triggered a modal close after API calls;
* removed duplicate code: use `sendBlockingStanza()` for block and unblock;
* return `true` in `handleBlockingStanza()` for handler to be invoked again;
* changed `Array.forEach()` to avoid garbage strings in `<block>` and  `<unblock>` elements.

In `handleBlockingStanza()` if you don't return `true`, the handler [won't be invoked again](https://strophe.im/strophejs/doc/1.6.0/files/strophe-umd-js.html#Strophe.Connection.addHandler).

Hope this is useful.

What about making chat window readonly to avoid sending messages to blocked contacts and receiving this error from the server?
image

### DIFF
--- a/src/headless/plugins/blocking/api.js
+++ b/src/headless/plugins/blocking/api.js
@@ -22,9 +22,10 @@ export default {
      * @private
      * @method api.refreshBlocklist
      */
-    refreshBlocklist () {
+    async refreshBlocklist () {
         log.debug("refreshing blocklist");
-        if (!this.isBlockingAvailable()) {
+        const available = await this.isBlockingAvailable();
+        if (!available) {
             log.debug("XEP-0191 NOT available, not refreshing...");
             api.trigger('blockListFetched', []);
             return
--- a/src/headless/plugins/blocking/api.js
+++ b/src/headless/plugins/blocking/api.js
@@ -4,66 +4,112 @@ import { _converse, api, converse } from '@converse/headless/core.js';
 const { Strophe, $iq, sizzle, u } = converse.env;
 
 export default {
+
+    /**
+     * Checks if XEP-0191 is supported
+     */
+    async isBlockingAvailable () {
+        const has_feature = await api.disco.supports(Strophe.NS.BLOCKING, _converse.domain);
+        if (!has_feature) {
+            log.warn("XEP-0191 not supported, no blocklist available");
+            return false;
+        }
+        log.debug("XEP-0191 available");
+        return true;
+    },
     /**
      * Retrieves the blocklist held by the logged in user at a JID by sending an IQ stanza.
-     * Saves the model variable _converse.blocked.set
      * @private
      * @method api.refreshBlocklist
      */
-    async refreshBlocklist () {
-        const features = await api.disco.getFeatures(_converse.domain);
-        if (!features?.findWhere({ 'var': Strophe.NS.BLOCKING })) {
-            return false;
+    refreshBlocklist () {
+        log.debug("refreshing blocklist");
+        if (!this.isBlockingAvailable()) {
+            log.debug("XEP-0191 NOT available, not refreshing...");
+            api.trigger('blockListFetched', []);
+            return
         }
         if (!_converse.connection) {
             return false;
         }
-
-        const iq = $iq({
-            'type': 'get',
-            'id': u.getUniqueId('blocklist'),
-        }).c('blocklist', { 'xmlns': Strophe.NS.BLOCKING });
-
-        const result = await api.sendIQ(iq).catch((e) => {
-            log.fatal(e);
-            return null;
-        });
-        if (result === null) {
-            const err_msg = `An error occured while fetching the blocklist`;
-            const { __ } = converse.env;
-            api.alert('error', __('Error'), err_msg);
-            log(err_msg, Strophe.LogLevel.WARN);
-            return false;
-        } else if (u.isErrorStanza(result)) {
-            log.error(`Error while fetching blocklist`);
-            log.error(result);
-            return false;
-        }
-
-        const blocklist = sizzle('item', result).map((item) => item.getAttribute('jid'));
-        _converse.blocked.set({ 'set': new Set(blocklist) });
-        return true;
+        log.debug("getting blocklist...");
+        return this.sendBlockingStanza( 'blocklist', 'get' );
     },
 
     /**
-     * Handle incoming iq stanzas in the BLOCKING namespace. Adjusts the global blocked_set.
+     * Handle incoming iq stanzas in the BLOCKING namespace. Adjusts the global blocked.set.
      * @private
      * @method api.handleBlockingStanza
      * @param { Object } [stanza] - The incoming stanza to handle
      */
-    handleBlockingStanza (stanza) {
-        if (stanza.firstElementChild.tagName === 'block') {
-            const users_to_block = sizzle('item', stanza).map((item) => item.getAttribute('jid'));
-            users_to_block.forEach(_converse.blocked.get('set').add, _converse.blocked.get('set'));
-        } else if (stanza.firstElementChild.tagName === 'unblock') {
-            const users_to_unblock = sizzle('item', stanza).map((item) => item.getAttribute('jid'));
-            users_to_unblock.forEach(_converse.blocked.get('set').delete, _converse.blocked.get('set'));
+    handleBlockingStanza ( stanza ) {
+        const action = stanza.firstElementChild.tagName;
+        const items = sizzle('item', stanza).map(item => item.getAttribute('jid'));
+        const msg_type = stanza.getAttribute('type');
+
+        log.debug(`handle blocking stanza Type ${msg_type} action ${action}`);
+        if (msg_type == 'result' && action == 'blocklist' ) {
+            log.debug(`resetting blocklist: ${items}`);
+            _converse.blocked.set({'set': new Set()});
+            items.forEach((item) => { _converse.blocked.get('set').add(item)});
+
+            /**
+            * Triggered once the _converse.blocked list has been fetched
+            * @event _converse#blockListFetched
+            * @example _converse.api.listen.on('blockListFetched', () => { ... });
+            */
+            api.trigger('blockListFetched', _converse.blocked.get('set'));
+            log.debug("triggered blockListFetched");
+
+        } else if (msg_type == 'set' && action == 'block') {
+            log.debug(`adding people to blocklist: ${items}`);
+            items.forEach((item) => { _converse.blocked.get('set').add(item)});
+            api.trigger('blockListUpdated', _converse.blocked.get('set'));
+        } else if (msg_type == 'set' && action == 'unblock') {
+            log.debug(`removing people from blocklist: ${items}`);
+            items.forEach((item) => { _converse.blocked.get('set').delete(item)});
+            api.trigger('blockListUpdated', _converse.blocked.get('set'));
         } else {
-            log.error('Received blocklist push update but could not interpret it.');
+            log.error("Received a blocklist push update but could not interpret it");
         }
-        // TODO: Fix this to not use the length as an update key, and
-        // use a more accurate update method, like a length-extendable hash
-        _converse.blocked.set({ 'len': _converse.blocked.get('set').size });
+        return true;
+    },
+
+    /**
+     * Send block/unblock IQ stanzas to the server for the JID specified
+     * @method api.sendBlockingStanza
+     * @param { String } action - "block", "unblock" or "blocklist"
+     * @param { String } iq_type - "get" or "set"
+     * @param { Array } [jid_list] - (optional) The list of JIDs to block or unblock
+     */
+    async sendBlockingStanza ( action, iq_type = 'set', jid_list = [] ) {
+        if (!_converse.connection) {
+            return false;
+        }
+
+        const element = Strophe.xmlElement(action, {'xmlns': Strophe.NS.BLOCKING});
+        jid_list.forEach((jid) => {
+            const item = Strophe.xmlElement('item', { 'jid': jid });
+            element.append(item);
+        });
+
+        const iq = $iq({
+                'type': iq_type,
+                'id': u.getUniqueId(action)
+            }).cnode(element);
+
+        const result = await api.sendIQ(iq).catch(e => { log.fatal(e); return false });
+        const err_msg = `An error occured while trying to ${action} user(s) ${jid_list}`;
+        if (result === null) {
+            api.alert('error', __('Error'), err_msg);
+            log(err_msg, Strophe.LogLevel.WARN);
+            return false;
+        } else if (u.isErrorStanza(result)) {
+            log.error(err_msg);
+            log.error(result);
+            return false;
+        }
+        return true;
     },
 
     /**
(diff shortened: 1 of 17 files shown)


kickvicky (approved):

udanieli:
I did some step forward on the blocklist: as you can see, it's a work in progress.

Sorry for not adding tests (yet), I am more of a backend guy, not a JS expert. I have just seen how you did elsewhere in the codebase and I am trying to do the same. Moreover, I learnt Javascript in the Stone Age so I must acquaint myself with all the evolutions of the language.

Any advice is welcome.

src/headless/plugins/blocking/api.js
```diff
@@ -4,97 +4,97 @@ import { _converse, api, converse } from '@converse/headless/core.js';
 const { Strophe, $iq, sizzle, u } = converse.env;
 
 export default {
+
+    /**
+     * Checks if XEP-0191 is supported
+     */
+    isBlockingAvailable () {
+        if (!_converse.disco_entities.get(_converse.domain)?.features?.findWhere({'var': Strophe.NS.BLOCKING})) {
```
jcbrand, line 12: The `api.disco.supports` function should be used to check whether a feature is supported.

That is because it makes sure that the relevant disco info has been queried first, before checking for the feature. This is also means that `isBlockingAvailable` must be `async`.

Using `api.disco.supports` to check whether the feature is supported was one of the main changes I made in this commit:
https://github.com/conversejs/converse.js/pull/3164/files

jcbrand (changes requested):

jcbrand:
Hi @udanieli, thanks a lot for your contribution.

I had a quick look and left a review comment.

Can you please also squash all your commits into a single commit? This will make it much easier to incorporate your changes.

In the branch you were referring to, [conversejs:jcbrand/block-users](https://github.com/conversejs/converse.js/tree/jcbrand/block-users), I was attempting to clean up and improve the code contributed by someone else.

To be honest, there are still quite a few things to improve there (beyond what you've already done in this MR), but looks like I ran out of steam and left the branch in an incomplete state.

Once you've addressed the review comment and squashed your commits, I'll run this PR's branch locally and take a close look (and probably make more follow-up changes).

jcbrand:
> I have also added a blocking-views plugin to display blocked users in the controlbox instead of the profile

Having a permanent list of blocked users in the control box does not sound like good UI/UX to me.

Perhaps there's a better place to put them. For example in a new modal that can be opened from the control box. Similar to the modal that shows MUC bookmarks.

> > Should I add that commit to this PR?

I think putting it in a separate PR would keep things cleaner and more manageable.

> Maybe I am missing translations too. Should I update the .pot file and all the .po ones?

You can run `make po && make po` to regenerate pot and po files, but I wouldn't do it as part of this PR because it will add lots of new files to the diff.

It would be best to do it after the merge.

src/headless/plugins/blocking/api.js
```diff
@@ -4,97 +4,102 @@ import { _converse, api, converse } from '@converse/headless/core.js';
...
      * @private
      * @method api.refreshBlocklist
      */
-    async refreshBlocklist () {
-        const features = await api.disco.getFeatures(_converse.domain);
-        if (!features?.findWhere({ 'var': Strophe.NS.BLOCKING })) {
-            return false;
+    refreshBlocklist () {
+        log.debug("refreshing blocklist");
+        if (!this.isBlockingAvailable()) {
```
jcbrand, line 27: `isBlockingAvailable` is now `async`, so you need to `await` here.

src/headless/plugins/blocking/api.js
```diff
@@ -4,97 +4,102 @@ import { _converse, api, converse } from '@converse/headless/core.js';
...
 
 export default {
+
+    /**
+     * Checks if XEP-0191 is supported
+     */
+    async isBlockingAvailable () {
+        const has_feature = await api.disco.supports(Strophe.NS.BLOCKING, _converse.domain);
+        if (!has_feature) {
+            log.warn("XEP-0191 not supported, no blocklist available");
```
jcbrand, line 14: This can rather be `log.info` than warn.

src/plugins/profile/templates/profile_modal.js
```diff
@@ -57,7 +57,9 @@ export default (el) => {
         </li>`
     );
 
-    if (_converse.pluggable.plugins['converse-blocking']?.enabled(_converse)) {
+    const blocking_available = (_converse.pluggable.plugins['converse-blocking']?.enabled(_converse) && api.isBlockingAvailable());
```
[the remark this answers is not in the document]
udanieli: Wouldn't this lead to an infinite chain of edits of `await` and `async` everywhere?

Maybe it's better to remove that logic from the template, and do that with a listener on the `blocklistFetched` and `blocklistUpdated` events, so the interface is rendered/updated according to a changed property of the LitElement.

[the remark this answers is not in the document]
udanieli: I made a little change, don't know if it's enough.

src/headless/plugins/blocking/api.js
```diff
@@ -4,97 +4,103 @@ import { _converse, api, converse } from '@converse/headless/core.js';
...
      */
     async refreshBlocklist () {
-        const features = await api.disco.getFeatures(_converse.domain);
-        if (!features?.findWhere({ 'var': Strophe.NS.BLOCKING })) {
-            return false;
+        log.debug("refreshing blocklist");
+        const available = await this.isBlockingAvailable();
+        if (!available) {
+            log.debug("XEP-0191 NOT available, not refreshing...");
+            api.trigger('blockListFetched', []);
```
jcbrand, line 30: Is triggering this event necessary here? Strictly speaking, I don't think it's correct, because the block list wasn't fetched (since blocking isn't supported).

Looks like you should also `return` in this `if` statement.

src/headless/plugins/blocking/api.js
```diff
@@ -4,97 +4,103 @@ import { _converse, api, converse } from '@converse/headless/core.js';
...
+            return
         }
         if (!_converse.connection) {
```
jcbrand, line 33: This should be called (if at all) before `this.isBlockingAvailable()` since the latter will make use of `_converse.connection`.

src/headless/plugins/blocking/api.js
```diff
@@ -4,97 +4,103 @@ import { _converse, api, converse } from '@converse/headless/core.js';
...
+        log.debug("XEP-0191 available");
+        return true;
+    },
     /**
      * Retrieves the blocklist held by the logged in user at a JID by sending an IQ stanza.
-     * Saves the model variable _converse.blocked.set
      * @private
      * @method api.refreshBlocklist
      */
     async refreshBlocklist () {
```
jcbrand, line 25: The API function names don't conform to the convention used in Converse.

The API is broken up into namespaces with short function names.

Look for example at the `disco` API: https://github.com/conversejs/converse.js/blob/master/src/headless/plugins/disco/api.js

There is a `disco` namespace, and then functions inside it.

The same needs to be done here. A `blocklist` namespace needs to be added, and then the functions here should be put inside that namespace and renamed.

So `api.isBlockingAvailable()` becomes `api.blocklist.available()`.
`api.blockUser()` becomes `api.blocklist.add()` and `api.unblockUser()` becomes `api.blocklist.remove()`.

And so forth.

src/headless/plugins/blocking/api.js
```diff
@@ -4,97 +4,103 @@ import { _converse, api, converse } from '@converse/headless/core.js';
...
+     * @method api.sendBlockingStanza
+     * @param { String } action - "block", "unblock" or "blocklist"
+     * @param { String } iq_type - "get" or "set"
+     * @param { Array } [jid_list] - (optional) The list of JIDs to block or unblock
      */
-    async blockUser (jid_list) {
-        if (!_converse.disco_entities.get(_converse.domain)?.features?.findWhere({ 'var': Strophe.NS.BLOCKING })) {
-            return false;
-        }
+    async sendBlockingStanza ( action, iq_type = 'set', jid_list = [] ) {
```
jcbrand, line 86: I think this function should be removed out of the `api.js` module and put inside `utils.js`.

It's a utility method used by API methods, but not an API method itself.

src/plugins/profile/templates/profile_modal.js
```diff
@@ -57,7 +57,7 @@ export default (el) => {
         </li>`
     );
 
-    if (_converse.pluggable.plugins['converse-blocking']?.enabled(_converse)) {
+    if (el.isBlockingAvailable()) {
```
jcbrand, line 60: This function is `async` so `await` must be used.

src/plugins/profile/templates/profile_modal.js
```diff
@@ -133,7 +133,7 @@ export default (el) => {
                 ${ el.tab === 'passwordreset' ? html`<converse-change-password-form></converse-change-password-form>` : '' }
             </div>
 
-            ${ _converse.pluggable.plugins['converse-blocking']?.enabled(_converse) ? tplBlockUsersPage(el) : '' }
+            ${el.isBlockingAvailable() ? tplBlockUsersPage(el) : '' }
```
jcbrand, line 136: Same here. This function is `async` so `await` must be used.

jcbrand (changes requested):

src/headless/plugins/blocking/api.js
udanieli: What about `_converse.blocked`? Should we put that in a namespace too, even if it's not a method?
Like `_converse.disco._identities`, should be `_converse.blocking.blocked`

src/headless/plugins/blocking/api.js
jcbrand: Yes, any function that's not being exposed as part of the public API should be moved to `utils.js`.

src/headless/plugins/blocking/api.js
jcbrand: :+1:  Yeah, it's better to namespace it to avoid potential name clashes with other (e.g. 3rd party) plugins.

jcbrand:
Hi @udanieli, you've done great work so far, thank you!

How are you feeling so far?

Are you enjoying working on Converse and motivated to continue? Or are you becoming tired?

I have more feedback on this PR, but I want to be respectful of your time.

udanieli:
> Hi @udanieli, you've done great work so far, thank you!

Thank you for sharing Converse with the community.

> How are you feeling so far?
> 
> Are you enjoying working on Converse and motivated to continue? Or are you becoming tired?

Well, I have never written so much Javascript in my life, but these are the times we live in. Working with something I have not fully mastered is sometimes frustrating, but all in all I am happy with the result.

I am still working on Converse, at the moment I am extending the `pubsub` plugin with the `subscribe()` method and a listener for the notifications.

> 
> I have more feedback on this PR, but I want to be respectful of your time.

All your reviews are welcome.

src/headless/plugins/blocking/utils.js
```diff
@@ -1,8 +1,87 @@
...
+        const item = Strophe.xmlElement('item', { 'jid': jid });
+        element.append(item);
+    });
+
+    const iq = $iq({
+            'type': iq_type,
+            'id': u.getUniqueId(action)
+        }).cnode(element);
+
+    const result = await api.sendIQ(iq).catch(e => { log.fatal(e); return false });
```
jcbrand, line 75: I think it might be better to here handle the `result` stanza by calling `handleBlockingStanza(result)` instead of doing it via `addHandler` on line 46 (although we'd still need `addHandler` for `set` IQ stanzas).

We can then use `await` to wait for the result stanza to be processed and then in the API methods we do `return sendBlockingStanza` so that the promise is returned to the caller.

The advantage of doing it like this is that the caller of the API method can decide to `await` for the operation to have succeeded and thereby be sure that the blocklist has been updated accordingly.

src/headless/plugins/blocking/utils.js
```diff
@@ -1,8 +1,87 @@
...
+export async function sendBlockingStanza ( action, iq_type = 'set', jid_list = [] ) {
+    if (!_converse.connection) {
+        return false;
+    }
+
+    const element = Strophe.xmlElement(action, {'xmlns': Strophe.NS.BLOCKING});
+    jid_list.forEach((jid) => {
+        const item = Strophe.xmlElement('item', { 'jid': jid });
+        element.append(item);
+    });
```
jcbrand, line 68: `{ 'jid' : jid }` can be shortened to `{ jid }` and we can make this a one-liner.

`jid_list.forEach((jid) => element.append(Strophe.xmlElement('item', { jid }));`

src/headless/plugins/chat/model.js
```diff
@@ -78,9 +79,24 @@ const ChatBox = ModelWithContact.extend({
          * @example _converse.api.listen.on('chatBoxInitialized', model => { ... });
          */
         await api.trigger('chatBoxInitialized', this, {'Synchronous': true});
+
+        const blocking_supported = await api.blocking?.supported();
+        if (blocking_supported) {
+            await api.waitUntil("blockListFetched");
+            this.checkIfContactBlocked(api.blocking.blocklist());
```
jcbrand, line 86: `await api.waitUntil("blockListFetched")` should be called inside `api.blocking.blocklist()`.

We don't want consumers of our API to need to know that they have to first wait for the `blockListFetched` promise. That's an implementation detail that we need to handle inside the API.

src/headless/plugins/chat/model.js
```diff
@@ -78,9 +79,24 @@ const ChatBox = ModelWithContact.extend({
...
+          api.listen.on('blockListUpdated', this.checkIfContactBlocked, this);
+        } else this.set({'contact_blocked': undefined});
```
jcbrand, line 88: Please add curly brackets here.

src/headless/plugins/chat/model.js
```diff
@@ -78,9 +79,24 @@ const ChatBox = ModelWithContact.extend({
...
+
         this.initialized.resolve();
     },
 
+    checkIfContactBlocked (jid_set) {
+        if (jid_set.has(this.get('jid')))
+            return this.set({'contact_blocked': true});
```
jcbrand, line 95: As a general rule, we always use curly brackets for `if` statements, unless it's a single `if` (no `else`) one a single line.

src/headless/plugins/chat/model.js
```diff
@@ -1104,7 +1120,7 @@ const ChatBox = ModelWithContact.extend({
             } else if (
                 this.isHidden() || (
                     pluggable.plugins['converse-blocking'] &&
-                    api.blockedUsers()?.has(message?.get('from_real_jid'))
+                    api.blocking?.blocklist().has(message?.get('from_real_jid'))
```
jcbrand, line 1123: This will have to be updated once `api.blocking.blocklist()` is `async`.

jcbrand (changes requested):
Hi @udanieli 

As promised I left more review comments.

Tests aren't running for this PR, I think because you're not basing it on `master` but a different branch instead.

Have you been running tests locally to see whether they're still green?

jcbrand:
@udanieli I think I've figured out how we can run the CI tests against your code.

You'll have to make a new branch in this repo (and not in your fork of this repo) and then make a new pull request.

Then, I can manually trigger a tests workflow against that branch. I however can't do it for branches that are not in this repo, like your fork.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

Neustradamus:
@udanieli: Have you progressed on it?

jcbrand:
Closing in favor of #3574

PR #3272, opened 2023-09-25
Fixes in overlayed mode overflow issue in chat box in home page

### LINKED ISSUE #3133: In overlayed mode the input form icons scroll with the page
**Describe the bug**
In overlayed mode, scrolling causes the input icons on the login form to scroll outside the inputs

**To Reproduce**
1. Go to converse.js
2. Click on Toggle to open the chat
3. Move mouse to main page and scroll
4. See icons move on the chat window form

(notice that they also scroll off the inputs, but stay on the form, if you ONLY scroll the form itself and not just the page)

**Expected behavior**
Icons should stay inside the inputs

**Video**

https://user-images.githubusercontent.com/1156100/216807126-bbf09c6c-7d3e-499f-b67b-900b8e1249d2.mp4

 - Desktop with overlayed mode
 - Browser Chrome latest
 - Converse.js 8

love-sonkar:
Fixes #3133 issue changes overflow-y to none

- [Y] Add a changelog entry for your change in `CHANGES.md`
- [Y] When adding a configuration variable, please make sure to
- [Y] Please add a test for your change. Tests can be run in the commandline

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,10 +1,10 @@
 # Changelog
 
 ## 11.0.0 (Unreleased)
+- #fixes #3133 changes _controlbox.scss to fix the overflow issue
 
 Breaking changes:
 - Remove the old `_converse.BootstrapModal` in favor of `_converse.BaseModal` which is a web component.
-- #fixes #3133 changes _controlbox.scss to fix the overflow issue
 
 ## 10.1.6 (2023-08-31)
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,7 @@
 
 Breaking changes:
 - Remove the old `_converse.BootstrapModal` in favor of `_converse.BaseModal` which is a web component.
+- #fixes #3133 changes _controlbox.scss to fix the overflow issue
 
 ## 10.1.6 (2023-08-31)
 
--- a/src/plugins/controlbox/styles/_controlbox.scss
+++ b/src/plugins/controlbox/styles/_controlbox.scss
@@ -53,6 +53,7 @@
     #controlbox {
         order: -1;
         color: var(--controlbox-text-color);
+        overflow-y:none;
 
         .chat-status--avatar {
             border: 1px solid var(--controlbox-pane-background-color);
--- a/src/plugins/controlbox/styles/_controlbox.scss
+++ b/src/plugins/controlbox/styles/_controlbox.scss
@@ -53,7 +53,7 @@
     #controlbox {
         order: -1;
         color: var(--controlbox-text-color);
-        overflow-y:none;
+        overflow-y:hidden;
 
         .chat-status--avatar {
             border: 1px solid var(--controlbox-pane-background-color);
--- a/src/plugins/controlbox/styles/controlbox-head.scss
+++ b/src/plugins/controlbox/styles/controlbox-head.scss
@@ -5,7 +5,7 @@
             flex-direction: row-reverse;
             flex-wrap: nowrap;
             justify-content: space-between;
-            min-height: 30px;
+            min-height: 0;
 
             .brand-heading {
                 color: var(--controlbox-text-color);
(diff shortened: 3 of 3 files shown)


CHANGES.md
```diff
@@ -4,6 +4,7 @@
 
 Breaking changes:
 - Remove the old `_converse.BootstrapModal` in favor of `_converse.BaseModal` which is a web component.
+- #fixes #3133 changes _controlbox.scss to fix the overflow issue
```
jcbrand, line 7: This isn't a breaking change, so shouldn't go under the `Breaking changes` heading.

src/plugins/controlbox/styles/_controlbox.scss
```diff
@@ -53,6 +53,7 @@
     #controlbox {
         order: -1;
         color: var(--controlbox-text-color);
+        overflow-y:none;
```
jcbrand, line 56: Is `none` a valid value?

It's not shown here as one: https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-y

What's the effect of this? Does it disable vertical scrolling in the controlbox?
If so, I'm not convinced that this is the right solution, since the controlbox might go off-screen on certain displays.

Did you check what happens in fullscreen mode? Might be that this bug should only be fixed for overlayed view mode, in which case you need to put any CSS changes under the following:

```
.conversejs {
    converse-chats {
        &.converse-overlayed {
            #controlbox {
```

src/plugins/controlbox/styles/controlbox-head.scss
```diff
@@ -5,7 +5,7 @@
             flex-direction: row-reverse;
             flex-wrap: nowrap;
             justify-content: space-between;
-            min-height: 0;
+            min-height: 30px;
```
jcbrand, line 8: Why was this necessary?

jcbrand (commented):
Thanks for your pull request. I've left some review comments.

CHANGES.md
love-sonkar: okay i'm new here so i don't know where i put those changes list. can you tell me whare i write these changes

src/plugins/controlbox/styles/_controlbox.scss
love-sonkar: yeah it's valid value it's work like hidden property but i can changes to the hidden for convinence,

i found a one more thing while checking the chat box there is a dragabble property, and we can resize the chat box but we should stop the resize property to a specific height so we can set the flyout min-height to 500px so that when user try to resize this to 400px it won't be an we can't see any scrollbar this is my approch to solve this bug.

src/plugins/controlbox/styles/controlbox-head.scss
love-sonkar: sorry i haven't check well i forgot to check properly now it's not needed to set min-height to 30px because the chat box is resizable box so it won't affect

CHANGES.md
jcbrand: Just move it up above the `Breaking changes` heading (but below `# 11.0.0 (Unreleased)`).

src/plugins/controlbox/styles/_controlbox.scss
jcbrand: > i can changes to the hidden for convinence

Yes, `hidden` would be better.

love-sonkar:
hey sir i have fixed the issue as you told me

src/plugins/controlbox/styles/_controlbox.scss
jcbrand: I've thought about this some more, and I don't think this is the right solution.

You're simply disabling scrolling, which means that if the contents would overflow, it would be invisible and the user cannot scroll to see it.

Also, since I don't know where those icons reported in #3133 come from, it's not clear to me what needs to be fixed, how it should be done or whether there really is anything to fix.

jcbrand:
Closing based on https://github.com/conversejs/converse.js/pull/3272#discussion_r1341013375

Thanks for the effort.

Issue #2716, opened 2021-11-09
Chat via URL does not display when you have one already open


polianomartini:
**Describe the bug**
It is not displaying a chat room started via URL when you already have a room open.

**To Reproduce**

1. Initialize Converse with `view_mode: 'fullscreen'`;
2. Open a new chat via URL `#converse/room?jid=room@domain` with an already open chat;
3. The chat is displayed in the list of conversations but it does not load automatically, the previous one is displayed.

**Expected behavior**
That open chat via URL overlaps the current conversation.

**Screenshots**
No open chat
![Screenshot_4](https://user-images.githubusercontent.com/13044849/141010811-90fbc39b-136d-4ba7-8ee9-c23271ebed79.png)

Opening the first chat via URL
![Screenshot_5](https://user-images.githubusercontent.com/13044849/141010962-af4da73b-b129-4933-9558-10efb9573b53.png)

Opening the second chat via URL with a chat already open
![Screenshot_6](https://user-images.githubusercontent.com/13044849/141010992-2298fe58-8374-42f2-8e74-99b84a66532c.png)

**Environment (please complete the following information):**
 - Desktop
 - Edge 94.0.992.50 / Chrome 95.0.4638.54
 - Converse.js version 8.0.1

Trevy05:
is this still open? I would like to try.

jcbrand:
Hi @Trevy05 

That would be great, thanks.

I think you can look at the `routeToRoom` function in utils.js

https://github.com/conversejs/converse.js/blob/605c02901aa424b79cbce79d0fda498d47c7f0bf/src/headless/plugins/muc/utils.js#L59

It gets called when the URL changes:

https://github.com/conversejs/converse.js/blob/605c02901aa424b79cbce79d0fda498d47c7f0bf/src/headless/plugins/muc/index.js#L245

I think to fix this issue you might just have to update the call `api.rooms.open` to pass a third `force` parameter to force it to open.

https://github.com/conversejs/converse.js/blob/605c02901aa424b79cbce79d0fda498d47c7f0bf/src/headless/plugins/muc/api.js#L64

iota-008:
Is this still open?

PR #3273, opened 2023-09-28
Fix issue with chat display when opening via URL

azr-arch:
This commit addresses issue #2716 where a chat opened via URL does not display when there is already an open chat. The changes ensure that the new chat overlaps the current conversation as expected.

I have made the necessary changes to fix the issue, but I was unable to add tests due to limited internet access and not being able to clone the repository. I would appreciate any assistance or guidance on how to proceed.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,7 @@ Breaking changes:
 
 - #3246: Badge color not responsive to dark theme
 - Fix a GIF rendering bug that causes a memory overflow
+- #2716: Fix issue with chat display when opening via URL
 
 ## 10.1.5 (2023-06-29)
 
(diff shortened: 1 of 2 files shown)


CHANGES.md
```diff
@@ -9,6 +9,7 @@ Breaking changes:
 
 - #3246: Badge color not responsive to dark theme
 - Fix a GIF rendering bug that causes a memory overflow
+- #2716: Fix issue with chat display when opening via URL
```
jcbrand, line 12: Please move this line to under the latest release heading `## 11.0.0 (Unreleased)`

Other than that, looks good thanks!

azr-arch: @jcbrand  I've moved the line, please let me know if there are any other changes needed.

jcbrand: Thanks, however this isn't a breaking change, so your line needs to go above the `Breaking changes` heading.

azr-arch: @jcbrand  moved it, is it correct now ?

jcbrand:
Thanks!

Issue #3033, opened 2022-09-24
Sort MUC bookmarks after muc component URL?


poVoq:
**Is your feature request related to a problem? Please describe.**
It is not uncommon to join multiple related MUCs on separate MUC component URL, like for example all the joinjabber.org chat rooms. It would be nice to have them sorted all together in the control-box under a common header.

**Describe the solution you'd like**
Similar to roster-groups, there could be collapsible group headers for the components, named after the muc-component name.

In addition it would be interesting to have a "add all" button on the join MUC dialog after discovering the available MUCs of a muc-component to quickly join an entire "space" of related MUCs.

**Additional context**
Despite not really used for such, muc-components could be used as a way to group mucs similar to how other chat systems do it with so called "spaces" or "guilds/servers". The backend is already there, but no client implements an UI for it.

licaon-kter:
"Spaces" is a separate thing that's work in progress in XMPP world, eg. Gajim has them iirc

poVoq:
Gajim has "workspaces", which are conceptually a quite different thing. They are more like client side custom filters for channels.

Although Gajim could add an option to input a muc component domain when creating a new workspace to automatically add all discovered rooms to the workspace.

licaon-kter:
I'd rather have a standardised thing that you can apply to your case, rather that a separate thing for your case that is not able to handle "other servers".

poVoq:
Exactly, MUC components URLs are the XMPP standardized way to compartmentalize channels on the server side, which is exactly what "spaces" are. I don't understand why we even seem to disagree here?

BetaRays:
I’m also interested by this.

Though I’m no web developer, I managed to change the chatroom list ordering to prioritize domain names with the following diff:
```patch
diff --git a/src/plugins/roomslist/templates/roomslist.js b/src/plugins/roomslist/templates/roomslist.js
index e20fc5276..35d6cf44e 100644
--- a/src/plugins/roomslist/templates/roomslist.js
+++ b/src/plugins/roomslist/templates/roomslist.js
@@ -70,7 +70,17 @@ function tplRoomItem (el, room) {
 export default (el) => {
     const { chatboxes, CHATROOMS_TYPE, CLOSED } = _converse;
     const rooms = chatboxes.filter(m => m.get('type') === CHATROOMS_TYPE);
-    rooms.sort((a, b) => (a.getDisplayName().toLowerCase() <= b.getDisplayName().toLowerCase() ? -1 : 1));
[...]

PR #3283, opened 2023-10-17
Muc domain grouping

BetaRays:
This implements the proposal made in https://github.com/conversejs/converse.js/issues/3033.
When the `muc_grouped_by_domain` configuration option is set to `true`, chatrooms will be displayed by groups, based on their domain.
Note that this currently shows groups even when only one channel is being followed in the domain, unlike what was proposed in https://github.com/conversejs/converse.js/issues/3033.
This would require either moving all non-grouped channels before any groups, or adding an indentation or other visual indacator that channel are in groups, but the current version is fine for my use-case.

This is still missing:
 - [x] ~~the new translatable string in the correct place (I don’t know where this is yet)~~ This isn’t actually necessary.
 - [x] defining the new color in the themes (though it looks fine on the default theme as it is)
 - [x] adding a changelog entry for the change in `CHANGES.md` ~~(I’m not sure what the format for doing this should be, as it’s not currently in any version.)~~
 - [x] any tests of the new functionality (which I would need to learn how to write)

Other possible changes:
 - use `Strophe.getDomainFromJid` instead of `.split('@').at(-1)`
 - put the code that generates `grouped_rooms` and `sorted_domains` in a separate (`utils.js`?) file/function.

Also, note that the group names use the same font as the more global togglable groups (this comes from `list-toggle`, which uses `heading-font`).

### DIFF
--- a/src/plugins/roomslist/styles/roomsgroups.scss
+++ b/src/plugins/roomslist/styles/roomsgroups.scss
@@ -1,0 +1,14 @@
+.conversejs {
+    #chatrooms {
+        .muc-domain-group-toggle {
+            margin: 0.75em 0 0.25em 0;
+        }
+
+        .muc-domain-group-toggle, .muc-domain-group-toggle .fa {
+            color: var(--groupchats-header-color) !important;
+            &:hover {
+                color: var(--chatroom-head-bg-color-dark) !important;
+            }
+        }
+    }
+}
--- a/src/plugins/roomslist/styles/roomsgroups.scss
+++ b/src/plugins/roomslist/styles/roomsgroups.scss
@@ -5,9 +5,9 @@
         }
 
         .muc-domain-group-toggle, .muc-domain-group-toggle .fa {
-            color: var(--groupchats-header-color) !important;
+            color: var(--groupchats-header-color);
             &:hover {
-                color: var(--chatroom-head-bg-color-dark) !important;
+                color: var(--chatroom-head-bg-color-dark);
             }
         }
     }
--- a/src/plugins/roomslist/templates/roomslist.js
+++ b/src/plugins/roomslist/templates/roomslist.js
@@ -67,10 +67,44 @@ function tplRoomItem (el, room) {
         </div>`;
 }
 
+function tplRoomDomainGroup (el, domain, rooms) {
+    const i18n_title = __('Click to hide these rooms');
+    const collapsed = el.model.get('collapsed_domains');
+    const is_collapsed = collapsed.includes(domain);
+    return html`
+    <div class="muc-domain-group" data-domain="${domain}">
+        <a href="#" class="list-toggle muc-domain-group-toggle controlbox-padded" title="${i18n_title}" @click=${ev => el.toggleDomainList(ev, domain)}>
+            <converse-icon
+                class="fa ${ is_collapsed ? 'fa-caret-right' : 'fa-caret-down' }"
+                size="1em"
+                color="var(--muc-group-color)"></converse-icon>
+            ${domain}
+        </a>
+        <ul class="items-list muc-domain-group ${ is_collapsed ? 'collapsed' : '' }" data-domain="${domain}">
+            ${ rooms.map(room => tplRoomItem(el, room)) }
+        </ul>
+    </div>`;
+}
+
 export default (el) => {
     const { chatboxes, CHATROOMS_TYPE, CLOSED } = _converse;
+    const group_by_domain = api.settings.get('muc_grouped_by_domain');
     const rooms = chatboxes.filter(m => m.get('type') === CHATROOMS_TYPE);
     rooms.sort((a, b) => (a.getDisplayName().toLowerCase() <= b.getDisplayName().toLowerCase() ? -1 : 1));
+    // The rooms should stay sorted as they are iterated and added in order
+    const grouped_rooms = new Map();
+    if (group_by_domain) {
+        for (const room of rooms) {
+            const roomdomain = room.get('jid').split('@').at(-1).toLowerCase();
+            if (grouped_rooms.has(roomdomain)) {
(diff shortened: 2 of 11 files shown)


src/plugins/roomslist/templates/roomslist.js
```diff
@@ -67,10 +67,44 @@ function tplRoomItem (el, room) {
         </div>`;
 }
 
+function tplRoomDomainGroup (el, domain, rooms) {
```
jcbrand, line 70: This function should ideally go into a separate file. Generally we put template functions into their own files so that it's easier for 3rd party integrators to change/override them (e.g. by using webpack aliases).

src/plugins/roomslist/templates/roomslist.js
```diff
@@ -67,10 +67,44 @@ function tplRoomItem (el, room) {
...
+            const roomdomain = room.get('jid').split('@').at(-1).toLowerCase();
+            if (grouped_rooms.has(roomdomain)) {
+                grouped_rooms.get(roomdomain).push(room);
+            } else {
+                grouped_rooms.set(roomdomain, [room]);
+            }
+        }
+    }
+    const sorted_domains = Array.from(grouped_rooms.keys());
+    sorted_domains.sort();
```
jcbrand, line 107: Would be a bit nicer if these lines 94 to 107 are put in a function that is lazily called on line 149.

It avoids unnecessarily creating a `Map` (i.e. when `group_by_domain` is `false`) and keeps the template function smaller.

jcbrand (commented):
Thanks @BetaRays, looks really good already.

I'm always happy when people are able to figure out things themselves and write good idiomatic code.

> the new translatable string in the correct place (I don’t know where this is yet)

You don't have to do anything else, just wrapping the text with `__()` is enough.
Running `make pot` will add the string to the `.pot` file and then running `make po` will generate the `.po` files from that. It's probably be better to not do that in this PR though, to keep it small.

>  adding a changelog entry for the change in CHANGES.md (I’m not sure what the format for doing this should be, as it’s not currently in any version.)

Just add a line under `11.0.0 (Unreleased)`

> any tests of the new functionality (which I would need to learn how to write)

Based on what you've done so far, I think you'll be able to figure it out.
Take a look at the tests here:
https://github.com/conversejs/converse.js/blob/master/src/plugins/roomslist/tests/roomslist.js

I think you'll get the gist of it when you read the test code. You can replace `it(` with `fit(` to run only that single test, which makes things easier.

I generally run `make watch` in one terminal, and then `make tests` in another terminal.

Ideally the tests should go into a new file in `src/plugins/roomslist/tests`.

If you create such a file, you need to add it to add it to karma.conf.js
https://github.com/conversejs/converse.js/blob/master/karma.conf.js

BetaRays:
I added some tests, though I’m not sure I did them in the best way.
In particular, this is relying on timeouts, and while this shouldn’t be a big problem in this case, I would have liked to avoid this if possible (it seems that some emoji picker tests fail on my old computer because of “too short” timeouts).
Also, as mentioned in the comments, I couldn’t make `u.isVisible` work in a specific case, and I think it would have been a better choice if it worked.

jcbrand:
I had a quick look at the tests and they look fine to me. With "timeouts" I guess you mean the `waitUntil` function?
We use that in most tests because of the asynchronous nature of the UI updating based on state changes. They can be a bit brittle sometimes, but for the most part they work well enough.

Please just fix the tests to use `it` and not `fit` and fix the merge conflict in CHANGES.md (e.g. you can do a `git pull rebase master` locally and then re-push).

Once that's done we can run CI and check if it's green.

BetaRays:
> I had a quick look at the tests and they look fine to me. With "timeouts" I guess you mean the `waitUntil` function? We use that in most tests because of the asynchronous nature of the UI updating based on state changes. They can be a bit brittle sometimes, but for the most part they work well enough.

That’s what I meant, yes. It’s not a big problem in this case but I wanted to make sure that there wasn’t another way.

> Please just fix the tests to use `it` and not `fit` and fix the merge conflict in CHANGES.md (e.g. you can do a `git pull rebase master` locally and then re-push).
> 
> Once that's done we can run CI and check if it's green.

The tests should use `it` since 83f8df55eb19405daed28448880b182b9a420ba5, you might have looked at the previous commit.

I’ll do a rebase on `master`.

jcbrand:
That was a mistake, can you please remove it?

BetaRays:
I tried to add some SCSS to make the display/colors more consistent with roster groups.
Please tell me if there’s a better way to import it, a different way to define things that would be easier to theme, or anything else I might have forgotten.

src/plugins/roomslist/styles/roomsgroups.scss
```diff
@@ -0,0 +1,14 @@
+.conversejs {
+    #chatrooms {
+        .muc-domain-group-toggle {
+            margin: 0.75em 0 0.25em 0;
+        }
+
+        .muc-domain-group-toggle, .muc-domain-group-toggle .fa {
+            color: var(--groupchats-header-color) !important;
+            &:hover {
+                color: var(--chatroom-head-bg-color-dark) !important;
```
jcbrand, line 10: Is the `!important` directive really necessary? We try to avoid using it unless absolutely necessary, which is rare.

BetaRays: I did it to copy what was done for [`group-toggle`](https://github.com/conversejs/converse.js/blob/master/src/plugins/rosterview/styles/roster.scss#L86) and [`open-rooms-toggle`](https://github.com/conversejs/converse.js/blob/master/src/plugins/muc-views/styles/controlbox.scss#L17), but it does seem to work without it, so I’ll remove it.

jcbrand:
Thanks @BetaRays, great work.

PR #3301, opened 2024-01-16
Fix #3300: Adding the maxWait option for `debouncedPruneHistory`

### LINKED ISSUE #3300: Prune history under heavy flooding does never happen
**Describe the bug**

If you are in a room where users are heavily flooding, even if you set ConverseJS to prune history, it will never happen.
At the end, the browser will use 100% CPU, and sometimes even crash.

This is because of this line:

https://github.com/conversejs/converse.js/blob/59753231b6f8e2a65a893015cb8f1c7c2efe8cc7/src/headless/shared/chat/utils.js#L111

```javascript
export const debouncedPruneHistory = debounce(pruneHistory, 500);
```

Unless there is a 500ms window where no one is talking, pruning will never happen.
In fact, it can be worse: the browser can start to lag, and continue to process message several dozen of seconds after they were sent. So, it will need a long time with no message to be able to prune.

I'll propose a PR with a fix: adding the `maxWait` option, to force pruning after a few seconds.

**To Reproduce**
1. Join a room
2. Flood the room (using bots for example)
3. See the browser using ConverseJS CPU usage

**Screenshots**

I have some testing tools. Here are the results i got.

Test conditions:
* 1 browser, ConverseJS with prune history set to keep 100 messages
* 1 bot emitting messages every 100ms
* 1 bot emitting messages every 150ms
* 1 bot emitting messages every 1000ms, but who will continue to speak longer

Here is a graph representing the browser CPU usage when joining a room where bots are flooding:

image

After adding a `maxWait: 2000`:

image

As you can see, in first case the CPU raise to 100%, inducing some lag (the CPU is still at 100% 50 seconds after bot stops talking).
With the fix, the CPU never reach 100%.

JohnXLivingston:
Adding `maxWait: 2000` to force history pruning after 2 seconds max.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -5,6 +5,7 @@
 - #2716: Fix issue with chat display when opening via URL
 - #3033: Add the `muc_grouped_by_domain` option to display MUCs on the same domain in collapsible groups
 - Add an occupants filter to the MUC sidebar
+- #3300: Adding the maxWait option for `debouncedPruneHistory`
 
 ### Breaking changes:
 
--- a/src/headless/shared/chat/utils.js
+++ b/src/headless/shared/chat/utils.js
@@ -108,4 +108,6 @@ export async function handleCorrection (model, attrs) {
 }
 
 
-export const debouncedPruneHistory = debounce(pruneHistory, 500);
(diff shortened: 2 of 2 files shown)


jcbrand:
Thanks, merged manually.

PR #3304, opened 2024-01-17
Fix: MUC occupant list does not sort itself on nicknames or roles changes

JohnXLivingston:
This fix was discussed with @jcbrand  on the ConverseJS chat room.
The Collection object does not sort itself on item changes.
So we manually add a call to sort when nickname or role of a MUC participant change.


jcbrand:
Thanks :pray:, merged manually.

Issue #1195, opened 2018-09-10
Add icon to quote a message


contrapunctus-1:
~Current behavior -~
~1. Type something in the entry box without sending it~
~2. Press Up enough times to switch edit/reuse a previous message (you aren't likely to do this after Step 1, but it often happens by accident, especially in multi-line messages)~
~OR~
~Press Down 1-2 times when you're on the last line (more likely to happen)~
~3. Press Down to get back to the message you were typing - the message you were typing has vanished and you have a blank box again.~

~That's really frustrating, especially if you were typing a big message :)~

~Suggestion - if you press Up, restore the draft when you get back to it via pressing Down. Don't wipe the message box when you press Down with the cursor at the end of the last line.~

Complementary improvement - just as hovering over messages displays a pencil button to edit, hovering * should also display a quote button to quote that message. (* And perhaps also when selecting some text.)

Nyco:
Isn't this one related? https://github.com/conversejs/converse.js/issues/1142

jcbrand:
Yes, this is a duplicate of #1142, although we can keep it open for the "quote message" feature you suggsted.

PR #3336, opened 2024-02-29
Add copy and quote message actions

BetaRays:
This allows copying the contents of a message, or quoting it in the current draft (https://github.com/conversejs/converse.js/issues/1195).

 - [ ] As this adds new menu actions, it’s easy to get a longer menu (Edit, Retract, Copy, Quote), which may drop up instead of down in the wrong cases: this may require changing [this heuristic](https://github.com/conversejs/converse.js/blob/master/src/shared/chat/message-actions.js#L53-L57). Capture d’écran_2024-02-29_23-41-55


jcbrand:
Great work, thanks!

JohnXLivingston:
I have similar issues, where the dropdown is truncated. I have many custom actions in my project.

Here is a screenshot when everything is fine:
image

And here when it is truncated:
image

JohnXLivingston:
I think this will fix: https://github.com/conversejs/converse.js/pull/3474 (the heuristic seems to not be relevant anymore, maybe thanks to bootstrap v5)

Issue #3337, opened 2024-03-04
Multi-line nested quotes display incorrectly


BetaRays:
**Describe the bug**
A message such as
```
> > a
> > b
> c
d
```
or
```
> > > a
> > b
c
```
should display quote levels going multiple lines, but only the first line has multiple levels, and only the first level spans multiple lines.

**To Reproduce**
Send (or receive) one of the above blocks as a message.

**Expected behavior**
Second-level quotes (and more) should be able to span multiple lines.

**Screenshots**
![image](https://github.com/conversejs/converse.js/assets/26482956/2cb69957-849c-4be6-9e0e-e3369684b8d6)
![image](https://github.com/conversejs/converse.js/assets/26482956/11b749f5-1e28-4472-93de-4e5f30386f96)

**Environment (please complete the following information):**
 - Desktop
 - Browser: Firefox
 - Converse.js version: both v10.1.6 and current master (04c37537dd7615d0a2a7d72b4b85f95a19275f5f).

BetaRays:
Apparently this patch fixes it, but I don’t know why the zero-width spaces were there in the first place.
```patch
diff --git a/src/shared/styling.js b/src/shared/styling.js
index bf09a649a..d18e881c3 100644
--- a/src/shared/styling.js
+++ b/src/shared/styling.js
@@ -150,8 +150,7 @@ export function getDirectiveTemplate (d, text, offset, options) {
     if (isQuoteDirective(d)) {
         const newtext = text
             // Don't show the directive itself
-            .replace(/\n>\s/g, '\n\u200B\u200B')
-            .replace(/\n>/g, '\n\u200B')
+            .replace(/\n>\s?/g, '\n')
             .replace(/\n$/, ''); // Trim line-break at the end
         return template(newtext, offset, options);
     } else {
```

@jcbrand Should I open a merge request for this? Or is this a wrong solution?

BetaRays:
The zero-width spaces were added in this commit: https://github.com/conversejs/converse.js/commit/97be0bd8acee5bcdba32cbea08d74698a915f3de, but they don’t seem to be the part that fixes the linked issue.
In any case, the zero-width spaces break this line: https://github.com/conversejs/converse.js/blob/master/src/shared/styling.js#L109.

BetaRays:
I also tried this in order to keep the zero-width spaces, but that prevents the removal of `>` characters, which are added on the next lines.
```patch
diff --git a/src/shared/styling.js b/src/shared/styling.js
[...]

PR #3339, opened 2024-03-05
Correctly display multi-line nested quotes

BetaRays:
This fixes https://github.com/conversejs/converse.js/issues/3337.

Additionally, this considers the case of multiple empty lines in quotes, and correctly displaying empty lines not including trailing spaces.


jcbrand:
Thanks!

Issue #1349, opened 2018-11-23
XEP-0392: Consistent Color Generation


stevenroose:
Tracking issue for XEP-0392 compliance

- [x] make MUC nicks use the XEP-compliant color
- [ ] (bonus: make them bold)
- [x] give users without avatar a XEP-compliant colored default avatar
- [x] (precondition: add avatars to the MUC list)
- [x] give MUCs without avatar a XEP-compliant colored default avatar

stevenroose:
I think @Deleolajide (correct username?) has a plugin that partially implements the colored avatars

deleolajide:
The [plugin ](https://github.com/igniterealtime/Pade/blob/master/extension/inverse/plugins/pade.js)does many other **non standard** requested things for Pàdé. For reference, the avatar creation is done by the function below that uses a canvas element to create a data URI.

```
function createAvatar(nickname, width, height, font)
{
    if (!width) width = 32;
    if (!height) height = 32;
    if (!font) font = "16px Arial";

    var canvas = document.createElement('canvas');
    canvas.style.display = 'none';
    canvas.width = width;
    canvas.height = height;
    document.body.appendChild(canvas);
    var context = canvas.getContext('2d');
    context.fillStyle = getRandomColor(nickname);
    context.fillRect(0, 0, canvas.width, canvas.height);
    context.font = font;
    context.fillStyle = "#fff";

    var first, last;
    var name = nickname.split(" ");
    var l = name.length - 1;

    if (name && name[0] && name.first != '')
    {
        first = name[0][0];
        last = name[l] && name[l] != '' && l > 0 ? name[l][0] : null;

        if (last) {
            var initials = first + last;
            context.fillText(initials.toUpperCase(), 3, 23);
        } else {
            var initials = first;
            context.fillText(initials.toUpperCase(), 10, 23);
        }
        var data = canvas.toDataURL();
        document.body.removeChild(canvas);
    }

    return canvas.toDataURL();
}

var nickColors = {}

function getRandomColor(nickname)
{
    if (nickColors[nickname])
    {
        return nickColors[nickname];
    }
    else {
        var letters = '0123456789ABCDEF';
        var color = '#';

        for (var i = 0; i < 6; i++) {
            color += letters[Math.floor(Math.random() * 16)];
        }
        nickColors[nickname] = color;
        return color;
    }
}
```

Nyco:
Doesn't this present ticket include this other ticket #1350 ?

[...]

PR #3347, opened 2024-03-14
XEP-0392 implementation: colorize nicknames

JohnXLivingston:
This PR implements XEP-0392 (Consistent Color Generation) for message author nicknames, and for MUC occupant list.

This XEP compliance is tracked here: https://github.com/conversejs/converse.js/issues/1349

This PR does not implement the default avatar color (but it should be easy to add it).

I added a configuration settings to enable this feature (`colorize_username`). By default i set it to false, but this can be changed if you think it is worth enabling it be default.

Here are some screenshot (the ConverseJS theme is not the default one on my project, but it is close enough to have a good idea of the result):

image

image

Note: the XEP is currently in "last call for comment" state, ending the 2024-03-25


src/utils/color.js
```diff
@@ -0,0 +1,33 @@
...
+  // We cache results in `cache`, to avoid unecessary computing (as it can be called very often)
+  const v = cache.get(s);
+  if (v) return v;
+
+  // Run the input through SHA-1
+  const digest = sha1.digest(s);
+  // Treat the output as little endian and extract the least-significant 16 bits.
+  // (These are the first two bytes of the output, with the second byte being the most significant one.)
+  // Divide the value by 65536 (use float division) and multiply it by 360 (to map it to degrees in a full circle).
+  const angle = (((digest[0] & 0xFF) + ((digest[1]) & 0xFF) * 256) / 65536.0) * 360;
```
linkmauve, line 22: You don’t need the `& 0xFF` here, the `digest()` method already returns only numbers between 0 and 255.

JohnXLivingston: Right, i will fix it. Thanks.

jcbrand:
Thanks @JohnXLivingston, this is great, although if it was up to me, I'd only make default avatars coloured, and not the nicknames as well.

Personally I find the coloured nicknames quite garish. Perhaps the feature could be extended to allow either nicknames or avatars or both to be coloured.

That said, I could see someone defending colouring the nicknames as a security feature (i.e. making it more difficult to impersonate someone).

package.json
```diff
@@ -122,7 +122,9 @@
     "dompurify": "^3.0.8",
     "favico.js-slevomat": "^0.3.11",
     "gifuct-js": "^2.1.2",
+    "hsluv": "^1.0.1",
     "jed": "1.1.1",
+    "js-sha1": "^0.7.0",
```
jcbrand, line 127: Is this really necessary? Modern browsers support basic crypto primitives and operations.

https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest

src/utils/color.js
```diff
@@ -0,0 +1,33 @@
+import { sha1 } from 'js-sha1';
+import { Hsluv } from 'hsluv';
```
jcbrand, line 2: To keep the build small and avoid loading unwanted code (e.g. if `colorize_nicknames` is `false`), I think we should see if this can be loaded dynamically.

See for example here: https://github.com/conversejs/converse.js/blob/d4dfc3785b962973c8e17de9fa8f3ee7204329bc/src/headless/plugins/emoji/index.js#L80

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/import

Hopefully it just works and Webpack should know what to do (i.e. create a separate file for `Hsluv` in `./dist`).

JohnXLivingston:
> Thanks @JohnXLivingston, this is great, although if it was up to me, I'd only make default avatars coloured, and not the nicknames as well.

In my project, i generate random avatars for user that don't have avatars. So i don't need colored avatars. (but that is just my case, and as i said, it will be easy to re-use the `colorize` function to apply to default avatar SVG).

> Personally I find the coloured nicknames quite garish. Perhaps the feature could be extended to allow either nicknames or avatars or both to be coloured.

I also found it quite garish, but i got good feedbacks on Mastodon.

We could add another settings for avatars (`colorize_default_avatar`?), or change the settings to `colorize` that could get multiple values: `none`, `default_avatar`, `username`, `both`.
What do you prefer?
The second solution seems elegant, but will not work if we want to add a third object to colorize (i don't know if this will be the case one day).

> That said, I could see someone defending colouring the nicknames as a security feature (i.e. making it more difficult to impersonate someone).

Yes, and i think it will be usefull in my use case.

package.json
JohnXLivingston: ok, will fix.

src/utils/color.js
JohnXLivingston: oh, great! Will fix. I just have to check if making the `colorize` function async is possible.

jcbrand:
@JohnXLivingston I've rebased this PR to get the tests green.

PR #3363, opened 2024-04-02
Fix missing import when trying to display into modal popin an uploaded image.

SilverYoCha:
If this import is not indicated, the component is not registered.
So when it is called directly by the modal API's show method, the component is not found.


jcbrand:
Thanks

PR #3407, opened 2024-05-28
Fix custom emoji category documentation:

JohnXLivingston:
* custom default value is null.
* adding text to explain why custom category is disabled by default, and how to enable it.

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -846,15 +846,16 @@ settings, it's not possible to add new emoji categories. There is however a
 "stickers").
 
 Please note that the ``custom`` category is disabled by default, as it does not
-rely on a standard for now, and will only work for users that are using the same
-Converse.
-If you want to enable the ``custom`` category, please use ``:converse:`` as
-value.
+rely on a standard for now, and will only work for users that are using versions
+of Converse that support it.
 
-To add custom emojis, you need to edit ``src/headless/emojis.json`` to add new
-entries to the map under the  ``custom`` key.
+To enable the ``custom`` category, set it to one of the keys under the
+``custom`` section in ``src/headless/plugins/emoji/emojis.json`` (for example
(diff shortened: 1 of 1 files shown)


docs/source/configuration.rst
```diff
@@ -845,6 +845,12 @@ settings, it's not possible to add new emoji categories. There is however a
 ``custom`` category where you can put your own custom emojis (also known as
 "stickers").
 
+Please note that the ``custom`` category is disabled by default, as it does not
+rely on a standard for now, and will only work for users that are using the same
+ConverseJS.
+If you want to enable the ``custom`` category, please use ``:converse:`` as
```
jcbrand, line 851: I don't think you need to use `:converse:` as value. It can be any custom shortname that's in the emojis JSON.

Also, we don't use `ConverseJS` in the docs, only `Converse`.

JohnXLivingston: This is just for enabling the custom category, without changing the `emojis.json` content. So there is only `:converse:` and `:xmpp:` by default. (and `:converse:` was the value before the commit that disabled it).

I will correct `Converse` instead of `ConverseJS`.

jcbrand: I understand, but I don't think the text is particularly helpful, because it's not clear that you can enable `custom` with another value as well.

Something better would be:

_Please note that the ``custom`` category is disabled by default, as it does not
rely on a standard for now, and will only work for users that are using versions of Converse that support it._

_To enable the ``custom`` category, set it to one of the keys under the ``custom`` section in ``src/headless/plugins/emoji/emojis.json`` (for example ``:converse:``). You can also add new entries there, to create more custom emojis._

JohnXLivingston: Done.

jcbrand:
Thanks!

PR #3405, opened 2024-05-23
New loadEmojis hook, to customize emojis at runtime.

JohnXLivingston:

PR #3408, opened 2024-05-28
Fix custom emojis path when assets_path is not the default path.

JohnXLivingston:
Custom emojis are broken when the assets_path is not the default one.
Here is a fix. Could be done otherwise, just tell me if this is ok for you.

### DIFF
--- a/src/headless/plugins/emoji/plugin.js
+++ b/src/headless/plugins/emoji/plugin.js
@@ -15,6 +15,27 @@ converse.emojis = {
     'initialized_promise': getOpenPromise(),
 };
 
+/**
+ * The emoji.json file uses path starting with '/dist' for custom emojis.
+ * But if the assets_path settings is not the default one, these paths are wrong.
+ * This functions fixes this.
+ * @param json
+ */
+function fixCustomEmojisPath (json) {
+    let path = api.settings.get('assets_path')
+    if (path[path.length - 1] !== '/') { path = path + '/' }
+    if (path === '/dist/') {
+        return;
+    }
+    if (!('custom' in json) || (typeof json.custom !== 'object')) {
+        return;
+    }
+    for (const e of Object.values(json.custom)) {
+        if (typeof e.url !== 'string') { continue; }
+        e.url = e.url.replace(/^\/dist\//, path);
+    }
+}
+
 converse.plugins.add('converse-emoji', {
     initialize () {
         /* The initialize function gets called as soon as the plugin is
@@ -99,6 +120,7 @@ converse.plugins.add('converse-emoji', {
                          *  });
                          */
                         const json = await api.hook('loadEmojis', {}, module.default);
+                        fixCustomEmojisPath(json);
                         converse.emojis.json = json;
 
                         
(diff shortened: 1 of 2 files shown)


src/headless/plugins/emoji/plugin.js
```diff
@@ -15,6 +15,27 @@ converse.emojis = {
...
+    if (path[path.length - 1] !== '/') { path = path + '/' }
+    if (path === '/dist/') {
+        return;
+    }
+    if (!('custom' in json) || (typeof json.custom !== 'object')) {
+        return;
+    }
+    for (const e of Object.values(json.custom)) {
+        if (typeof e.url !== 'string') { continue; }
+        e.url = e.url.replace(/^\/dist\//, path);
```
jcbrand, line 35: This feels very hacky.

I think it would be better if the path in emojis.json was relative to `assets_path`, and not absolute as it's now.

That would remove the need for this function entirely.

JohnXLivingston: This could break backward compatibility.

Moreover, in my use case, i have to add emojis that are under another path.

To explain my use case: as you know, i use Converse to display chat in Peertube, a video streaming platform. I want to enable custom emojis for streamers (as twitch does for example). So i have to add customs emojis that are not stored in the asset_paths (i plan to use the new [`loadEmojis` hook](https://github.com/conversejs/converse.js/pull/3405))

jcbrand: > This could break backward compatibility.

The next release of Converse is already a backwards incompatible major release, so that's not a train smash, as long as the change is documented properly in the CHANGES.md.

> To explain my use case: as you know, i use Converse to display chat in Peertube, a video streaming platform. I want to enable custom emojis for streamers (as twitch does for example). So i have to add customs emojis that are not stored in the asset_paths.

I was referring to the default custom emojis (`:converse:` and `:xmpp:`) that should rather have relative paths, so that your `fixCustomEmojisPath` function is no longer necessary. You can still use absolute paths for your own custom emojis, right?

JohnXLivingston: > The next release of Converse is already a backwards incompatible major release, so that's not a train smash, as long as the change is documented properly in the CHANGES.md.
> 

Ok.

> I was referring to the default custom emojis (`:converse:` and `:xmpp:`) that should rather have relative paths, so that your `fixCustomEmojisPath` function is no longer necessary. You can still use absolute paths for your own custom emojis, right?

Ok for absolute path.

Unless i'm wrong, the url is directly injected as img.src attribute. So if we make it relative to asset_paths, we still need a function to check if the url is absolute or not, to prepend or not the asset_paths.
Where would you do this? Any suggestion?

jcbrand: I quickly did it in this commit:
https://github.com/conversejs/converse.js/commit/4fbb6e15caf458c6b63e4c4a2f810efced14c8f4

JohnXLivingston: Perfect! Thanks a lot

PR #3433, opened 2024-07-05
bugfixing: image modal dialog cannot be opened second time

izinin:
Bugfixing: Image modal dialog is opened only once. Next time user clicks image modal dialog is not shown

I did not update unit test for the fix verification because unit test to check showing image in modal dialog does not exist. I tested the fix manually

https://github.com/conversejs/converse.js/assets/2598832/b6e60846-60c7-44f0-b36b-fbf0e885c81c

 

- [ x] Add a changelog entry for your change in `CHANGES.md`

### DIFF
--- a/src/shared/chat/message-body.js
+++ b/src/shared/chat/message-body.js
@@ -37,6 +37,7 @@ export default class MessageBody extends CustomElement {
 
     onImgClick (ev) { // eslint-disable-line class-methods-use-this
         ev.preventDefault();
+        api.modal.remove('converse-image-modal');
         api.modal.show('converse-image-modal', {'src': ev.target.src}, ev);
     }
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ## 11.0.0 (Unreleased)
 
+- Fix: image modal dialog cannot be opened the second time because the dialog is left in undefined, not initial state
 - #1174: Show MUC avatars in the rooms list
 - #1195: Add actions to quote and copy messages
 - #1349: XEP-0392 Consistent Color Generation


src/shared/chat/message-body.js
```diff
@@ -37,6 +37,7 @@ export default class MessageBody extends CustomElement {
 
     onImgClick (ev) { // eslint-disable-line class-methods-use-this
         ev.preventDefault();
+        api.modal.remove('converse-image-modal');
```
jcbrand, line 40: Thanks for this PR. I need to look into the issue in a bit more detail. I don't think this is the right approach, ideally the modal should be removed when closed (not re-opened) or it should be re-used without having to close it first.

I'll take a look once I have some time available.

jcbrand: I'm not able to reproduce this bug. I tried multiple times in different chats and every time I can open the image modal multiple times.

Can you reproduce the bug without your fix (and on the latest `master` branch) and explain to me how I can also reproduce it?

izinin: Hi , @jcbrand , i cannot reproduce the problem in not customised Conversejs either, [my test](https://www.loom.com/share/d38beb2e01ac4772b18148c02b1f65e4?sid=25c9b3b6-8c14-43d2-8142-56adcb06a9ea). 

Please kindly reject this PR. 

To provide background info on this error: we are working on Converse customisation and decided to contribute back fixes that we made in our implementation. In our implementation the problem did exist and fixed the way as it is submitted with this PR. Or customisation changes is more mobile device oriented UI layout.  Since it is hard to find the problem root cause we decided just to re-init modal window every time it is invoked. I do not see drawbacks in this approach i might be unaware of some conversejs working details. Is the modal window reusing justified in this context ? IMO it adds complexity to the code that impacts development experience and cost. I.e. no benefits in this particular case.

PR #3438, opened 2024-07-11
Documentation: how to override custom elements.

JohnXLivingston:
Sorry, i was not able to test that the RST format is valid. As said on the Converse XMPP room, `make html` does not work, because of a pip dependency issue:

I go an error because sphinxcontrib.applehelp needs Sphinx 5. But the documentation won't compile with Sphinx5.
Similar issue here:
https://github.com/sphinx-doc/sphinx/issues/11890

One way to fix: find someone who has a working env, and copy all exact packages versions in the requirements.txt.


jcbrand:
LGTM, thanks!

PR #3443, opened 2024-07-11
Fix #3439: form field regression.

### LINKED ISSUE #3439: Form regression for number fields
**Describe the bug**

Using Converse upstream (v11 unreleased), i have some integer fields in room config that are now displayed using datetime fields:

image

This should be an input type=number (and it was with Converse v10).
Here is a screenshot of the same field in Gajim:

image

Here is the field definition:

```xml
<field label="Slow Mode (0=disabled, any positive integer= users can send a message every X seconds.)" var="muc#roomconfig_slow_mode_duration" type="text-single">
        <validate xmlns="http://jabber.org/protocol/xdata-validate" datatype="xs:integer">
          <range min="0" />
        </validate>
        <value>5</value>
</field>
```

The code is in `src/utils/html.js`:

```javascript
} else {
        const date = xfield.value ? dayjs(xfield.value) : null;
        if (date?.isValid()) {
            return tplDateInput({
                ...default_vals,
                ...xfield,
                value: date.format('YYYY-MM-DDTHH:mm:ss'),
            });
        }
```

I don't understand why this is done so. It seems to consider that if the value is a date, the field should be a date.
This has 2 drawbacks:
* if a text field current value is a date, it does not mean that the field should be a date
* it seams "5" is a valid date for dayjs

JohnXLivingston:
Trying to fix https://github.com/conversejs/converse.js/issues/3439

I am absolutly not sure that this fix is correct. I only tried to guess what you meant to write.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -90,8 +90,8 @@
         "max-depth": "error",
         "max-len": ["error", {
             "code": 120,
-            "comments": 120,
             "ignoreStrings": true,
+            "ignoreComments": true,
             "ignoreUrls": true
         }],
         "max-lines": "off",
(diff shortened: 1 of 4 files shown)


jcbrand:
I think what you did makes sense, thanks.

JohnXLivingston:
For the record, the type definition for `XFormFieldTypes` is still missing some possible values. For example `number`.

PR #3449, opened 2024-07-15
Fix: MUC config form fails when no image avatar field.

JohnXLivingston:
When the MUC component offers no avatar field, the MUC room config save fails.
This PR fixes it.


JohnXLivingston:
Hum, the issue was not what i meant when i did the first commit.
I haven't the image picker, because in my project i  disable the "profile" plugin, so the converse-image-picker custom element was not here.
The second commit fixes this.

But there is still a bug: if the server does not support MUC avatars, the picker is still displayed. And won't do anything.

JohnXLivingston:
Ok, should be ok now.

PR #3452, opened 2024-07-16
Fix regression for MUC message avatars.

JohnXLivingston:
There is a regression in some rare case (comes with the new colorized avatars).

If for some reason we haven't loaded all occupants vcards, the code will load avatars, but not display them next to messages.
This PR fixes it: if the vcard is not on the contact model, we fallback to the message (as before, but code wasn't checking if the vcard is on the contact).

Note: this bug is visible in my project, were i have an optimization: i do not load all occupants vcards on join, if the occupant list is hidden (for performance reason). So, without this fix, message won't show the avatar, because `contact.vcard` is not set (we haven't called setVCardOnOccupant), but `message.vcard` is (thanks to setVCardOnMUCMessage).

### DIFF
--- a/src/shared/chat/templates/message.js
+++ b/src/shared/chat/templates/message.js
@@ -20,6 +20,10 @@ export default (el, o) => {
     const contact = el.model.occupant || el.model.contact;
     const author_style = getAuthorStyle(contact);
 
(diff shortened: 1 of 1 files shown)


jcbrand:
Thanks!

PR #3463, opened 2024-07-30
Fix typo: occupant filter is buggy in MUC with less than 5 occupants.

JohnXLivingston:
There was a typo. When less than 5 MUC occupants, the filter action should be removed, but it was not the case.
This was resulting in an empty filter element (see screenshot).

image

Note: perhaps we should always show the filter for MUCs? I'm aware that it is not very usefull when only few participants, but it would make the code simpler.
Currently, the 5 occupant limit is hard coded in several places. Could cause some troubles in the future.


jcbrand:
> Could cause some troubles in the future.

Ideally we catch these via tests, to avoid regressions.

PR #3472, opened 2024-08-05
Fix singleton mode with bootstrap5:

JohnXLivingston:
Seems that negative margin is no more needed with the recent changes.

### DIFF
--- a/src/plugins/singleton/singleton.scss
+++ b/src/plugins/singleton/singleton.scss
@@ -18,13 +18,6 @@
             .chatbox {
                 margin: 0;
                 position: relative;
-                margin-left: -15px;
-                @media screen and (max-width: $mobile-portrait-length) {
-                    margin-left: 0;
-                }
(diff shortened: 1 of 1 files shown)


jcbrand:
LGTM (without testing it). 

BTW, I'm planning to get rid of the `converse-mobile` view mode. Instead, any other view mode should be responsive and therefore work on mobile viewports as well.

That's at least the idea right now.

JohnXLivingston:
> BTW, I'm planning to get rid of the `converse-mobile` view mode. Instead, any other view mode should be responsive and therefore work on mobile viewports as well.
> 
> That's at least the idea right now.

Seems a good idea.

JohnXLivingston:
I just made another commit. I copied the `left: 1em` from the overlayed mode, but this was a mistake. Must be `0`.

jcbrand:
Merged it manually, thanks.

PR #3498, opened 2024-09-03
MUC occupants badges: displays short labels, with full label as title.

JohnXLivingston:
By default, we take the first letter of all badge, and only display this letter (full badge name is set as title and aria-label for screen readers). If there is ambiguous letters, we use 2 (or 3, ... up to 4) letters.

image

### DIFF
--- a/src/plugins/muc-views/templates/occupant.js
+++ b/src/plugins/muc-views/templates/occupant.js
@@ -13,6 +13,72 @@ const i18n_occupant_hint = /** @param {MUCOccupant} o */(o) => {
     return __('Click to mention %1$s in your message.', o.get('nick'));
 }
 
+let badges_definitions; // will be initialized at first call (to be sure that the __ function is correctly loaded).
+
+/**
+ * Inits badges definitions.
+ * For short labels, it will use the label first letter. If there is ambigous short labels, it will try to add up to 4 letters.
+ * Letters will be uppercase.
+ */
+function initBadgesDefinitions () {
+    badges_definitions = {}
+    badges_definitions['owner'] = {
+        label: __('Owner'),
+        classname: 'badge-groupchat'
+    };
+    badges_definitions['admin'] = {
+        label: __('Admin'),
+        classname: 'badge-info'
+    };
+    badges_definitions['member'] = {
+        label: __('Member'),
+        classname: 'badge-info'
+    };
+    badges_definitions['moderator'] = {
+        label: __('Moderator'),
+        classname: 'badge-info'
+    };
+    badges_definitions['visitor'] = {
+        label: __('Visitor'),
+        classname: 'badge-secondary'
+    };
+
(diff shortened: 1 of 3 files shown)


JohnXLivingston:
Tests failed on the CI, but seems not related to my PR. Tests are ok on my computer.

src/plugins/muc-views/templates/occupant.js
```diff
@@ -13,6 +13,72 @@ const i18n_occupant_hint = /** @param {MUCOccupant} o */(o) => {
...
+    };
+    badges_definitions['visitor'] = {
+        label: __('Visitor'),
+        classname: 'badge-secondary'
+    };
+
+    // And now we must compute unique short labels.
+    let seen;
+    for (
+        let currentLength = 1;
```
jcbrand, line 49: I know this is different than how 99% of JavaScript projects write variables, and it's a habit I brought over from the Python world, but in Converse we write variables with underscores.

So this should be `current_length`.

I've thought about changing this rule, but it would be a lot of work to rename all variables!
At the same time, having a mix of different styles for writing variables is also not great.

jcbrand:
> Tests failed on the CI, but seems not related to my PR. Tests are ok on my computer.

If you rebase on to the master branch then this should be fixed.

src/plugins/muc-views/templates/occupant.js
JohnXLivingston: Oops. I did it for other variables and functions, i just missed this one. I just forced push the fix (and rebased on master to fix the CI).

src/plugins/muc-views/templates/occupant.js
```diff
@@ -13,6 +13,72 @@ const i18n_occupant_hint = /** @param {MUCOccupant} o */(o) => {
...
 }
 
+let badges_definitions; // will be initialized at first call (to be sure that the __ function is correctly loaded).
+
+/**
+ * Inits badges definitions.
+ * For short labels, it will use the label first letter. If there is ambigous short labels, it will try to add up to 4 letters.
+ * Letters will be uppercase.
+ */
+function init_badges_definitions () {
```
jcbrand, line 23: Function should be camelCase.

src/plugins/muc-views/templates/occupant.js
```diff
@@ -13,6 +13,72 @@ const i18n_occupant_hint = /** @param {MUCOccupant} o */(o) => {
...
+    };
+
+    // And now we must compute unique short labels.
+    let seen;
+    for (
+        let current_length = 1;
+        current_length < 5 && (!seen || Object.values(seen).find(count => count > 1));
+        current_length++
+    ) {
+        const currentlySeen = {}
```
jcbrand, line 53: Should be with underscores.

JohnXLivingston: Fixed

Issue #3529, opened 2024-10-19
Clicking on unbookmarked rooms with automatic "letters" avatar duplicates it into room title


BetaRays:
**Describe the bug**
When joining a room without an avatar, Converse.js will automatically make one based on the room’s title, for example, for a MUC named "broken@muc.localhost", it will use the letter “B”.
When clicking on this room, if it is not bookmarked, its display name will change to "B broken@muc.localhost", which makes the avatar change to "BB": clicking any more times will prepend "BB" to the room name and the avatar won’t change.
This may also happen when an unread message count is shown (e.g. if you are mentioned in a room you aren’t currently looking at), so the bug exists outside of the new room icon feature.

**To Reproduce**
1. Join a MUC without an avatar, ensure it is *not* bookmarked.
2. Click on the MUC’s entry in the sidebar (multiple times)
3. Notice the name getting longer each time.

**Expected behavior**
MUC names shouldn’t automatically change, becoming unreadable when clicking on them.

**Screenshots**

![example screenshot](https://github.com/user-attachments/assets/4dc363b5-112e-4e46-b6e1-67a8b5d1ba21)

**Environment (please complete the following information):**
 - Desktop or Mobile? Desktop
 - Browser [e.g. Chrome, Firefox] Firefox
 - Converse.js version [e.g. 4.2.0] v11.0.0, current master: 37dfc252a93535785784c08ff0f9ffa3b683b03a

**Additional context**
I think the issue comes from [this line](https://github.com/conversejs/converse.js/blob/master/src/plugins/roomslist/view.js#L58), specifically, `target.textContent` seems to include the automatic avatar.

Neustradamus:
@BetaRays: For example in Psi/Psi+, when an user has not an Avatar in vCard, there is a default Avatar picture:
- https://github.com/psi-im/psi/blob/master/iconsets/system/default/default_avatar.png

PR #3530, opened 2024-10-20
Prevent duplication of text avatar into display name

BetaRays:
Fixes https://github.com/conversejs/converse.js/issues/3529.

- ~~When adding a configuration variable, please make sure to
      document it in `docs/source/configuration.rst`~~

### DIFF
--- a/src/plugins/roomslist/view.js
+++ b/src/plugins/roomslist/view.js
@@ -55,7 +55,7 @@ export class RoomsList extends CustomElement {
     async openRoom (ev) {
         ev.preventDefault();
         const target = u.ancestor(/** @type {HTMLElement} */(ev.target), '.open-room');
-        const name = target.querySelector(".room-name")?.textContent;
+        const name = target.getAttribute('data-room-name');
         const jid = target.getAttribute('data-room-jid');
         const data = {
             'name': name || Strophe.unescapeNode(Strophe.getNodeFromJid(jid)) || jid
--- a/src/plugins/roomslist/view.js
+++ b/src/plugins/roomslist/view.js
@@ -55,7 +55,7 @@ export class RoomsList extends CustomElement {
     async openRoom (ev) {
         ev.preventDefault();
         const target = u.ancestor(/** @type {HTMLElement} */(ev.target), '.open-room');
-        const name = target.textContent;
+        const name = target.querySelector(".room-name")?.textContent;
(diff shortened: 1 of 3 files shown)


BetaRays:
I tried making a test for it, but I’m not sure how to wait until the name has been updated, if it’s supposed to not change.

BetaRays:
Also, I’m not sure if adding an entry to the changelog is necessary here if the bug didn’t appear in the previous version (or any released version).

src/plugins/roomslist/view.js
```diff
@@ -55,7 +55,7 @@ export class RoomsList extends CustomElement {
     async openRoom (ev) {
         ev.preventDefault();
         const target = u.ancestor(/** @type {HTMLElement} */(ev.target), '.open-room');
-        const name = target.textContent;
+        const name = target.querySelector(".room-name")?.textContent;
```
jcbrand, line 58: Thank you for the fix.

I think a slightly better solution is to put the name as a `data-` attribute on the `a` element, similar to `data-room-jid`.

Then this becomes: `const name = target.getAttribute('data-room-name');`

jcbrand:
> I tried making a test for it, but I’m not sure how to wait until the name has been updated, if it’s supposed to not change.

Maybe just look for an existing test in the roomslist plugin and add a check that the name stays the same when clicking on the item in the list to open a room.

BetaRays:
> > I tried making a test for it, but I’m not sure how to wait until the name has been updated, if it’s supposed to not change.
> 
> Maybe just look for an existing test in the roomslist plugin and add a check that the name stays the same when clicking on the item in the list to open a room.

My issue is that I need to put some kind of delay between clicking and checking for the new name (otherwise the test passes when it shouldn’t because the name hasn’t been updated yet): this is usually done with `u.waitUntil` but I’m not sure what condition to use (waiting for the name to change would fail if it stays the same as it’s supposed to).

jcbrand:
> My issue is that I need to put some kind of delay between clicking and checking for the new name (otherwise the test passes when it shouldn’t because the name hasn’t been updated yet): this is usually done with `u.waitUntil` but I’m not sure what condition to use (waiting for the name to change would fail if it stays the same as it’s supposed to).

You could set a different name before clicking and then checking that it is set back to the right name, but I don't think a test is that important in this case, so I'll merge.

PR #3548, opened 2024-11-21
Reject non-secure WebSocket or BOSH endpoints

linkmauve:
When discovering endpoints using XEP-0156, the server admin can list any kind of URL, but we want to use only secure ones using TLS.  In order to achieve that, we filter out the lists before using the first one available.

This was causing connection to fail with the step.im server, which exposes in order ws:, wss: and http:, and we were previously using only the first and third ones, instead of the second like we should.

Should fix the issue reported by @VnPower at https://misskey.pm/notes/a0v0aaw0tbknyojk


jcbrand:
Thanks!

PR #3608, opened 2025-03-11
Replace incorrectly twice-duplicated Romanian translated string with two correctly translated Romanian strings.

tonicwave:

- [ x] Add a changelog entry for your change in `CHANGES.md`
- [ x] When adding a configuration variable, please make sure to
- [ x] Please add a test for your change. Tests can be run in the commandline

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -60,6 +60,7 @@
 - Fix: removing the "add to contact" button in occupant modal in singleton mode (as there is no roster).
 - Fix: trying to use emojis with an uppercase letter breaks the message field.
 - Fix: renaming getEmojisByAtrribute to getEmojisByAttribute.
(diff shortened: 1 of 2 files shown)


jcbrand:
The two English sentences are not the same, so the Romanian translations should also not be the same.

Looks like you got confused by the fact that the two original Romanian translations are the same, but the `fuzzy` keyword disables the second translation and indicates that it needs to be looked at again.

tonicwave:
> The two English sentences are not the same, so the Romanian translations should also not be the same.
> 
> Looks like you got confused by the fact that the two original Romanian translations are the same, but the `fuzzy` keyword disables the second translation and indicates that it needs to be looked at again.

I have committed an update which corrects the 2nd string correctly.  (Sadly, what I actually did was fat-finger my copy paste of the second string. )

PR #3680, opened 2025-05-30
Fix auto zoom in in ios safari fullpage view mode when input message

narutogo:
Fix auto zoom in  page in ios safari fullpage view mode when input message, it will result in a scroll bar present in x and y.


narutogo:
The reason is that ios safari auto zoom in fonts when input fonts in smaller than 16px

jcbrand:
Thanks :+1:

PR #3691, opened 2025-06-08
ui: display locked_domain behind login field [tested]

Arup3201:

- [✅] Add a changelog entry for your change in `CHANGES.md`
- [✅] Please add a test for your change. Tests can be run in the commandline

### DIFF
--- a/src/plugins/controlbox/templates/loginform.js
+++ b/src/plugins/controlbox/templates/loginform.js
@@ -72,6 +72,10 @@ function tplPasswordInput() {
     `;
 }
 
+function tplDomain(domain) {
+    return html` <span class="input-group-text addon" id="basic-addon2">@${domain}</span> `;
+}
+
 function tplAuthFields() {
     const authentication = api.settings.get('authentication');
     const i18n_login = __('Log in');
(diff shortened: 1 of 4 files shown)


src/plugins/controlbox/templates/loginform.js
```diff
@@ -84,17 +88,20 @@ function tplAuthFields() {
...
-                ?autofocus=${api.settings.get('auto_focus') ? true : false}
-                value="${api.settings.get('jid') ?? ''}"
-                required
-                class="form-control"
-                type="text"
-                name="jid"
-                placeholder="${placeholder_username}"
-            />
-        </div>
+            <div class="input-group">
```
jcbrand: The `input-group` class should only be added if `locked_domain` is truthy.

src/plugins/controlbox/tests/login.js
```diff
@@ -3,6 +3,25 @@
 const u = converse.env.utils;
 
 describe("The Login Form", function () {
+    
+    it("contains an addon when locked_domain is set",
```
jcbrand:
```suggestion
    it("shows the domain when locked_domain is set", 
```

The word `addon` is very ambiguous.

src/plugins/controlbox/tests/login.js
```diff
@@ -3,6 +3,25 @@
...
+        
+        mock.toggleControlBox();
+        const cbview = await u.waitUntil(() => _converse.chatboxviews.get('controlbox'));
+        await u.waitUntil(() => cbview.querySelectorAll('div.input-group').length)
+        
+        const addons = cbview.querySelectorAll('div.input-group');
+        expect(addons.length).toBe(1);
+        
+        const addon = addons[0];
+        expect(addon.innerHTML, 'jabber.hot-chilli.eu');
```
jcbrand: It's better to use `.textContent` instead of `.innerHTML`.

That said, even better would be to do `cbview.querySelector('.input-group .addon').textContent`

Arup3201: I have updated the code, but wanted to know why `cbview.querySelector('.input-group .addon').textContent` is a better way?

jcbrand: It's more specific and explicit because you're getting the actual `.addon` element and not the containing `.input-group` element (which is more ambiguous and could also match other elements in the future, thereby breaking the test).

jcbrand: With the change you made, you're still just querying for the `.input-group` and not the inner `.addon`. I'll merge it however, since I want to make a new release.

jcbrand:
Thanks @Arup3201 and congratulations on your first merged PR in this repo :)

PR #2986, opened 2022-07-21
commit to be deleted

PawBud:

PR #3765, opened 2025-07-25
index.js: change link to list of pulic XMPP providers for registrations

yurtpage:
The registration process is the main pain for regular users (just imagine explaining this to your relatives). 
The old compliance.conversations.im is not intended for users to pick a server. But this can be added to it with low efforts but this may be difficult to negotiate.
The Kaidan client uses the https://providers.xmpp.net/ as a categorized list of providers.
According to changelog the Converse used before some list of providers from the site #1206: 

> v4.0.1 (2018-09-19) Use https://compliance.conversations.im instead of dead link tot st https://xmpp.net

Unfortunately from the recommended A category only three servers doesn't require for a CAPTCHA which is annoying: https://jabber.fr, https://yax.im, https://xmpp.eath. And the BOSH is supported only by the jabber.fr. Basically we can just add it to the text. 

We should ask other servers to enable the BOSH (and maybe disable the CAPTCHA). Anyway, it would be really good to put the `conversejs.org` as a default value (not just a placeholder) but also it's possible to add other domains into auto-completion to make it easier for user to type them (#2986). Can we add the `conversejs.org` to the list of public providers or you don't want to have many registrations from side clients?

BTW the conversations.im server has BOSH and is working but in the rating it has a low category because it has something disabled. I remember that asked its owner to fix that but he is too busy but also he wasn't happy to be used by anyone as a general server.

The compliance test checks if a server has XEP-0156: Discovering Alternative XMPP Connection Methods (HTTP)
https://compliance.conversations.im/test/xep0156/
So some of this servers with green check potentially may allow BOSH and allow registrations. If found a few from the list:
agnos.is
ari.lt 
asozial.org

Ideally we need to have the list of such servers.

There are some other places that use the Compliance page that I didn't changed:
https://github.com/search?q=repo%3Aconversejs%2Fconverse.js%20compliance.conversations.im&type=code

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
 - Add approval banner in chats with requesting contacts or unsaved contacts
 - Some fixes regarding manually resized chats in `overlayed` view mode.
 - Replace webpack with [rspack](https://rspack.rs)
+- Registration: Use https://providers.xmpp.net instead of https://compliance.conversations.im
 
 ## 11.0.1 (2025-06-09)
 
--- a/index.html
+++ b/index.html
@@ -228,8 +228,8 @@
                 <div class="h-100 p-3 rounded-3 text-center">
                     <h2>Get Started in Minutes 🚀</h2>
                     <p>Use our <a href="/fullscreen.html">web app</a> to connect to any XMPP server. You can log in with your existing XMPP account.</p>
-                    <p>No account? No problem! With Converse you can register an account on any public XMPP server that allows in-band registration.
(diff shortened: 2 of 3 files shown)


jcbrand:
Thanks for your pull request.

> Anyway, it would be really good to put the conversejs.org as a default value 

I agree it would be better from a user perspective, but I run that server and don't want the headaches and hassles of running a large public server. At least not at the moment.

On the topic of BOSH, it's not really that important anymore because we can now have websocket connections which are much better, so I don't consider BOSH support to be very important.

yurtpage:
New usual users often confused that they need to register :) So we need to simplify this step for them as much as possible.

Issue #3772, opened 2025-08-01
Quick Start warning about version number should apply to XMPP server plugins


julien-picalausa:
Greetings,

While setting up converse.js, I discovered an apparent inconsistency in the quick start guide that might not be immediately obvious.

https://github.com/conversejs/converse.js/blob/d7a41b5ab6a7175faa0d5d338ec44be18b1f8cbe/docs/source/quickstart.rst?plain=1#L20-L47

The quick start guide initially recommends using XMPP server plugins. Then, the following section, ends in a warning to "Always specify a version number in production." However, looking at the source code of ejabberd's mod_conversejs, it appears to directly link to `https://cdn.conversejs.org/dist/converse.min.js` and `https://cdn.conversejs.org/dist/converse.min.css` (no version number) in its default configuration (which I imagine is what most people would use.) Even if one could argue that this is an issue with ejabberd, it would regardless be valuable, for consistency, to put a similar warning in the `Host it via your XMPP Server` section, providing an example configuration for ejabberd that enforces a specific version number, at least until something changes on the ejabberd side.

I did not check if the plugins for OpenFire or Prosody require a similar mention.

jcbrand:
Hi @julien-picalausa, thanks for writing the issue. I invite you to make a pull request with your proposed change. Thanks.

julien-picalausa:
Ok. I'll take a stab at it. I'll admit that I never touched .rst files, but there is a first time for everything.

PR #3777, opened 2025-08-02
Adda a warning to quickstart about using version numbers when configuring xmpp server plugins.

julien-picalausa:
This adds a warning to the `Host it via your XMPP Server` part of the QuickStart guide, consistent with the warning in the `Using the CDN` section. A configuration example specific to ejabberd is also added to illustrate how to satisfy that warning.

fixes: #3772

### DIFF
--- a/docs/source/quickstart.rst
+++ b/docs/source/quickstart.rst
@@ -22,7 +22,7 @@ For example, the following XMPP servers have plugins available:
 * **ejabberd**: `mod_conversejs <https://docs.ejabberd.im/admin/configuration/modules/#mod-conversejs>`_
 
 .. warning::
-    When configuring one of these plugins in production, ensure that they use a specific version of the conversjs resources to avoid breaking changes.
+    When configuring one of these plugins in production, it's good practice to use a specific version of the Converse resources to avoid breaking changes.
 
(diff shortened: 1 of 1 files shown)


docs/source/quickstart.rst
```diff
@@ -21,6 +21,19 @@ For example, the following XMPP servers have plugins available:
 * **Prosody**: `mod_conversejs <https://modules.prosody.im/mod_conversejs.html>`_
 * **ejabberd**: `mod_conversejs <https://docs.ejabberd.im/admin/configuration/modules/#mod-conversejs>`_
 
+.. warning::
+    When configuring one of these plugins in production, ensure that they use a specific version of the conversjs resources to avoid breaking changes.
```
jcbrand: Thanks. I would soften this sentence a bit, to something like:

>  When configuring one of these plugins in production, it's good practice to use a specific version of the Converse resources to avoid breaking changes.

The warning further down below has plugin developers in mind, where a new version of Converse can break their plugins. I think someone who hosts via the XMPP server will most likely not have their own plugins included.

julien-picalausa: Sure, that makes sense. I'll fix it :)

jcbrand:
Thanks!

PR #3779, opened 2025-08-03
Hide reset password tab when authentication is external (Fixes #3719)

### LINKED ISSUE #3719: "Reset Password" option visible although authentication is set to "external"
I managed to setup ConverseJS with Azure Entra ID authentication with help of Apache OIDC module. I have configured "authentication: 'external'" while initializing ConverseJS. I want to hide "Reset Password" option in ConverseJS as authentication is done by Azure Entra ID and no need of this option. How to hide this option? or showing "Reset Password" is the expected behaviour?

Aswanth-v:
Title:
Hide Password Reset tab when external auth is used

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,8 @@
 - Some fixes regarding manually resized chats in `overlayed` view mode.
 - Replace webpack with [rspack](https://rspack.rs)
 - Registration: Use https://providers.xmpp.net instead of https://compliance.conversations.im
+- Hide password reset tab when external authentication is enabled. (Fixes #3719)
+
 
 ## 11.0.1 (2025-06-09)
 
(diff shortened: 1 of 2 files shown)


jcbrand:
Thanks, but this doesn't look right. Firstly, your change doesn't hide the tab, but instead hides the contents of the tab.

Also, there is already a check `shouldShowPasswordResetForm()` which is used to show/hide the tab:
https://github.com/conversejs/converse.js/blob/2b5456cce487bb29e12dfb37ffb3e31b77ed395e/src/plugins/profile/templates/profile_modal.js#L79

Can you put a breakpoint inside that function and check why it returns `true` for you?

What is the `mechinism.mechname` value?

https://github.com/conversejs/converse.js/blob/2b5456cce487bb29e12dfb37ffb3e31b77ed395e/src/plugins/profile/utils.js#L28

Aswanth-v:
@jcbrand 
I’ll fix it. Sorry about the mistake — could you please assign it to me?

PR #3780, opened 2025-08-04
Hide reset password tab when authentication is external

Aswanth-v:
Updated the shouldShowPasswordResetForm function to correctly hide the reset password tab for external authentication methods.

Hope this fixes the previous implementation issue.

### DIFF
--- a/src/plugins/profile/utils.js
+++ b/src/plugins/profile/utils.js
@@ -26,16 +26,18 @@ export function getPrettyStatus(stat) {
  */
 export function shouldShowPasswordResetForm() {
     const conn = _converse.api.connection.get();
-    const mechanism = conn._sasl_mechanism;
- console.log('🔍 SASL Mechanism:', mechanism?.mechname, mechanism);
-    if (
-        mechanism.mechname === 'EXTERNAL' ||
-        mechanism.mechname === 'ANONYMOUS' ||
-        mechanism.mechname === 'X-OAUTH2' ||
-        mechanism.mechname === 'OAUTHBEARER'
-    ) {
+    const mechanism = conn?._sasl_mechanism;
+
+    console.log('🔍 SASL Mechanism:', mechanism?.mechname, mechanism);
+
+    if (!mechanism?.mechname) {
+        // Fallback: no mechanism detected
         return false;
     }
-    return true;
+
+    const externalMechs = ['EXTERNAL', 'ANONYMOUS', 'X-OAUTH2', 'OAUTHBEARER'];
+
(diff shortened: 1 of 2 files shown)


src/plugins/profile/utils.js
```diff
@@ -26,15 +26,18 @@ export function getPrettyStatus(stat) {
...
+
+    if (!mechanism?.mechname) {
+        // Fallback: no mechanism detected
         return false;
     }
-    return true;
+
+    const externalMechs = ['EXTERNAL', 'ANONYMOUS', 'X-OAUTH2', 'OAUTHBEARER'];
+
+    return !externalMechs.includes(mechanism.mechname.toUpperCase());
```
jcbrand, line 40: Looking at the code, it's still not clear to me what the actual bug is that is causing the password reset form to show for you as reported in #3719.

Have you checked what the `mechname` value is in your specific case?

I would expect it to be `EXTERNAL`, in which case this function should return `false` and the form shouldn't be shown.

jcbrand: Also, it would be better to just work in the existing PR instead of creating a new one.

I'll close the other one now in favor of this one.

Aswanth-v: Hi @jcbrand,
Thanks again for the feedback.

I checked again — in my setup, the reset password tab only appears when the auth mechanism is not one of 'EXTERNAL', 'ANONYMOUS', 'X-OAUTH2', or 'OAUTHBEARER'. Based on that, I refined the logic to return false for these cases.

I understand the original function should have handled it, so I’ll double-check the exact value of mechanism.mechname in my environment and get back with a clearer fix if needed.

Thanks again for the opportunity — even though this PR wasn’t merged, I really appreciate being able to contribute and learn from it.

jcbrand: Sorry @Aswanth-v, I see now that you are not the author of #3719. Perhaps @NarayanamoorthySrinivasan could check what the value of `mechanism.mechname` in their setup and let us know.

Otherwise, it might be enough to add an additional check for `api.settings.get('authentication') === 'external'`, but I'd prefer to avoid it and to rely on the SASL mechanism if possible.

jcbrand: https://github.com/conversejs/converse.js/issues/3719#issuecomment-3152192398

jcbrand:
Thanks for your work on this @Aswanth-v and sorry that your commit didn't get merged.

I became impatient and made a quick commit here:
https://github.com/conversejs/converse.js/commit/301995a335b03cde14300de07a619d6a2fb531e5

PR #3782, opened 2025-08-05
Load avatar data over HTTP if available in vCard EXTVAL

edshot99:
Its possible to point the user avatar to a URL address and store it in PEP. It then becomes accessible from vcard-temp in the EXTVAL tag.

The same can be done with MUC avatars by setting the EXTVAL tag with the url, but this is not oficially supported by XEP-0486.

Edit: I think handleVCardUpdatePresence() in src/headless/plugins/vcard/utils.js should be modified, but i'm not sure if its only needed for the binary avatars or not since i'm not too familiar with Javascript or Sizzle. Seems to function fine without any changes to it however.

### DIFF
--- a/src/headless/plugins/vcard/parsers.js
+++ b/src/headless/plugins/vcard/parsers.js
@@ -12,6 +12,7 @@ export async function parseVCardResultStanza(iq) {
         fullname: iq.querySelector(':scope > vCard FN')?.textContent,
         image: iq.querySelector(':scope > vCard PHOTO BINVAL')?.textContent,
         image_type: iq.querySelector(':scope > vCard PHOTO TYPE')?.textContent,
+        image_url: iq.querySelector(':scope > vCard PHOTO EXTVAL')?.textContent,
         nickname: iq.querySelector(':scope > vCard NICKNAME')?.textContent,
         role: iq.querySelector(':scope > vCard ROLE')?.textContent,
         stanza: iq, // TODO: remove?
@@ -21,7 +22,10 @@ export async function parseVCardResultStanza(iq) {
         vcard_error: undefined,
         image_hash: undefined,
     };
-    if (result.image) {
+    if (result.image_url) {
+        result['image_url'] = result.image_url;
+    }
+    else if (result.image) {
         const buffer = u.base64ToArrayBuffer(result.image);
         const ab = await crypto.subtle.digest('SHA-1', buffer);
         result['image_hash'] = u.arrayBufferToHex(ab);
--- a/src/headless/plugins/vcard/parsers.js
+++ b/src/headless/plugins/vcard/parsers.js
@@ -10,9 +10,6 @@ export async function parseVCardResultStanza(iq) {
     const result = {
         email: iq.querySelector(':scope > vCard EMAIL USERID')?.textContent,
         fullname: iq.querySelector(':scope > vCard FN')?.textContent,
-        image: iq.querySelector(':scope > vCard PHOTO BINVAL')?.textContent,
-        image_type: iq.querySelector(':scope > vCard PHOTO TYPE')?.textContent,
-        image_url: iq.querySelector(':scope > vCard PHOTO EXTVAL')?.textContent,
         nickname: iq.querySelector(':scope > vCard NICKNAME')?.textContent,
         role: iq.querySelector(':scope > vCard ROLE')?.textContent,
         stanza: iq, // TODO: remove?
@@ -20,15 +17,26 @@ export async function parseVCardResultStanza(iq) {
         vcard_updated: new Date().toISOString(),
         error: undefined,
         vcard_error: undefined,
-        image_hash: undefined,
     };
-    if (result.image_url) {
-        result['image_url'] = result.image_url;
-    }
-    else if (result.image) {
-        const buffer = u.base64ToArrayBuffer(result.image);
+
+    const image = iq.querySelector(':scope > vCard PHOTO BINVAL')?.textContent;
+    const image_type = iq.querySelector(':scope > vCard PHOTO TYPE')?.textContent;
+    const image_url = iq.querySelector(':scope > vCard PHOTO EXTVAL')?.textContent;
+
+    if (image) {
+        const buffer = u.base64ToArrayBuffer(image);
         const ab = await crypto.subtle.digest('SHA-1', buffer);
-        result['image_hash'] = u.arrayBufferToHex(ab);
+
+        Object.assign(result, {
+            image,
+            image_type,
+            image_hash: u.arrayBufferToHex(ab),
+        });
+    }
+    else if (image_url) {
+        Object.assign(result, {
+            image_url,
+        });
     }
     return result;
 }
--- a/src/headless/plugins/vcard/parsers.js
+++ b/src/headless/plugins/vcard/parsers.js
@@ -32,8 +32,7 @@ export async function parseVCardResultStanza(iq) {
             image_type,
             image_hash: u.arrayBufferToHex(ab),
         });
-    }
-    else if (image_url) {
+    } else if (image_url) {
         Object.assign(result, {
             image_url,
         });
--- a/src/headless/plugins/vcard/utils.js
+++ b/src/headless/plugins/vcard/utils.js
@@ -46,7 +46,11 @@ export function onOccupantAvatarChanged(occupant) {
         vcards.push(_converse.state.vcards.get(occupant.get('jid')));
     }
     vcards.push(_converse.state.vcards.get(occupant.get('from')));
-    vcards.forEach((v) => (hash || url) && v && (v?.get('image_hash') !== hash || v?.get('image_url') !== url) && api.vcard.update(v, true));
+    vcards.filter((v) => v).forEach((v) => {
+        if (hash && v.get('image_hash') !== hash || url && v.get('image_url') !== url) {
+            api.vcard.update(v, true);
+        }
+    });
 }
 
 /**
--- a/src/headless/plugins/vcard/utils.js
+++ b/src/headless/plugins/vcard/utils.js
@@ -39,13 +39,14 @@ export function createStanza(type, jid, vcard_el) {
  * @param {MUCOccupant} occupant
  */
 export function onOccupantAvatarChanged(occupant) {
+    const url = occupant.get('image_url');
     const hash = occupant.get('image_hash');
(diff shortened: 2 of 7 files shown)


src/headless/plugins/vcard/parsers.js
```diff
@@ -12,6 +12,7 @@ export async function parseVCardResultStanza(iq) {
         fullname: iq.querySelector(':scope > vCard FN')?.textContent,
         image: iq.querySelector(':scope > vCard PHOTO BINVAL')?.textContent,
         image_type: iq.querySelector(':scope > vCard PHOTO TYPE')?.textContent,
+        image_url: iq.querySelector(':scope > vCard PHOTO EXTVAL')?.textContent,
```
jcbrand: XEP-0054 and XEP-0292 both support `PHOTO EXTVAL` but funnily enough XEP-0153 VCard-based Avatars discourages it. I'm not sure why.

```
The <PHOTO/> element SHOULD NOT contain an <EXTVAL/> that points to a URI for the image file.
```

In any case, the changes look OK to me, but I think we should put `result.image` first in the `if` statement below, and then check for `result.image_url`.

src/headless/plugins/vcard/parsers.js
```diff
@@ -21,7 +22,10 @@ export async function parseVCardResultStanza(iq) {
         vcard_error: undefined,
         image_hash: undefined,
     };
-    if (result.image) {
+    if (result.image_url) {
+        result['image_url'] = result.image_url;
```
jcbrand: This line is redundant and doesn't do anything.

I think it would be better to get the `image`, `image_type` and `image_url` values outside of the  declaration of the `result` object, and then to update the result object based on whether `image` or `image_url` is available.

For example something like:

```javascript
const result = { 
    // redacted the code that goes here
}

const image = iq.querySelector(':scope > vCard PHOTO BINVAL')?.textContent;
const image_type = iq.querySelector(':scope > vCard PHOTO TYPE')?.textContent;
const image_url = iq.querySelector(':scope > vCard PHOTO EXTVAL')?.textContent;

if (image) {
        const buffer = u.base64ToArrayBuffer(result.image);
        const ab = await crypto.subtle.digest('SHA-1', buffer);
        Object.assign(result, {
                image,
                image_type,
                image_hash: u.arrayBufferToHex(ab),
       });
} else if (image_url) {
     result.image_url = image_url;
}

File: src/headless/plugins/vcard/utils.js [hunk 3]
```diff
@@ -39,13 +39,14 @@ export function createStanza(type, jid, vcard_el) {
...
 export function onOccupantAvatarChanged(occupant) {
+    const url = occupant.get('image_url');
     const hash = occupant.get('image_hash');
     const vcards = [];
     if (occupant.get('jid')) {
         vcards.push(_converse.state.vcards.get(occupant.get('jid')));
     }
     vcards.push(_converse.state.vcards.get(occupant.get('from')));
-    vcards.forEach((v) => hash && v && v?.get('image_hash') !== hash && api.vcard.update(v, true));
+    vcards.forEach((v) => hash && v && (v?.get('image_url') !== url || v?.get('image_hash') !== hash) && api.vcard.update(v, true));
```
jcbrand: There's a bug here.

`hash` will be `undefined` when only `image_url` is set, so the `hash &&` clause will prevent the VCard from being updated when `image_url` has changed.

src/shared/avatar/templates/avatar.js
```diff
@@ -5,7 +5,12 @@ import { html, nothing } from 'lit';
  * @param {string} image_type
```
jcbrand: `image_type` needs to be made optional.

```suggestion
 * @param {string} [image_type]
```

src/shared/avatar/avatar.js
```diff
@@ -29,23 +29,25 @@ export default class Avatar extends CustomElement {
...
             image_type = this.pickerdata.image_type;
             data_uri = this.pickerdata.data_uri;
         } else {
+            image_url = this.model?.vcard?.get('image_url');
             image_type = this.model?.vcard?.get('image_type');
             image = this.model?.vcard?.get('image');
         }
 
-        if (image_type && (image || data_uri)) {
+        if (image_type && (image_url || image || data_uri)) {
```
jcbrand: After the change I proposed above,  `image_type` will be `undefined` when `image_url` is set, so this needs to be updated to:


```suggestion
        if (image_type && (image || data_uri) || image_url) {
```

src/shared/avatar/templates/avatar.js
```diff
@@ -5,7 +5,12 @@ import { html, nothing } from 'lit';
...
  */
 const getImgHref = (image, image_type) => {
```
jcbrand: While you're at it, please make this a non-arrow function:

```suggestion
function getImgHref(image, image_type) {
```

jcbrand (changes requested):

src/headless/plugins/vcard/parsers.js
JohnXLivingston: > XEP-0054 and XEP-0292 both support `PHOTO EXTVAL` but funnily enough XEP-0153 VCard-based Avatars discourages it. I'm not sure why.

I guess it is to prevent IP leaking. An attacker could use this field to point to their server, and get other users IP. Which could be an issue when you have high privacy expectation.

jcbrand: Perhaps, but this is the same with HTTP file upload.

My guess is it's because that XEP deals with creating hashes of BINVAL values and advertising those in order to let clients know when they need to update the VCard.

When you're using a URL, presumably you'd change the URL when there's a new avatar so the usage of a hash is not necessary (but then you'd ideally advertise the new image URL similarly to how the hash is advertised).

JohnXLivingston: > Perhaps, but this is the same with HTTP file upload.

yes, and that's why some clients don't display images automatically.



> My guess is it's because that XEP deals with creating hashes of BINVAL values and advertising those in order to let clients know when they need to update the VCard.
> 
> When you're using a URL, presumably you'd change the URL when there's a new avatar so the usage of a hash is not necessary (but then you'd ideally advertise the new image URL similarly to how the hash is advertised)

Could be because of this too, indeed.

jcbrand: > yes, and that's why some clients don't display images automatically.

This is actually a good feature which I'd like to have in Converse as well.

edshot99: I didn't see an issue with IP leaks since many clients already block untrusted domains and i'd assume deployments of Converse would set up content security policy rules. I'm not sure if the same applies with Converse Desktop so that would have to be checked first.

If CSP isn't enough then creating some configuration to allow trusted domains for avatars can be created.

edshot99 (commented):
Everything should be fixed now.

src/headless/plugins/vcard/utils.js
```diff
@@ -40,12 +40,13 @@ export function createStanza(type, jid, vcard_el) {
...
 export function onOccupantAvatarChanged(occupant) {
     const hash = occupant.get('image_hash');
+    const url = occupant.get('image_url');
     const vcards = [];
     if (occupant.get('jid')) {
         vcards.push(_converse.state.vcards.get(occupant.get('jid')));
     }
     vcards.push(_converse.state.vcards.get(occupant.get('from')));
-    vcards.forEach((v) => hash && v && v?.get('image_hash') !== hash && api.vcard.update(v, true));
+    vcards.forEach((v) => (hash || url) && v && (v?.get('image_hash') !== hash || v?.get('image_url') !== url) && api.vcard.update(v, true));
```
jcbrand: This still doesn't look right, because of the `(hash || url)`  the `hash` value could be undefined, but then the `v?.get('image_hash') !== hash` check could still execute (because `url` is defined), potentially resulting in an unnecessary call to `api.vcard.update`.

I think it's better to write this logic out.


```suggestion
    vcards.filter((v) => v).forEach((v) => {      
        if (hash && v.get('image_hash') !== hash || url && v.get('image_url') !== url) {
            api.vcard.update(v, true);
        }
    });
```

src/headless/plugins/vcard/parsers.js
```diff
@@ -10,21 +10,33 @@ export async function parseVCardResultStanza(iq) {
...
         const ab = await crypto.subtle.digest('SHA-1', buffer);
-        result['image_hash'] = u.arrayBufferToHex(ab);
+
+        Object.assign(result, {
+            image,
+            image_type,
+            image_hash: u.arrayBufferToHex(ab),
+        });
+    }
+    else if (image_url) {
```
jcbrand: Nitpick: Please put the `else` on the same line as the curly bracket.

```suggestion
    } else if (image_url) {
```

src/shared/avatar/templates/avatar.js
```diff
@@ -2,10 +2,15 @@ import { html, nothing } from 'lit';
...
- * @param {string} image_type
+ * @param {string} [image_type]
  */
-const getImgHref = (image, image_type) => {
-    return image.startsWith('data:') ? image : `data:${image_type};base64,${image}`;
+function getImgHref(image, image_type) {
+    if (image.startsWith('https:') || image.startsWith('data:')) {
+        return image;
+    }
+    else {
```
jcbrand: Nitpick: Please put the `else` on the same line as the curly bracket.

```suggestion
    } else {
```

edshot99:
Added new changes.

jcbrand:
Thanks @edshot99. Can you please rebase your branch onto the latest `master` and fixed the conflict?

Also, can you please add a test here:
https://github.com/conversejs/converse.js/blob/b8c5137ef55de1377c8754a9c36ccfe6a897b76d/src/headless/plugins/vcard/tests/update.js#L5

It can be similar to the `'will cause a VCard avatar to be replaced'` test, with the difference that you check that an HTTP avatar is received.

To run that test, you can run `npm run test:headless` (to run just that one test, you can rename `it(` to `fit(`, just remove the `f` again before committing/pushing).

edshot99:
I'm still having issues with the tests. Specifically the vCard removal test so i'm just going to put this on hold until I get that sorted out when I have time.

jcbrand:
@edshot99 If you commit the tests to this PR, then I can take a look at what you've done so far and perhaps be able to help unblock you.

PR #3794, opened 2025-08-11
Add pubsub# prefix to field names

### LINKED ISSUE #3792: Node reconfiguration attempt uses incorrect field names
**Describe the bug**
When attempting to re-configure a pubsub node, converse.js downloads the current configuration, merges the new configuration into the existing one, then re-submits the configuration. 

When downloading and parsing the existing configuration, the parser removes the `pubsub#` prefix from the field names.  This prefix does not get re-added when submitting the new configuration resulting in the new configuration being ignored by the server.

Tested with Openfire

kicken:
Fixes #3792


jcbrand:
Thanks, I've merged manually

Issue #3815, opened 2025-08-26
Removal of boomark leads to new bookmark named 'Symbol(lit-nothing)'


ChaosKid42:
**Describe the bug**
Every time I try to remove a bookmark (by clicking on the little 'x') the bookmark is remove from the list. After logout and login, the bookmark is still there and a new bookmark appears. The boommark is named 'Symbol(lit-nothing)' (see attached screenshot).

**To Reproduce**
1. Go to 'Groupchats-Bookmarks'
2. Click on the little 'x' to confirm removal of bookmark
3. Confirm removal of bookarm in dialog
4. Log out of converse
5. Log in
6. Go to 'Groupchats-Bookmarks'

**Expected behavior**
The bookmark is supposed to be removed permanently removed even after logout and login.

**Screenshots**
<img width="517" height="883" alt="Image" src="https://github.com/user-attachments/assets/1bd7dca3-23f2-416c-8e8d-379f103fc405" />

**Environment (please complete the following information):**
 - Desktop
 - Browser Edge Chromium
 - Converse.js version v11.0.1

**Additional context**
Using ejabberd 25.08

PR #3821, opened 2025-08-30
Add proper native bookmarks deletion

yureitzk:
This PR addresses an issue where the old bookmark deletion function was incorrectly adding empty bookmarks to user's PubSub service.  This was due to an invalid `item` used for deletion and a type mismatch in the `getPublishedItems` function.  This change ensures proper bookmark deletion when the server adheres to XEP-0402.

Related issue: https://github.com/conversejs/converse.js/issues/3815

### DIFF
--- a/src/headless/plugins/bookmarks/collection.js
+++ b/src/headless/plugins/bookmarks/collection.js
@@ -37,8 +37,14 @@ class Bookmarks extends Collection {
         this.on('change:autojoin', this.onAutoJoinChanged, this);
         this.on(
             'remove',
-            /** @param {Bookmark} bookmark */
-            (_, bookmark) => this.sendBookmarkStanza(bookmark),
+            /** @param {Bookmark} bookmarks */
+            async (bookmark, bookmarks) => {
+                const bare_jid = _converse.session.get('bare_jid');
+                const sourceBookmark = (await api.disco.supports(`${Strophe.NS.BOOKMARKS2}#compat`, bare_jid))
+                    ? bookmark
+                    : bookmarks;
+                this.removeBookmarkStanza(sourceBookmark)
+            },
             this
         );
 
@@ -125,6 +131,36 @@ class Bookmarks extends Collection {
         if (send_stanza) {
             this.sendBookmarkStanza(bookmark).catch((iq) => this.onBookmarkError(iq));
         }
+    }
+
+    /**
+     * @param {Bookmark} bookmark
+     * @returns {Promise<void|Element>}
+     */
+    async removeBookmarkStanza (bookmark) {
+        const bare_jid = _converse.session.get('bare_jid');
+
+        const node = (await api.disco.supports(`${Strophe.NS.BOOKMARKS2}#compat`, bare_jid))
+            ? Strophe.NS.BOOKMARKS2
+            : Strophe.NS.BOOKMARKS;
+
+        if (node === Strophe.NS.BOOKMARKS2) {
+
+            const retractItem = stx`
+                <item id="${bookmark.get('jid')}">
+                    <remove/>
+                </item>`;
+
+            return api.pubsub.publish(
+                null,                         
+                node,        
+                retractItem,
+                { persist_items: true }       
+            );
+
+        }
+
+        return this.sendBookmarkStanza(bookmark);
     }
 
     /**
--- a/src/headless/types/plugins/bookmarks/collection.d.ts
+++ b/src/headless/types/plugins/bookmarks/collection.d.ts
@@ -25,6 +25,11 @@ declare class Bookmarks extends Collection {
      */
     setBookmark(attrs: import("./types").BookmarkAttrs, create?: boolean, options?: object): void;
     /**
+     * @param {Bookmark} bookmark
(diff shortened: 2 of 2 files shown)


src/headless/plugins/bookmarks/collection.js
```diff
@@ -37,8 +37,14 @@ class Bookmarks extends Collection {
         this.on('change:autojoin', this.onAutoJoinChanged, this);
         this.on(
             'remove',
-            /** @param {Bookmark} bookmark */
-            (_, bookmark) => this.sendBookmarkStanza(bookmark),
+            /** @param {Bookmark} bookmarks */
```
jcbrand, line 40:
```suggestion
            /** @param {Bookmark} bookmarks */
```

src/headless/plugins/bookmarks/collection.js
```diff
@@ -37,8 +37,14 @@ class Bookmarks extends Collection {
...
+            async (bookmark, bookmarks) => {
+                const bare_jid = _converse.session.get('bare_jid');
+                const sourceBookmark = (await api.disco.supports(`${Strophe.NS.BOOKMARKS2}#compat`, bare_jid))
+                    ? bookmark
+                    : bookmarks;
+                this.removeBookmarkStanza(sourceBookmark)
+            },
```
jcbrand, line 47: Thanks for your pull request!

From what I can tell, the code looks correct, but I have a few suggestions regarding structure.

I think here we can have:

```
this.on('remove', this.removeBookmark, this);
```

Note: `removeBookmarkStanza` makes it sound like a stanza is being removed, which is not the case.

Then in `removeBookmark`, you can put the logic above to determine `sourceBookmark`.

BTW, in this codebase we name variables with snake-case, so `sourceBookmark` should be `source_bookmark`.
I know this is not how it's done in most JS codebases, but this repo started 10 years ago before JS conventions were well established and for consistency's sake we should keep to this convention.

Also, would you be willing to look into adding tests for this?
You can add the tests here: https://github.com/conversejs/converse.js/blob/master/src/headless/plugins/bookmarks/tests/bookmarks.js

There are multiple examples which should hopefully make it clear to you how to write a test.

src/headless/plugins/bookmarks/collection.js
```diff
@@ -127,6 +133,36 @@ class Bookmarks extends Collection {
...
+        const node = (await api.disco.supports(`${Strophe.NS.BOOKMARKS2}#compat`, bare_jid))
+            ? Strophe.NS.BOOKMARKS2
+            : Strophe.NS.BOOKMARKS;
+
+        if (node === Strophe.NS.BOOKMARKS2) {
+
+            const retractItem = stx`
+                <item id="${bookmark.get('jid')}">
+                    <remove/>
+                </item>`;
```
jcbrand, line 152: Where did you get this markup?

According to the XEP it should look like this:
```
    <retract node='urn:xmpp:bookmarks:1' notify='true'>
      <item id='theplay@conference.shakespeare.lit'/>
    </retract>
```
See: https://xmpp.org/extensions/xep-0402.html#removing-a-bookmark

yureitzk: I initially put it in as a placeholder but forgot to replace it with the code from the spec because it didn't throw any errors and actually worked. I couldn't find any mentions of it being officially supported, except for the fact that a similar syntax is used for account deletion (where I got it from) and removing a contact from the WaitingList, though.

The corrected version looks like this:

```js
if (node === Strophe.NS.BOOKMARKS2) {

	const stanza = stx`
	<iq type="set" from="${api.connection.get().jid}" xmlns="jabber:client">
	  <pubsub xmlns='http://jabber.org/protocol/pubsub'>
		<retract node="${node}">
		  <item id="${bookmark.get('jid')}"/>
		</retract>
	  </pubsub>
	</iq>`;

	return await api.sendIQ(stanza);
}
```

If that's okay, I can change the stanza to this in my commit

src/headless/plugins/bookmarks/collection.js
yureitzk: Yes, no problem. As for the tests, would it be okay if I make 2 separate tests for both `Strophe.NS.BOOKMARKS2` and `Strophe.NS.BOOKMARKS`?

jcbrand: > Yes, no problem. As for the tests, would it be okay if I make 2 separate tests for both Strophe.NS.BOOKMARKS2 and Strophe.NS.BOOKMARKS?

Yes that would be great thanks!

We put the BOOKMARKS tests inside a file called `deprecated.js`, see here for example in the `bookmark-views` plugin:
https://github.com/conversejs/converse.js/blob/master/src/plugins/bookmark-views/tests/deprecated.js

So you'll have to create that file in `src/headless/plugins/bookmarks/tests/` and put those tests there.

src/headless/plugins/bookmarks/collection.js
jcbrand: Yes please!

jcbrand:
Fixed in #3912

PR #3822, opened 2025-08-30
Fix missing room display name when a room has no specified name

yureitzk:
This pull request addresses a bug that occurs when a room initially lacks a name. After logging out and logging back, the room's name displayed in bookmarks will show as `nothing` due to how the bookmark data was uploaded to the server. 

The `getDisplayName` function doesn't check if the room’s name is still missing when a bookmark is created. The `get()` method always returns a string, so there's no automatic fallback to JID.

I don't think it's the only way to solve this issue, but look like the easiest one.

### DIFF
--- a/src/headless/plugins/muc/muc.js
+++ b/src/headless/plugins/muc/muc.js
@@ -1281,6 +1281,7 @@ class MUC extends ModelWithVCard(ModelWithMessages(ColorAwareModel(ChatBoxBase))
             this.save({ name: identity.get('name') });
         } else {
             log.error(`No identity or name found for ${this.get('jid')}`);
+            this.save({ name: this.get('jid') });
         }
         await this.getDiscoInfoFields();
         await this.getDiscoInfoFeatures();


src/headless/plugins/muc/muc.js
```diff
@@ -1281,6 +1281,7 @@ class MUC extends ModelWithVCard(ModelWithMessages(ColorAwareModel(ChatBoxBase))
             this.save({ name: identity.get('name') });
         } else {
             log.error(`No identity or name found for ${this.get('jid')}`);
+            this.save({ name: this.get('jid') });
```
jcbrand, line 1284: Thanks for the pull request @yureitzk 

In the past we used to store the `JID` as a fallback value for `name` for different entities, but stopped doing so because strictly speaking the JID is not the name and it's better to have an `undefined` name to make it clear that there is no known `name` value.

Instead, we have the `getDisplayName` method (which you mentioned in your PR description) that gets called to get the correct name to display for a entity (e.g. a MUC, occupant, roster contact or 1:1 chat).

So I think the fix needs to go into `getDisplayName` or code related to it.

yureitzk: I was thinking about this as well, but `getDisplayName` will not fallback to `JID`, since if `name` attribute is missing, `get` will return `Symbol(lit-nothing) string`. But the main issue was that the bookmark's name is also `Symbol(lit-nothing)` . You'll see this upon the next login, when converse will fetch the bookmarks from the server. 

Also, when you click on this empty bookmark, the room's name in main UI will be set to the same empty string. 

This problem will go away if the chat will get a proper `name` aka `title`.

jcbrand: > I was thinking about this as well, but getDisplayName will not fallback to JID, since if name attribute is missing, get will return Symbol(lit-nothing) string

Yes, so a better fix (more like a workaround until #3815 is fixed) is to check whether a `Symbol(lit-nothing)` string is returned and in that case fall back to the JID.

It's ugly, but it prevents us from having to set the `name` to the JID, which I'd like to avoid.

Neustradamus:
@yureitzk: Have you progressed on it?

PR #3823, opened 2025-08-31
fix remove bookmark

ChaosKid42:
fixes https://github.com/conversejs/converse.js/issues/3815

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 12.0.1 (Unreleased)
+
+- #3823: Bookmark removal not working
+
 ## 12.0.0 (2025-08-28)
 
 - #3581: Don't unnecessarily regenerate pot and po files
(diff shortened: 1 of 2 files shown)


jcbrand:
Thanks @ChaosKid42 for making this PR.

There's a slightly older PR to fix #3815 here:
https://github.com/conversejs/converse.js/pull/3821

I think that PR is more correct because with XEP-0048 PEP it's a single item that contains all bookmarks which needs to be updated and so a `retract` element shouldn't be sent if old XEP-0048 bookmarks are being used.

ChaosKid42:
I don't mind if the other PR is used. Whatever works is fine for me.

I thought BOOKMARKS2 was associated with XEP-0402. The examples in the document make use of retract. That's the reason I implemented it the way it is.

jcbrand:
> I thought BOOKMARKS2 was associated with XEP-0402.

Yes, but the fallback BOOKMARKS uses XEP-0048 (with PEP).

ChaosKid42:
Ah I see. The other PR is supposed to work for XEP-048, too.

So I basically agree with your assessment with one exception: Shouldn't the other PR use retract in case of XEP-0402?

jcbrand:
Oh, I missed that difference. I've asked for clarification: https://github.com/conversejs/converse.js/pull/3821/files#r2315090585

PR #3836, opened 2025-09-09
Fix a typo

marclaporte:

PR #3837, opened 2025-09-09
Run codespell on the whole codebase

linkmauve:
I’ve had to ignore some words (“pres” is used a lot to abbreviate “presence”, as well as some proper names), but overall it reduces the amount of typos by a lot.

It didn’t find them all though, for instance the one from #3836 hasn’t been fixed, but it’s still a valid step towards more proper English.

In the future it could make sense to integrate `codespell` as a CI pass, to avoid reintroducing typos.


jcbrand:
Thanks!

Issue #1247, opened 2018-10-12
Register xmpp protocol handler


Ppjet6:
## Steps to reproduce

- On any other website, or application, click on an `xmpp:` uri. e.g., `xmpp:discuss@conference.conversejs.org?join` (github won't convert this into a link)

## What I expect

A  popup like the following appears, and conversejs is available in the list:
![launch-application](https://user-images.githubusercontent.com/138302/46888066-edf7a180-ce4e-11e8-9fb1-e57ea75dfc80.png)

Movim also displays the url of the instance, that I have greyed out here. But I assume it's just the browser adding some context to the entry, saying which website has registered the handler.

## What happens

Converse doesn't appear in the list.

## Misc

Implementation wise, the API to use seems to be [navigator.registerProtocolHandler](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler/Web-based_protocol_handlers)

This is what it looks like in Firefox:
![add-application](https://user-images.githubusercontent.com/138302/46888698-be499900-ce50-11e8-9916-b3e4d92d686d.png)

A few things to consider before implementing:
- Not all conversejs instance will want to use this. I hardly see a need for a customer support platform for example, they will certainly want it off. The feature should probably be behind a configuration flag?
- Every conversejs instance I navigate to will start asking me to register the protocol handler. Is that something we want?
- Remembering that the user didn't add the handler might also be useful to avoid spamming them, (that happens too often in web applications).
- In the future, if this is implemented, removing the `registerProtocolHandler` call doesn't mean the handler is gone. I don't think there is a way to remove it even, from the application itself. That seems to be a user action, in the browser settings.

jcbrand:
Related to #537

jcbrand:
[Web based protocol handlers](https://developer.mozilla.org/en-US/docs/Web-based_protocol_handlers) might be useful for this.

marcellintacite:
I am working on this issue

PR #3852, opened 2025-09-20
Web link for XMPP address via xmpp.link

yurtpage:
The biggest problem of the `xmpp:` links is that they won't work without installed client. So here is web based https://xmpp.link/
similar to Matrix.to or t.me for Telegram.
In the PR I added the `web` link that can be copied by a user shared.


marclaporte:
Thank you @yurtpage 

What happens if this domain name is not renewed in a few years?

yurtpage:
In 99% this won't be a problem because a typical usage is a short living invitations. For the long standing links it's more likely that a user or channel disappear than the domain expires.
The service provided by [Modern XMPP](https://docs.modernxmpp.org/) which is same people as Prosody and Snikket, so it can be trusted to some extent.
Sources are here https://github.com/modernxmpp/easy-xmpp-invitation
The Kaidan client has own instance of the service on own domain https://i.kaidan.im/
The Conversations client also has "HTTP link" to share your account but it uses own service that is not open sourced.

Sorry, I had to make it configurable and translatable but I can do this only later.

jcbrand:
Thanks @yurtpage, but I think a better approach would be to open those links inside Converse (which AFAIK is doable), given that the user is already seeing them inside an instance of Converse.

It doesn't make sense to me to be inside Converse, click an XMPP link and then be taken somewhere else...

marclaporte:
Ok then, perhaps material for a "Good first issue"?

jcbrand:
@marclaporte There is actually an old ticket for this: https://github.com/conversejs/converse.js/issues/1247

I've added the `Good first issue` label, although I'm not sure the complexity involved.

yurtpage:
This a link to share so you can send to a friend or publish in internet. The Converse.js have nothing to do with them

jcbrand:
>  This a link to share so you can send to a friend or publish in internet. The Converse.js have nothing to do with them

Ok, but the fields that you modified in this PR don't say that, they say "XMPP Address" and with your PR's change the value that is shown for that field is not an XMPP address anymore but a web URL.

If we want to add these links, then I think they should go in a new field, called something like "Shareable link", so that it's clear that this is a URL meant to be shared with others.

This option would also have to be configurable, since Converse can be used in setups where other clients cannot be used, in which case such a link is not usable and confusing.

yurtpage:
Let's make two buttons instead "Copy address" and "Copy invitation web link".

jcbrand:
>  Let's make two buttons instead "Copy address" and "Copy invitation web link".

I don't think the buttons should have these labels, instead we can just use normal labels and then have a small button with a copy icon next to the value.

For example: :point_down: 

image

yurtpage:
I agree, two buttons with copy icon and the "web copy" with 🔗 or 🌐 as an icon or maybe even it should be a "Share ➤" dialog with other IMs (TG, WhatsApp), but not sure if this is possible with web app.
The important thing is that links to the https://xmpp.link/#juliet@capulet.com should be replaced with xmpp:juliet@capulet.com in chat window, so that a user can click and and open chat directly in the Converse without leaving to the xmpp.link

deleolajide:
> but not sure if this is possible with web app.

See https://developer.mozilla.org/en-US/docs/Web/API/Web_Share_API

jcbrand:
@yurtpage Are you planning on still finishing this PR?

Issue #3800, opened 2025-08-13
Sponsor logos overlapping on Contacts page


BalramMardi:
**Describe the bug**
On the Contacts page, the sponsor logos are overlapping each other, causing a messy layout and making the logos hard to see. This affects the page’s visual clarity and user experience.

**To Reproduce**
1. Open the Contacts page. [https://conversejs.org/#contact](https://conversejs.org/#contact)
2. Observe the sponsor logos section.

**Expected behavior**
Sponsor logos should be displayed side by side or properly spaced without overlapping, and the layout should be responsive across different screen sizes.

**Screenshots**

<img width="1895" height="910" alt="Image" src="https://github.com/user-attachments/assets/34e4a4a9-19e8-4360-ba94-6aee8119b88d" />

**Environment (please complete the following information):**
 - Chrome, Firefox

PR #3858, opened 2025-09-23
[FIX] Sponsor logos overlapping on Contacts page

marcellintacite:
Here's the issue : https://github.com/conversejs/converse.js/issues/3800

The original layout used Bootstrap's grid system (row, row-cols-*, and col), which was causing the logos to shrink and overlap before they wrapped. This was due to the default flex-shrink behavior of the col class.

This PR replaces the grid classes with a simpler, more flexible flexbox approach. The parent ul element now uses d-flex and flex-wrap, allowing the list items  to wrap to a new line naturally when the container shrinks.
By also adding a gap-4 class to the container, we ensure there is consistent space between the logos, preventing them from touching or overlapping. 


jcbrand:
Hi @marcellintacite, could you please fix the merge conflict in this branch?

marcellintacite:
> Hi @marcellintacite, could you please fix the merge conflict in this branch?

I am working on it and also i have issue with test. I am trying to figure it out

marclaporte:
There is only a change in CHANGES.md now

PR #3865, opened 2025-09-27
Added docker file + fix on translations

loviuz:
Hi, I added the files to make the project work with Docker. I also added the option to translate the months and days in the chat history, and completed the Italian string update. Sorry for creating a single PR... :-/


jcbrand:
Hi @loviuz 

Thank you for your pull request.

We have a repo with a Docker setup here: https://github.com/conversejs/converse.js-docker
Have you seen it?

Perhaps instead of adding a Dockerfile in this repo, we could update the docs to point people to that repo and if necessary improve and modernize it.

> Sorry for creating a single PR... :-/

Yes, this should ideally have been 3 different PRs.

jcbrand:
Another thing to mention here is that all translations should be done via weblate:
https://hosted.weblate.org/projects/conversejs/

The reason for this is because if we allow translations to happen via PRs and via Weblate, then we get merge conflicts which can be difficult and time-consuming to resolve.

If we keep translations in Weblate, we avoid this problem.

jcbrand:
Hi @loviuz, thanks again for your pull request.

However, I'm going to close it now for the following reasons:

* The problem that months and days weren't translated has been fixed here: https://github.com/conversejs/converse.js/commit/4fd4b0dffd3037a7f3c791781ebef1ad0a431bba, by fixing how DayJS is instantiated.
* The Italian translations should go in via Weblate
* I'm not convinced we should have a Dockerfile inside this repo. Docker is very useful, but people will use it in different ways. Some people might want to mount the `src` directory from the host system, while others don't. Some might want other Docker containers with an XMPP server, while others don't. Besides that, there is a repo with a Docker setup, namely https://github.com/conversejs/converse.js-docker

Despite that, we appreciate your contribution and hope you'll be willing to make more pull requests in the future if you find something you'd like to fix or improve.

PR #3870, opened 2025-10-01
Adding support for protocol handler

marcellintacite:
Here's a draft for this issue : #1247 
This merge request adds support for handling xmpp: protocol links in Converse.js, allowing users to click links like xmpp:user@example.com to automatically open a chat with the specified JID if logged in. This enhances user experience by enabling seamless integration with external XMPP links (e.g., from emails, websites, or other apps).

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -102,14 +102,15 @@ before the server's replay of queued stanzas is handled.
 - #3023: Fix empty roster group name treated as a valid group instead of "Ungrouped"
 - #3382: Fix audio player accessibility for screen reader users (NVDA, JAWS)
 - #3502: Fix custom emoji shortnames overlapping with built-in emoji shortnames
-- #3815: Removal of bookmark leads to new b
+- #3815: Removal of bookmark leads to new bookmark named `Symbol(lit-nothing)`
+- #3889: MUC join: Use room jids localpart as name in case name or identity not found
+- #3824: Dates and times are not translated
 - #3829: Rich text from LibreOffice Calc is sent as screenshots
 - #3830: The message textarea blocks undo of the pasted text
 - #3839: Fix contacts menu displayed off-screen in fullscreen mode
 - #3863: Shift `unescapeHTML` helper function (used to set `isOnlyEmoji` on a message) to converse-headless
 - #3885: Don't show contact approval alerts when `allow_contact_requests` is false (e.g. anonymous mode)
 - #3886: Don't show OMEMO padlock icon when libsignal is not available (e.g. anonymous mode)
-- #3889: MUC join: Use room jids localpart as name in case name or identity not found
 - #3710: Fix autocomplete adding a trailing space when selecting a group name with the mouse
 - #3916: Add support for XEP-0461 Message Replies, allowing users to reply to specific messages
 - #3939: Don't show invitations to groupchats in which the user is already present
@@ -128,6 +129,7 @@ before the server's replay of queued stanzas is handled.
 - #3769: Various OMEMO fixes
 - #3791: Fetching pubsub node configuration fails
 - #3792: Node reconfiguration attempt uses incorrect field names
+- Adds support for opening XMPP URIs in Converse and for XEP-0147 query actions
 - Fix documentation formatting in security.rst
 - Add approval banner in chats with requesting contacts or unsaved contacts
 - Add mongolian as a language option
--- a/conversejs.doap
+++ b/conversejs.doap
@@ -130,6 +130,15 @@
     </implements>
     <implements>
       <xmpp:SupportedXep>
+        <xmpp:xep rdf:resource="https://xmpp.org/extensions/xep-0147.html"/>
+        <xmpp:status>partial</xmpp:status>
+        <xmpp:version>1.2</xmpp:version>
+        <xmpp:since>14.0.0</xmpp:since>
+        <xmpp:note>Supports XMPP URI scheme with query actions for message and roster management</xmpp:note>
+      </xmpp:SupportedXep>
+    </implements>
+    <implements>
+      <xmpp:SupportedXep>
         <xmpp:xep rdf:resource="https://xmpp.org/extensions/xep-0153.html"/>
         <xmpp:since>11.0.0</xmpp:since>
         <xmpp:status>complete</xmpp:status>
--- a/manifest.json
+++ b/manifest.json
@@ -26,5 +26,11 @@
   "background_color": "#397491",
   "display": "standalone",
   "scope": "/",
-  "theme_color": "#397491"
+  "theme_color": "#397491",
+  "protocol_handlers": [
+    {
+      "protocol": "xmpp",
+      "url": "#converse/action?uri=%s"
+    }
+  ]
 }
--- a/src/index.js
+++ b/src/index.js
@@ -12,6 +12,7 @@ import { CustomElement } from 'shared/components/element.js';
 import { VIEW_PLUGINS } from './shared/constants.js';
 import { _converse, converse } from '@converse/headless';
 import './utils/index.js';
+import { routeToQueryAction } from './plugins/chatview/utils.js';
 
 _converse.__ = i18n.__; // DEPRECATED
 Object.assign(converse.env, { i18n });
@@ -55,6 +56,16 @@ import './plugins/reactions-views/index.js'; // XEP-0444 Reactions
 
 _converse.exports.CustomElement = CustomElement;
 
+// Register XMPP protocol handler for xmpp: URIs
+if ('registerProtocolHandler' in navigator) {
+    try {
+        const handlerUrl = `${window.location.origin}${window.location.pathname}#converse/action?uri=%s`;
+        navigator.registerProtocolHandler('xmpp', handlerUrl);
+    } catch (error) {
+        console.warn('Failed to register protocol handler:', error);
+    }
+}
+
 const initialize = converse.initialize;
 
 converse.initialize = function (settings, callback) {
@@ -63,7 +74,13 @@ converse.initialize = function (settings, callback) {
     } else {
         settings.whitelisted_plugins = VIEW_PLUGINS;
     }
-    return initialize(settings, callback);
+    // Handle XEP-0147 query actions after initialization
+    // This must happen after plugins are loaded and _converse.env is ready
+    const result = initialize(settings, callback);
+    if (result && result.then) {
+        result.then(() => routeToQueryAction());
+    }
+    return result;
 };
 
 window['converse'] = converse;
--- a/src/plugins/chatview/tests/query-actions.js
+++ b/src/plugins/chatview/tests/query-actions.js
@@ -1,0 +1,113 @@
+import mock from '../../../shared/tests/mock.js';
+import converse from '../../../../dist/converse.js';
+
+const { u } = converse.env;
+
+describe("XMPP URI Query Actions (XEP-0147) - URI Parsing", function () {
+
+    /**
+     * Test URI extraction and parsing functionality
+     * This tests the extractXMPPURI and parseXMPPURI functions
+     */
+    it("extracts and parses XMPP URI correctly",
+        mock.initConverse(converse, ['chatBoxesFetched'], {}, async function (_converse) {
+
+        const originalHash = window.location.hash;
+        const originalReplaceState = window.history.replaceState;
+
+        // Spy on history.replaceState to verify URL cleanup
+        const replaceStateSpy = jasmine.createSpy('replaceState');
+        window.history.replaceState = replaceStateSpy;
+
+        // Simulate a protocol handler URI by setting the hash
+        window.location.hash = '#converse/action?uri=xmpp%3Aromeo%40montague.lit%3Fmessage%3Bbody%3DHello';
+
+        try {
+            // Call the function - this should extract and parse the URI
+            await u.routeToQueryAction();
+
+            // Verify that the URL was cleaned up (protocol handler removes ?uri=...)
+            const expected_url = `${window.location.origin}${window.location.pathname}`;
+            expect(replaceStateSpy).toHaveBeenCalledWith({}, document.title, expected_url);
+        } finally {
+            // Restore original globals to avoid test pollution
+            window.location.hash = originalHash;
+            window.history.replaceState = originalReplaceState;
+        }
+    }));
+
+    /**
+     * Test that routeToQueryAction triggers xmppURIAction event
+     * This tests the event-based delegation to plugin-specific handlers
+     */
+    it("triggers xmppURIAction event with parsed data",
+        mock.initConverse(converse, ['chatBoxesFetched'], {}, async function (_converse) {
+
+        const { api } = _converse;
+        const originalHash = window.location.hash;
+        const originalReplaceState = window.history.replaceState;
+
+        window.history.replaceState = jasmine.createSpy('replaceState');
+
+        // Track if the event was triggered with correct data
+        let eventTriggered = false;
+        let eventData = null;
+
+        api.listen.on('xmppURIAction', (data) => {
+            eventTriggered = true;
+            eventData = data;
+        });
+
+        // Mock URI with message action
+        window.location.hash = '#converse/action?uri=xmpp%3Aromeo%40montague.lit%3Fmessage%3Bbody%3DHello';
+
+        try {
+            // Execute the function
+            await u.routeToQueryAction();
+
+            // Verify that xmppURIAction event was triggered
+            expect(eventTriggered).toBe(true);
+            expect(eventData.jid).toBe('romeo@montague.lit');
+            expect(eventData.action).toBe('message');
+            expect(eventData.query_params.get('body')).toBe('Hello');
+        } finally {
+            window.location.hash = originalHash;
+            window.history.replaceState = originalReplaceState;
+        }
+    }));
+
+    /**
+     * Test handling of invalid JIDs in URI parsing
+     * This ensures the function gracefully handles malformed JIDs
+     */
+    it("handles invalid JID gracefully and does not trigger event",
+        mock.initConverse(converse, ['chatBoxesFetched'], {}, async function (_converse) {
+
+        const { api } = _converse;
+        const originalHash = window.location.hash;
+        const originalReplaceState = window.history.replaceState;
+
+        window.history.replaceState = jasmine.createSpy('replaceState');
+
+        // Track if the event was triggered
+        let eventTriggered = false;
+
+        api.listen.on('xmppURIAction', () => {
+            eventTriggered = true;
+        });
+
+        // Mock URI with invalid JID (missing domain)
+        window.location.hash = '#converse/action?uri=xmpp%3Ainvalid-jid';
+
+        try {
+            // Execute the function
+            await u.routeToQueryAction();
+
+            // Verify that xmppURIAction event was NOT triggered for invalid JID
+            expect(eventTriggered).toBe(false);
+        } finally {
+            window.location.hash = originalHash;
+            window.history.replaceState = originalReplaceState;
+        }
+    }));
+});
--- a/src/plugins/chatview/utils.js
+++ b/src/plugins/chatview/utils.js
@@ -1,5 +1,7 @@
 import { __ } from 'i18n';
-import { _converse, api } from '@converse/headless';
+import { _converse, api,u } from '@converse/headless';
+import log from "@converse/log";
+
 
 export function clearHistory (jid) {
     if (location.hash === `converse/chat?jid=${jid}`) {
@@ -71,3 +73,85 @@ export function resetElementHeight (ev) {
         ev.target.style = '';
     }
 }
+
+
+/**
+ * Handle XEP-0147 "query actions" invoked via xmpp: URIs.
+ * Extracts URI from window location, parses it, and triggers xmppURIAction event
+ * for plugin-specific handling via api.listen.
+ *
+ * @param {Event} [event] - Optional event object (used when called as protocol handler)
+ * @returns {Promise<void>}
+ * @example
+ * // Automatically called on initialization in src/index.js
+ * // Can be called manually after location.hash changes:
+ * window.location.hash = '#converse/action?uri=xmpp%3Auser%40example.com';
+ * await routeToQueryAction();
+ */
+export async function routeToQueryAction(event) {
+    const { u } = _converse.env;
+
+    const uri = extractXMPPURI(event);
+    if (!uri) return;
+
+    const { jid, action, query_params } = parseXMPPURI(uri);
+    if (!u.isValidJID(jid)) {
+        return log.warn(`routeToQueryAction: Invalid JID: "${jid}"`);
+    }
+    
+    // Trigger event to let specific plugins handle plugin-specific actions
+    api.trigger('xmppURIAction', { jid, query_params, action });
+}
+
+/**
+ * Extracts and decodes the xmpp: URI from the window location or hash.
+ */
+export function extractXMPPURI(event) {
+    let uri = null;
+    // hash-based (#converse/action?uri=...)
+    if (location.hash.startsWith('#converse/action?uri=')) {
+        event?.preventDefault();
+        uri = location.hash.split('#converse/action?uri=').pop();
+    }
+
+    if (!uri) return null;
+
+    // Decode URI and remove xmpp: prefix
+    uri = decodeURIComponent(uri);
+    if (uri.startsWith('xmpp:')) uri = uri.slice(5);
+
+    // Clean up URL (remove ?uri=... for a clean view)
+    const clean_url = `${window.location.origin}${window.location.pathname}`;
+    window.history.replaceState({}, document.title, clean_url);
+
+    return uri;
+}
+
+/**
+ * Splits an xmpp: URI into a JID, an action (querytype), and query parameters.
+ * XEP-0147 query parameters are separated by ';' instead of '&'.
+ */
+export function parseXMPPURI(uri) {
+    const [jid, query] = uri.split('?');
+    let action = null;
+    const query_params = new URLSearchParams();
+
+    if (query) {
+        const parts = query.split(';');
+        if (parts.length > 0) {
+            action = parts[0]; // The first part is the action (e.g., 'roster')
+            for (let i = 1; i < parts.length; i++) {
+                const [key, value] = parts[i].split('=');
+                if (key) {
+                    query_params.set(key, value ? decodeURIComponent(value) : '');
+                }
+            }
+        }
+    }
+    
+    return { jid, action, query_params };
+}
+
+Object.assign(u,{
+    routeToQueryAction,
+})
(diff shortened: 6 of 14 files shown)


src/index.js
```diff
@@ -42,6 +42,7 @@ import "./plugins/rosterview/index.js";
 import "./plugins/singleton/index.js";
 import "./plugins/dragresize/index.js";     // Allows chat boxes to be resized by dragging them
 import "./plugins/fullscreen/index.js";
+import "./plugins/protocol-handler/index.js"; // Handle xmpp: links
```
jcbrand: I don't think this functionality deserves its own plugin.

Converse already supports routing to a chat from the URL.

See here: https://github.com/conversejs/converse.js/blob/586649f682e47fa41730a61d627a5afc00ca5bb3/src/headless/plugins/chat/utils.js#L20

Perhaps that function can be updated and re-used?

src/plugins/protocol-handler/index.js
```diff
@@ -0,0 +1,58 @@
...
+ */
+import { api, converse } from '@converse/headless';
+
+converse.plugins.add('converse-protocol-handler', {
+    dependencies: [],  // No dependencies needed
+
+    initialize () {
+        let jidFromUrl = null;  // Store the JID from the URL for later use
+        // The first step is to register the protocol handler on app initialization
+        if ('registerProtocolHandler' in navigator) {
```
jcbrand: As mentioned in my other comment, I don't think this code should go into its own plugin.

Given that the goal is to open chats with JIDs, the code in this file can probably go to `src/plugins/chatview/index.html`.

src/plugins/protocol-handler/index.js
```diff
@@ -0,0 +1,58 @@
...
+            try {
+                // Defining the URL pattern for the protocol handler so that the browser knows where to redirect
+                // when an xmpp: link is clicked. The %s will be replaced by the full JID, using /dev as by default it was redirecting to the website.
+                const handlerUrl = `${window.location.origin}/dev?jid=%s`;
```
jcbrand: You can't hardcode to `dev` since that's not necessarily the path being used. You'll have to find another solution here. Perhaps just getting the path from `window.location` and using that. Another alternative is to add a configuration setting with the path to be used.

src/index.js
marcellintacite: I have updated the function

src/plugins/protocol-handler/index.js
marcellintacite: hello @jcbrand , i have updated codes, you can take a look

src/headless/plugins/chat/utils.js
```diff
@@ -22,7 +22,16 @@ export function routeToChat (event) {
...
     event?.preventDefault();
-    const jid = location.hash.split('=').pop();
+    let jid = location.hash.split('=').pop();
+    // decodeURIComponent is needed in case the JID contains special characters
+    // that were URL-encoded, e.g. `user%40domain` instead of `user@domain`.
+    jid = decodeURIComponent(jid);
+
+    // Remove xmpp: prefix if present
+    if (jid.startsWith('xmpp:')) {
+        jid = jid.slice(5);
```
jcbrand: Thanks @marcellintacite, your PR looks better now.

However, there are still various use-cases from XEP-0147 which aren't covered.

For example the message action:
https://xmpp.org/extensions/xep-0147.html#actions-message

And roster management actions:
https://xmpp.org/extensions/xep-0147.html#actions-roster-add

In the very least, you need to remove the query string (the part after `?`) from the xmpp URI, like you removed the `xmpp:` prefix, but ideally you could also implement the actions from XEP-0147.

src/plugins/chatview/index.js
```diff
@@ -65,6 +65,14 @@ converse.plugins.add('converse-chatview', {
         Object.assign(_converse, exports); // DEPRECATED
         Object.assign(_converse.exports, exports);
 
+        if ('registerProtocolHandler' in navigator) {
+        try {
+            const handlerUrl = `${window.location.origin}${window.location.pathname}#converse/chat?jid=%s`;
```
jcbrand: On second thought, given that XEP-0147 handles more use-cases than just opening a chat with a JID, I think we should perhaps create another URL with its own handler.

Something like ``${window.location.origin}${window.location.pathname}#converse/action?uri=%s``

And then instead of `routeToChat`, the handler could be called `routeToQueryAction` (because XEP-0147 uses the terminology "query actions").

src/index.js
jcbrand: Thank you! I've left two more review comments.

Let me know if you need help with this one: https://github.com/conversejs/converse.js/pull/3870#discussion_r2405198717

src/headless/plugins/chat/utils.js
marcellintacite: Thanks, i will learn more about it and if i have question, i'll reach out

marcellintacite: Hello @jcbrand , would check the implementation, i have added support for sending message and adding a contact to a rooster

src/plugins/chatview/index.js
marcellintacite: I have create the function under utils file of the chatview plugin

jcbrand:
Looking good so far @marcellintacite!

Please also remember to update the DOAP file to mention support for the new XEP
https://github.com/conversejs/converse.js/blob/master/conversejs.doap

manifest.json
```diff
@@ -26,5 +26,11 @@
   "background_color": "#397491",
   "display": "standalone",
   "scope": "/",
-  "theme_color": "#397491"
+  "theme_color": "#397491",
+  "protocol_handlers": [
+    {
+      "protocol": "xmpp",
+      "url": "/dev?jid=%s"
```
jcbrand: This URL needs to be updated, right?

src/plugins/chatview/utils.js
```diff
@@ -71,3 +73,121 @@ export function resetElementHeight (ev) {
...
+export async function routeToQueryAction(event) {
+    const { u } = _converse.env;
+
+    try {
+        const uri = extractXMPPURI(event);
+        if (!uri) return;
+
+        const { jid, queryParams } = parseXMPPURI(uri);
+        if (!u.isValidJID(jid)) {
+            return log.warn(`Invalid JID: "${jid}"`);
```
jcbrand:
```suggestion
            return log.warn(`routeToQueryAction: Invalid JID: "${jid}"`);
```

src/plugins/chatview/utils.js
```diff
@@ -71,3 +73,121 @@ export function resetElementHeight (ev) {
...
+        }
+
+        const action = queryParams.get('action');
+        if (!action) {
+            log.debug(`No action specified, opening chat for "${jid}"`);
```
jcbrand:
```suggestion
            log.debug(`routeToQueryAction: No action specified, opening chat for "${jid}"`);
```

src/plugins/chatview/utils.js
```diff
@@ -71,3 +73,121 @@ export function resetElementHeight (ev) {
...
+            case 'message':
+                await handleMessageAction(jid, queryParams);
+                break;
+
+            case 'add-roster':
+                await handleRosterAction(jid, queryParams);
+                break;
+
+            default:
+                log.warn(`Unsupported XEP-0147 action: "${action}"`);
```
jcbrand:
```suggestion
                log.warn(`routeToQueryAction: Unsupported XEP-0147 action: "${action}"`);
```

src/plugins/chatview/utils.js
```diff
@@ -71,3 +73,121 @@ export function resetElementHeight (ev) {
...
+        await chat.sendMessage({ body });
+    }
+}
+
+/**
+ * Handles the `action=add-roster` case.
+ */
+async function handleRosterAction(jid, params) {
+    await api.waitUntil('connected');
+    await api.waitUntil('rosterContactsFetched');
```
jcbrand: I think you only need the 2nd `waitUntil` because by the time `rosterContactsFetched` has resolved, `connected` will also have resolved. Or is there a specific reason you have both?

src/plugins/chatview/index.js
jcbrand: It would be great if you could also write tests for the work you've done, so that we have more confidence that it's correct and so that future regressions are caught.

To do so, you can create a new file here: https://github.com/conversejs/converse.js/tree/master/src/plugins/chatview/tests

You can call it `route-actions.js`.

I'm not 100% sure what the best approach will be to write the tests, perhaps just calling `routeToQueryActions` manually and then confirming what happens.

If you get stuck or need help, please ask and I'll see if I can help you.

marcellintacite: Thanks, i am going to work on it

manifest.json
marcellintacite: i will update

CHANGES.md
```diff
@@ -16,6 +16,7 @@
 - #3769: Various OMEMO fixes
 - #3791: Fetching pubsub node configuration fails
 - #3792: Node reconfiguration attempt uses incorrect field names
+- Adding support for protocol handler
```
jcbrand:
```suggestion
- Adds support for opening XMPP URIs in Converse and for XEP-0147 query actions
```

conversejs.doap
```diff
@@ -109,6 +109,15 @@
         <xmpp:xep rdf:resource="https://xmpp.org/extensions/xep-0144.html"/>
       </xmpp:SupportedXep>
     </implements>
+    <implements>
+      <xmpp:SupportedXep>
+        <xmpp:xep rdf:resource="https://xmpp.org/extensions/xep-0147.html"/>
+        <xmpp:status>partial</xmpp:status>
+        <xmpp:version>1.2</xmpp:version>
+        <xmpp:since>12.0.0</xmpp:since>
```
jcbrand: This should be since 12.0.1

marcellintacite:
@jcbrand Hello, i would like your review for the test i wrote

manifest.json
```diff
@@ -26,5 +26,11 @@
   "background_color": "#397491",
   "display": "standalone",
   "scope": "/",
-  "theme_color": "#397491"
+  "theme_color": "#397491",
+  "protocol_handlers": [
+    {
+      "protocol": "xmpp",
+      "url": "/?uri=%s"
```
jcbrand:
```suggestion
      "url": "#converse/action?uri=%s"
```

src/plugins/chatview/utils.js
```diff
@@ -71,3 +73,120 @@ export function resetElementHeight (ev) {
...
+ * Extracts and decodes the xmpp: URI from the window location or hash.
+ */
+function extractXMPPURI(event) {
+    let uri = null;
+
+    // Case 1: protocol handler (?uri=...)
+    const searchParams = new URLSearchParams(window.location.search);
+    uri = searchParams.get('uri');
+
+    // Case 2: hash-based (#converse/action?uri=...)
```
jcbrand: I think it  should always be hash based, and we should **not** support your `Case 1` of only the query string `?uri=...`.

The reason for this is because Converse gets embedded into other web pages, where they could have a different use for the `uri` query string parameter.

The usage of hash `#converse/action` mitigates against there being a clash between Converse and other code in the web page which want to use a `uri` query string parameter for some other purpose.

src/plugins/chatview/utils.js
```diff
@@ -71,3 +73,120 @@ export function resetElementHeight (ev) {
...
+}
+
+/**
+ * Extracts and decodes the xmpp: URI from the window location or hash.
+ */
+function extractXMPPURI(event) {
+    let uri = null;
+
+    // Case 1: protocol handler (?uri=...)
+    const searchParams = new URLSearchParams(window.location.search);
```
jcbrand: In this repo, we use `snake_case` to name variables (unless the variable refers to a function) and `camelCase` for functions.

I know this is not how most other JavaScript projects name their variables, but this is an old convention dating back to 2013, when I was influenced by the naming conventions in Python.

To avoid confusion and inconsistency, please stick to the convention of using `snake_case` for variables.

src/plugins/chatview/tests/query-actions.js
```diff
@@ -0,0 +1,206 @@
...
+            pathname: '/',
+        };
+
+        // Spy on history.replaceState to verify URL cleanup
+        const replaceStateSpy = jasmine.createSpy('replaceState');
+        window.history.replaceState = replaceStateSpy;
+
+        try {
+            // Import the function to test
+            const { routeToQueryAction } = await import('../utils.js');
```
jcbrand: Instead of importing `utils.js`, you can also assign `routeToQueryAction` to the `u` object, like it's done here: https://github.com/conversejs/converse.js/blob/8c206adff976e8266c97be405722e4513f366a38/src/plugins/omemo-views/utils.js#L307

Then you can use `u.routeToQueryAction` or `u.chatview.routeToQueryAction`

Not super important though and if the `import` works then I'm ok with it.

src/plugins/chatview/tests/query-actions.js
```diff
@@ -0,0 +1,206 @@
...
+        // Wait for roster to be initialized so we can open chats
+        await mock.waitForRoster(_converse, 'current', 1);
+
+        // Save original globals to restore them later
+        const originalLocation = window.location;
+        const originalReplaceState = window.history.replaceState;
+
+        // Mock window.location to simulate a protocol handler URI
+        // This simulates: ?uri=xmpp:romeo@montague.lit
+        delete window.location;
```
jcbrand: This line causes an error:
https://github.com/conversejs/converse.js/actions/runs/18488527118/job/52676664142?pr=3870#step:4:1972

Try running the tests locally.

You can change `it("opens a chat` to `fit("opens a chat` to only run this particular test.

Then you can run `npm run test` to test only this specific test. When the test browser is open, click the `Debug` button, then you can open the browser developer tools and step through the test line-by-line if need be.

jcbrand (changes requested):

src/plugins/chatview/tests/query-actions.js
marcellintacite: Hello @jcbrand , do you know a proper way of working with window.location in jasmine? i have been trying but when i run test i am getting : 
`Failed:` Cannot redefine property: location.` 
I am blocked with tests, i'll need your help.

marcellintacite:
@jcbrand i need your help on tests so i can get done with this PR. I am facing some errors

marcellintacite:
Hello @jcbrand i have fixed test issue

src/plugins/chatview/tests/query-actions.js
```diff
@@ -0,0 +1,167 @@
...
+            window.location.hash = originalHash;
+            window.history.replaceState = originalReplaceState;
+        }
+    }));
+
+    /**
+     * Test handling of invalid JIDs
+     * This ensures the function gracefully handles malformed JIDs
+     */
+    fit("handles invalid JID gracefully",
```
jcbrand: @marcellintacite Please rename these functions so that they're all `it`

src/plugins/chatview/utils.js
```diff
@@ -71,3 +73,117 @@ export function resetElementHeight (ev) {
...
+ * Supports message sending, roster management, and future actions.
+ *
+ * Example URIs:
+ *   xmpp:user@example.com?action=message&body=Hello
+ *   xmpp:user@example.com?action=add-roster&name=John&group=Friends
+ */
+export async function routeToQueryAction(event) {
+    const { u } = _converse.env;
+
+    try {
```
jcbrand: @marcellintacite  Why is it necessary to put everything in a `try/catch` here?

What happens when it's not there? Generally speaking, a try/catch should only be put around a single line or a small section of code that might be reasonably expected to throw an error that needs to be handled.

marcellintacite:
Hi @jcbrand , could you please provide an update on this Pull Request? I'd like to know what's left to address so we can proceed with merging it.

CHANGES.md
```diff
@@ -18,6 +17,7 @@
 - #3769: Various OMEMO fixes
 - #3791: Fetching pubsub node configuration fails
 - #3792: Node reconfiguration attempt uses incorrect field names
+- Adds support for opening XMPP URIs in Converse and for XEP-0147 query actions
```
jcbrand, line 132: This line is under an already released version, it should go under `## Next release`.

I think you should also rebase this PR onto the latest commit in the `master` branch.

src/plugins/chatview/utils.js
```diff
@@ -71,3 +73,113 @@ export function resetElementHeight (ev) {
...
+
+/**
+ * Extracts and decodes the xmpp: URI from the window location or hash.
+ */
+function extractXMPPURI(event) {
+    let uri = null;
+    // hash-based (#converse/action?uri=...)
+    if (location.hash.startsWith('#converse/action?uri=')) {
+        event?.preventDefault();
+        uri = location.hash.split('uri=').pop();
```
jcbrand: Have you tried splititng with the full string? If that works it will be safer.

```suggestion
        uri = location.hash.split('#converse/action?uri=').pop();
```

src/plugins/chatview/utils.js
```diff
@@ -71,3 +73,113 @@ export function resetElementHeight (ev) {
...
+    }
+
+    if (!uri) return null;
+
+    // Decode URI and remove xmpp: prefix
+    uri = decodeURIComponent(uri);
+    if (uri.startsWith('xmpp:')) uri = uri.slice(5);
```
jcbrand, line 121: When would it *not* start with `xmpp:` and would it then still be legitimate?

src/plugins/chatview/utils.js
```diff
@@ -71,3 +73,113 @@ export function resetElementHeight (ev) {
...
+
+    // Clean up URL (remove ?uri=... for a clean view)
+    const cleanUrl = `${window.location.origin}${window.location.pathname}`;
+    window.history.replaceState({}, document.title, cleanUrl);
```
jcbrand: We use snake_case for variable names.

```suggestion
    const clean_url = `${window.location.origin}${window.location.pathname}`;
    window.history.replaceState({}, document.title, clean_url);
```

src/plugins/chatview/utils.js
```diff
@@ -71,3 +73,113 @@ export function resetElementHeight (ev) {
...
+    }
+
+    switch (action) {
+        case 'message':
+            await handleMessageAction(jid, query_params);
+            break;
+
+        case 'add-roster':
+            await handleRosterAction(jid, query_params);
+            break;
```
jcbrand: Adding users to the roster is not the responsibility of the `chatview` plugin, which is mainly responsible with showing a UI for 1:1 chats.

We have a plugin architecture in Converse.js and we ideally want each plugin to be solely responsible for its domain.

So the `rosterview` plugin should be responsible for checking whether there is a roster-related query action.

And if we later have MUC related query actions, they should go into the `muc_views` plugin, and so on.

The main way in which we enable separation of concerns between the different plugins, is by triggering events and then letting those plugins listen for these events.

So, I think this function needs to be refactored and the roster specific code should be moved into  a function in `rosterview` plugin. You can give it the same name, so there's a `routeToQueryAction` function here which is only responsible for opening chats and sending messages, and another `routeToQueryAction` function in `rosterview/utils.js` which is responsible for handling roster actions.

Then inside the `rosterview` plugin you **also** call `routeToQueryAction`.

Additionally, the other two actions, sending a message and opening a chat apply to both MUCs and 1:1 chats, right? So they should go out of this plugin and into the `chatboxviews` plugin, since that plugin encompasses both MUC and 1:1 and this particular plugin is only concerned with 1:1 chats.

That way, we avoid spaghetti code and plugins that do things outside of their scope.

src/plugins/chatview/index.js
```diff
@@ -65,6 +65,15 @@ converse.plugins.add('converse-chatview', {
...
         Object.assign(_converse.exports, exports);
 
+        if ('registerProtocolHandler' in navigator) {
+            try {
+                const handlerUrl = `${window.location.origin}${window.location.pathname}#converse/action?uri=%s`;
+                navigator.registerProtocolHandler('xmpp', handlerUrl);
+            } catch (error) {
+                console.warn('Failed to register protocol handler:', error);
+            }
+        }
```
jcbrand: This section of code is more low-level and **not** directly related to the `chatview` plugin, so it should go somewhere else.

I think a better place would be to put it inside `src/entry.js`.

src/plugins/chatview/tests/query-actions.js
```diff
@@ -0,0 +1,167 @@
...
+            window.location.hash = originalHash;
+            window.history.replaceState = originalReplaceState;
+        }
+    }));
+
+    /**
+     * Test roster add functionality when action=add-roster
+     * This tests URI parsing and adding a contact to the roster
+     */
+    it("adds a contact to roster when action=add-roster",
```
jcbrand: This test should go into the `rosterview` plugin once you've done this:
https://github.com/conversejs/converse.js/pull/3870/changes#r2667473276

The other tests should go into the `chatboxviews` plugin, as explained in the above-linked comment.

jcbrand (changes requested):
Hi @marcellintacite 

I took some time to review this PR. Thanks a lot for the work you've done on this so far.

Please have a look at this comment first:
https://github.com/conversejs/converse.js/pull/3870/changes#r2667473276

Since it's the most important one of my review.

It's important that we stick to the plugin architecture and separation of concerns, which is why I think you need to move the functionality to separate plugins.

The roster-related stuff should go into the `rosterview` plugin and the other actions can go into `chatboxviews` (which encompasses both MUC and 1:1 whereas `chatview` is only 1:1).

Let me know if you have any questions or if anything is unclear.

marcellintacite:
Hi @jcbrand , i have updated the PR can you check please

jcbrand:
@marcellintacite Can you please squash and rebase onto the master branch, then I'll review? Thanks!

marcellintacite:
Hi @jcbrand , I have squashed and rebased my commits onto the master branch. I will now await your review.

src/plugins/rosterview/utils.js
```diff
@@ -488,3 +488,54 @@ Object.assign(u, {
...
+async function handleRosterAction(jid, params) {
+    const name = params.get('name') || jid.split('@')[0];
+    const group = params.get('group');
+    const groups = group ? [group] : [];
+
+    try {
+        // Check if user is connected before attempting to add contact
+        if (!api.connection.connected()) {
+            api.alert('warning', __('Not connected'), 
+                [__('You must be connected to add contacts to your roster.')]);
```
jcbrand:
```suggestion
                [__('Please login first before adding contacts to your roster.')]);
```

src/plugins/chatview/utils.js
```diff
@@ -71,3 +73,71 @@ export function resetElementHeight (ev) {
...
+ *
+ * @param {Event} [event] - Optional event object (used when called as protocol handler)
+ * @returns {Promise<void>}
+ * @example
+ * // Automatically called on initialization in src/index.js
+ * // Can be called manually after location.hash changes:
+ * window.location.hash = '#converse/action?uri=xmpp%3Auser%40example.com';
+ * await routeToQueryAction();
+ */
+export async function routeToQueryAction(event) {
```
jcbrand, line 91: I don't see where this function is ever called.

In the tests you call it manually, which is not useful. This function should be automatically called when the URI changes.

I'm also not convinced that this function belongs in the chatview plugin, since it provides more general functionality relied upon by other plugins.

jcbrand:
Thanks @marcellintacite, please see my review comments.

jcbrand:
@marcellintacite 

CI is failing

```
> tsc -p ./src/log/tsconfig.json && tsc -p ./src/headless/tsconfig.json && tsc

Error: src/plugins/rosterview/index.js(7,10): error TS2300: Duplicate identifier 'clearXMPPProvidersCache'.
Error: src/plugins/rosterview/index.js(7,35): error TS2300: Duplicate identifier 'highlightRosterItem'.
Error: src/plugins/rosterview/index.js(9,10): error TS2300: Duplicate identifier 'clearXMPPProvidersCache'.
Error: src/plugins/rosterview/index.js(9,35): error TS2300: Duplicate identifier 'highlightRosterItem'.
make: *** [Makefile:248: check] Error 2
/usr/bin/bash /home/runner/work/_actions/GabrielBB/xvfb-action/v1/cleanup.sh
No xvfb processes to kill
Error: The process '/usr/bin/xvfb-run' failed with exit code 2

```

marcellintacite:
> @marcellintacite
> 
> CI is failing
> 
> ```
> > tsc -p ./src/log/tsconfig.json && tsc -p ./src/headless/tsconfig.json && tsc
> 
> Error: src/plugins/rosterview/index.js(7,10): error TS2300: Duplicate identifier 'clearXMPPProvidersCache'.
> Error: src/plugins/rosterview/index.js(7,35): error TS2300: Duplicate identifier 'highlightRosterItem'.
> Error: src/plugins/rosterview/index.js(9,10): error TS2300: Duplicate identifier 'clearXMPPProvidersCache'.
> Error: src/plugins/rosterview/index.js(9,35): error TS2300: Duplicate identifier 'highlightRosterItem'.
> make: *** [Makefile:248: check] Error 2
> /usr/bin/bash /home/runner/work/_actions/GabrielBB/xvfb-action/v1/cleanup.sh
> No xvfb processes to kill
> Error: The process '/usr/bin/xvfb-run' failed with exit code 2
> ```

Thanks, i am fixing it

marcellintacite:
Hi @jcbrand ,

I've just pushed a major update to this PR to bring the implementation into strict compliance with the [XEP-0147 ](https://xmpp.org/extensions/xep-0147.html) specification.

The previous approach to parsing URIs and routing actions was not fully conforming to the standard. Specifically, the old parser was treating the action as a standard key-value query parameter (e.g., ?action=message&body=Text), whereas XEP-0147 explicitly dictates a semicolon-separated format where the action is the leading key (e.g., ?message;body=Text).

Because of this discrepancy, several core actions were failing silently or behaving incorrectly:
- Action Routing Fixed: Rewrote the URI parser and the xmppURIAction event payload to correctly isolate the leading action type from the subsequent query_params
- Draft Message Pre-filling: Because the old routing was broken, ?message links were opening empty chat windows 
- The newly compliant router now correctly extracts the body parameter and injects it as a draft immediately when the chat box is spawned
- Race Conditions Resolved: Added await api.waitUntil('chatBoxesFetched') across all routers to ensure Converse waits for authentication and state hydration before executing URI actions, preventing startup crashes when clicking external links
- Tests Updated: Completely rewrote the mocked URIs in our test suites to strictly enforce the XEP-0147 semicolon syntax, ensuring our automated tests accurately reflect the specification moving forward.

_Note regarding CI Failures: The TypeScript errors currently failing the CI pipeline (e.g., Property 'resolve' does not exist on type 'Promise<void>' and Type 'typeof PersistentStorage' is missing the following properties...) are originating from upstream changes in the @converse/skeletor 3.1.0 upgrade that this branch recently inherited from master. These errors are entirely unrelated to the XEP-0147 implementation in this PR and will require a separate fix on master to resolve the type discrepancies._

Please kindly review the actual implementation

jcbrand:
Thanks for this, @marcellintacite, and I apologise for the timing.

XEP-0147 URI handling landed on master in ae93376ba on 5 August, covering the same six query actions you did in this PR.

What you did in this PR that wasn't yet in master is registering Converse as the browser's handler so that links clicked
outside Converse open inside it. I refactored out that code in a new commit with you as co-author, and it closes #1247.

Issue #3916, opened 2026-01-22
[feature request] XEP-0461: Message Replies


BlueNebulaDev:
It would be awesome if converse.js could add support for message replies (XEP-0461).

If someone replies to a message from another client, converse.js only shows the replied message as a quote. Nothing happens if you click on the quote (while other clients highlight the replied message). The user who sent the replied message is not shown.

And there's no button to reply to a message.

It would be really nice if this XMPP feature could get supported.

Somebody even asked about this feature on Stack Overflow 5 years ago: https://stackoverflow.com/questions/65956139/how-can-i-add-reply-message-in-conversejs

aliraza556:
@jcbrand @BlueNebulaDev Please assign this to me. Also, if you don’t mind, could you mention the bounty amount?

jcbrand:
@aliraza556 Please read the rules regarding bounties here: https://github.com/conversejs/converse.js/issues/2481

aliraza556:
> [@aliraza556](https://github.com/aliraza556) Please read the rules regarding bounties here: [#2481](https://github.com/conversejs/converse.js/issues/2481)

@jcbrand Cool Please assign me

subhu339:
Hi @jcbrand, I’ve started working on the reply preview UI and message highlight behavior as groundwork for XEP-0461.

PR: #3918 – Add reply preview UI and highlight animation
This adds a clickable reply preview block and visual highlight for replied messages.

jcbrand:
@aliraza556 If you read the rules, you'd have seen we don't assign tickets.

jcbrand:
Thanks @subhu339, I'll take a look soon

aliraza556:
Hi @jcbrand,
I've opened PR #3919 implementing XEP-0461 Message Replies. It adds the ability to reply to specific messages in both 1:1 chats and MUCs, with a reply preview panel and clickable context on replied messages.
Tests are included. Would appreciate your review when you have time.
Thanks!

sreeshanth-soma:
Hello @jcbrand ,
I’ve completed the implementation and opened a PR for this. Whenever you have time, I’d appreciate you taking a look. Thanks!
PR #3921

jcbrand:
@sreeshanth-soma: Thank you for you making a pull request.

However, did you notice that @aliraza556 has already made a pull request 3 days earlier and that it's close to being merged?

Only one implementation, the one that gets merged, ultimately gets the bounty.

[...]

PR #3918, opened 2026-01-23
Add reply preview UI and highlight animation

subhu339:
#3916

This PR adds UI support for displaying reply previews in chat messages.
When a message is a reply, a small preview box is shown above the message body, containing the original author name and a short text snippet. Clicking the preview scrolls to and highlights the referenced message for easier context.

This change is UI-only and backward compatible. It does not modify any XMPP protocol logic or server behavior.

### DIFF
--- a/src/shared/chat/message.js
+++ b/src/shared/chat/message.js
@@ -14,7 +14,8 @@ import tplMessageText from './templates/message-text.js';
 import tplRetraction from './templates/retraction.js';
 import tplSpinner from 'templates/spinner.js';
 import { ObservableElement } from 'shared/components/observable.js';
-import { __ } from 'i18n';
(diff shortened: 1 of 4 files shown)


src/shared/chat/message.js
```diff
@@ -147,6 +162,27 @@ export default class Message extends ObservableElement {
         return this.model.occupant?.get('role');
     }
 
+    hasReply () {
+    return !!this.model.get('reply_to');
```
jcbrand, line 166: Thanks for your PR @subhu339!

However, it doesn't look finished or even usable in its current state.

For example, nowhere in your PR is the `reply_to` set, but you read it here as if it was set somewhere.

There is a competing PR from @aliraza556 that is almost finished, so I think we'll choose that one instead.

PR #3922, opened 2026-01-26
feat: Add XEP-0231 Bits of Binary support

### LINKED ISSUE #3611: XEP-0231: Bits of Binary - for custom smileys and CAPTCHA

XEP-0231 is used to transfer small bits of binary, hence the name between clients. In Pidgin it is used to transfer custom smileys. Those are limited to 8KiB as per the XEP.

Also some MUCs sent CAPTCHA pictures via this XEP. [Gajim does support XEP-0231 for CAPTCHAs](https://xmpp.org/software/gajim/). Issue #1092 also mentions XEP-0231 in its XEP spec.

**Describe the solution you'd like**
Implement at least receiving custom pictures, if possible also sending custom ones. 

**Describe alternatives you've considered**
Maybe Stickers, but unsure if those are always a sperate message. BOB pictures can usually be inlined.

**Additional context**
https://xmpp.org/extensions/xep-0231.html

Example picture from Pidgin: 

Image

sreeshanth-soma:
## Summary
Implements [XEP-0231: Bits of Binary](https://xmpp.org/extensions/xep-0231.html), enabling Converse.js to receive and display inline binary data such as custom smileys from clients like Pidgin.
Closes #3611
### New `converse-bob` Plugin
- **In-memory cache** with TTL and max-age expiration
- **API**: `api.bob.get()`, `api.bob.store()`, `api.bob.has()`  
- **IQ handling**: Fetches uncached BOB data via `<iq type="get">`
- **Message parsing**: Extracts `<data xmlns="urn:xmpp:bob">` elements
- **Disco feature**: Registers `urn:xmpp:bob`
- **Security**: Image MIME types only, 8KB size limit per spec
## Test Results
TOTAL: 94 SUCCESS

### New Tests (7)
- Cache storage, retrieval, expiration
- Size and MIME validation
- Message parsing and IQ requests
- Error handling
## Capability Hash
Adding `urn:xmpp:bob` updates XEP-0115 verification string from `1T0pIfIxYO645OaT9gpXVXOvb9s=` to `ZXDrJD54GWZYlZif9IuMxum/u+g=`

### DIFF
--- a/docs/source/configuration.rst
+++ b/docs/source/configuration.rst
@@ -1144,6 +1144,18 @@ Set this option to ``true`` to request all pages of archived messages, but be
 aware that this can have performance implications.
 
 
+max_bob_size
+------------
+
+* Default: ``8192``
+
+The maximum size (in bytes) of Base64-encoded data that may be stored via
+`XEP-0231: Bits of Binary <https://xmpp.org/extensions/xep-0231.html>`_.
+
+Any received BOB data that exceeds this limit will be discarded.
+This helps prevent excessive memory usage from large inline binary objects.
+
+
 muc_hats
 --------
 
--- a/src/headless/plugins/bob/bob.js
+++ b/src/headless/plugins/bob/bob.js
@@ -1,0 +1,47 @@
+import { Model } from '@converse/skeletor';
+import log from '@converse/log';
+
+/**
+ * @class BOB
+ * Represents a single Bits of Binary (XEP-0231) cache entry
+ */
+class BOB extends Model {
+    get idAttribute() {
+        return 'cid';
+    }
+
+    /**
+     * Check if this BOB entry has expired based on max_age
+     * @returns {boolean}
+     */
+    isExpired() {
+        const max_age = this.get('max_age');
+        if (!max_age) return max_age === 0 ? true : false;
+
+        const timestamp = this.get('timestamp');
+        const age = (Date.now() - timestamp) / 1000;
+        return age > max_age;
+    }
+
+    /**
+     * Get the BOB data as a Blob URL
+     * @returns {string|null}
+     */
+    getBlobURL() {
+        const data = this.get('data');
+        const type = this.get('type');
+
+        if (!data || !type) return null;
+
+        try {
+            const bytes = Uint8Array.from(atob(data), (c) => c.charCodeAt(0));
+            const blob = new Blob([bytes], { type });
+            return URL.createObjectURL(blob);
+        } catch (e) {
+            log.error(`Failed to create Blob URL for BOB ${this.get('cid')}:`, e);
+            return null;
+        }
+    }
+}
+
+export default BOB;
--- a/src/headless/plugins/bob/bobs.js
+++ b/src/headless/plugins/bob/bobs.js
@@ -1,0 +1,56 @@
+import { Collection } from '@converse/skeletor';
+import { getOpenPromise } from '@converse/openpromise';
+import api from '../../shared/api/index.js';
+import _converse from '../../shared/_converse.js';
+import { initStorage } from '../../utils/storage.js';
+import BOB from './bob.js';
+
+/**
+ * @class BOBs
+ * Collection of BOB (Bits of Binary) cache entries with persistent storage
+ * @extends {Collection<BOB>}
+ */
+class BOBs extends Collection {
+    constructor() {
+        super();
+        this.model = BOB;
+    }
+
+    async initialize() {
+        const { session } = _converse;
+        const bare_jid = session.get('bare_jid');
+        const cache_key = `${bare_jid}-converse.bobs`;
+        initStorage(this, cache_key);
+
+        await this.fetchBOBs();
+
+        this.cleanupExpired();
+
+        /**
+         * Triggered as soon as the `_converse.state.bobs` collection has
+         * been initialized and populated from cache.
+         * @event _converse#BOBsInitialized
+         */
+        api.trigger('BOBsInitialized');
+    }
+
+    fetchBOBs() {
+        const deferred = getOpenPromise();
+        this.fetch({
+            success: () => deferred.resolve(),
+            error: () => deferred.resolve(),
+        });
+        return deferred;
+    }
+
+    /**
+     * Remove expired BOB entries from the collection
+     */
+    cleanupExpired() {
+        this.forEach((bob) => {
+            if (bob.isExpired()) bob.destroy();
+        });
+    }
+}
+
+export default BOBs;
--- a/src/headless/plugins/bob/plugin.js
+++ b/src/headless/plugins/bob/plugin.js
@@ -1,0 +1,161 @@
+/**
+ * @module converse-bob
+ * @description
+ * XEP-0231: Bits of Binary
+ * Handles receiving and caching small binary data (custom smileys, CAPTCHAs)
+ */
+import sizzle from 'sizzle';
+import converse from '../../shared/api/public.js';
+import _converse from '../../shared/_converse.js';
+import log from '@converse/log';
+import BOB from './bob.js';
+import BOBs from './bobs.js';
+import bob_api from './api.js';
+
+const { Strophe } = converse.env;
+
+Strophe.addNamespace('BOB', 'urn:xmpp:bob');
+
+converse.plugins.add('converse-bob', {
+    dependencies: [],
+
+    initialize() {
+        const { api } = this._converse;
+
+        api.promises.add('BOBsInitialized');
+
+        api.settings.extend({
+            max_bob_size: 8192,
+        });
+
+        const exports = { BOB, BOBs };
+        Object.assign(_converse.exports, exports);
+
+        api.listen.on('addClientFeatures', () => {
+            api.disco.own.features.add(Strophe.NS.BOB);
+            api.disco.own.features.add('http://jabber.org/protocol/xhtml-im');
+        });
+
+        api.listen.on('connected', async () => {
+            const bobs = new _converse.exports.BOBs();
+            _converse.state.bobs = bobs;
+            await bobs.initialize();
+
+            _converse.state.bobs_cleanup_interval = setInterval(() => {
+                if (_converse.state.bobs) {
+                    _converse.state.bobs.cleanupExpired();
+                }
+            }, 600000);
+        });
+
+        api.listen.on('clearSession', () => {
+            if (_converse.state.bobs_cleanup_interval) {
+                clearInterval(_converse.state.bobs_cleanup_interval);
+                delete _converse.state.bobs_cleanup_interval;
+            }
+            if (_converse.state.bobs) {
+                _converse.state.bobs.clearStore();
+                delete _converse.state.bobs;
+            }
+        });
+
+        Object.assign(api, bob_api);
+
+        api.listen.on('parseMessage', (stanza, attrs) => {
+            const bob_data = [];
+            const data_els = sizzle(`data[xmlns="${Strophe.NS.BOB}"]`, stanza);
+
+            data_els.forEach((el) => {
+                const cid = el.getAttribute('cid');
+                const type = el.getAttribute('type');
+                const max_age = parseInt(el.getAttribute('max-age') || '86400', 10);
+                const data = el.textContent.trim();
+
+                if (cid && data) {
+                    bob_data.push({ cid, data, type, max_age });
+                    api.bob?.store(cid, data, type, max_age);
+                }
+            });
+
+            if (bob_data.length > 0) {
+                attrs.bob_data = bob_data;
+            }
+
+            // Extract cid: URIs from XHTML-IM <img> tags (e.g. Pidgin custom smileys).
+            // When a client sends a custom smiley, the plain <body> only contains
+            // fallback text. The cid: URI is only in the XHTML-IM <img src="cid:..."/>.
+            //
+            // Converse extracts attrs.body from the XHTML body's textContent, which
+            // drops <img> elements entirely. So we walk the XHTML body DOM to
+            // reconstruct the text with cid: URIs in place of <img> elements.
+            const xhtml_body = stanza.getElementsByTagNameNS('http://www.w3.org/1999/xhtml', 'body')[0];
+            if (xhtml_body) {
+                const cid_imgs = Array.from(
+                    xhtml_body.getElementsByTagNameNS('http://www.w3.org/1999/xhtml', 'img')
+                ).filter((img) => img.getAttribute('src')?.startsWith('cid:'));
+
+                if (cid_imgs.length > 0) {
+                    // Walk all nodes in the XHTML body and build text, replacing
+                    // <img src="cid:..."> with the cid: URI string.
+                    const walker = xhtml_body.ownerDocument.createTreeWalker(
+                        xhtml_body,
+                        4 | 1, // NodeFilter.SHOW_TEXT | NodeFilter.SHOW_ELEMENT
+                        null
+                    );
+                    const parts = [];
+                    let node;
+                    while ((node = walker.nextNode())) {
+                        if (node.nodeType === 3) {
+                            // Text node
+                            parts.push(node.textContent);
+                        } else if (
+                            node.nodeType === 1 &&
+                            node.localName === 'img' &&
+                            node.getAttribute('src')?.startsWith('cid:')
+                        ) {
+                            parts.push(node.getAttribute('src'));
+                        }
+                    }
+                    const body = parts.join('').trim();
+                    if (body) {
+                        attrs.body = body;
+                        attrs.message = body;
+                    }
+
+                    // Fetch BOB data via IQ for any CIDs not included inline
+                    // (e.g. Pidgin sends <img src="cid:..."> without inline <data>)
+                    const from_jid = attrs.from || stanza.getAttribute('from');
+                    const bare_from = Strophe.getBareJidFromJid(from_jid);
+                    const stored_cids = new Set((attrs.bob_data || []).map((b) => b.cid));
+                    for (const img of cid_imgs) {
+                        const src = img.getAttribute('src');
+                        const cid = src.slice(4); // strip "cid:" prefix
+                        if (!stored_cids.has(cid)) {
+                            log.info(`BOB: Fetching missing CID via IQ: ${cid}`);
+                            api.bob
+                                .fetch(cid, from_jid)
+                                .then(() => {
+                                    // Re-render messages containing this CID
+                                    const chatbox = _converse.chatboxes.get(bare_from);
+                                    if (chatbox) {
+                                        chatbox.messages.forEach((msg) => {
+                                            const body = msg.get('body') || '';
+                                            if (body.includes(cid)) {
+                                                // Force a real model change to trigger view re-render
+                                                msg.save('bob_updated', Date.now());
+                                            }
+                                        });
+                                    }
+                                })
+                                .catch((e) => log.error(`BOB: IQ fetch failed for ${cid}:`, e));
+                        }
+                    }
+                }
+            }
+
+            return attrs;
+        });
+
+        log.info('XEP-0231: Bits of Binary plugin initialized');
+    },
+});
--- a/src/headless/plugins/bob/tests/bob.js
+++ b/src/headless/plugins/bob/tests/bob.js
@@ -1,0 +1,188 @@
+/* global converse */
+import mock from '../../../tests/mock.js';
+const { u, Strophe, stx, sizzle } = converse.env;
+
+describe('XEP-0231: Bits of Binary', function () {
+    describe('BOB Cache', function () {
+        it(
+            'stores and retrieves BOB data',
+            mock.initConverse(['chatBoxesInitialized', 'BOBsInitialized'], {}, async (_converse) => {
+                const { api } = _converse;
+                const cid = 'cid:sha1+8f35fef110ffc5df08d579a50083ff9308fb6242@bob.xmpp.org';
+                const data = 'iVBORw0KGgoAAAANSUhEUg==';
+                const type = 'image/png';
+
+                // Store data
+                await api.bob.store(cid, data, type, 86400);
+
+                // Verify it's cached
+                expect(await api.bob.has(cid)).toBe(true);
+
+                // Retrieve as Blob URL
+                const blobUrl = await api.bob.get(cid);
+                expect(blobUrl).toBeTruthy();
+                expect(blobUrl.startsWith('blob:')).toBe(true);
+            }),
+        );
+
+        it(
+            'respects max-age expiration',
+            mock.initConverse(['chatBoxesInitialized', 'BOBsInitialized'], {}, async (_converse) => {
+                const { api } = _converse;
+                const cid = 'cid:sha1+test@bob.xmpp.org';
+                const data = 'iVBORw0KGgoAAAANSUhEUg==';
+
+                // Store with 1 second max-age
+                await api.bob.store(cid, data, 'image/png', 1);
+                expect(await api.bob.has(cid)).toBe(true);
+
+                // Wait 2 seconds
+                await new Promise((resolve) => setTimeout(resolve, 2000));
+
+                // Should be expired
+                expect(await api.bob.has(cid)).toBe(false);
+            }),
+        );
+
+        it(
+            'rejects oversized data',
+            mock.initConverse(['chatBoxesInitialized', 'BOBsInitialized'], {}, async (_converse) => {
+                const { api } = _converse;
+                const cid = 'cid:sha1+large@bob.xmpp.org';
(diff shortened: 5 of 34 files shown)


sreeshanth-soma:
Hello @jcbrand . I've addressed your feedback. The non-headless tests that were failing due to the urn:xmpp:bob feature have been fixed:

- Updated all capability hashes in the non-headless test files
- Added TypeScript type definitions for the BOB plugin
All tests are now passing locally (matching upstream/master behavior). Ready for re-review!

src/headless/plugins/bob/plugin.js
```diff
@@ -0,0 +1,199 @@
...
+import _converse from '../../shared/_converse.js';
+import log from '@converse/log';
+
+const { Strophe, $iq } = converse.env;
+
+// Maximum size for BOB data (8KB as per XEP-0231)
+const MAX_BOB_SIZE = 8192;
+
+// In-memory cache for BOB data
+const bob_cache = new Map();
```
jcbrand: I'm not convinced that an in-memory cache is the right way to go here.

Every time you reload the tab the cache will be wiped and then you lose all the images.
If you then try to refetch the images, you might not get all of them due to the sender being offline.

So instead, we should probably create a persistent collection of BOB images, similar to what we do with VCards: https://github.com/conversejs/converse.js/blob/master/src/headless/plugins/vcard/vcards.js

src/shared/texture/texture.js
```diff
@@ -187,6 +189,46 @@ export class Texture extends String {
         }
     }
 
+    /**
+     * Look for `cid:` URIs (XEP-0231 Bits of Binary) and render as inline images
+     * @param {String} text
+     * @param {number} offset - The index of the passed in text relative to
+     *  the start of the message body text.
+     */
+    async addBOBImages(text, offset) {
```
jcbrand: This function should go into a `bob-views` plugin and then in that plugin you should listen for the `afterMessageBodyTransformed` event in order to call this function.

See here: https://github.com/conversejs/converse.js/blob/0ea38d74be1fcc18616051bde0d361631dd608b2/src/plugins/omemo-views/index.js#L34

src/headless/plugins/bob/plugin.js
sreeshanth-soma: Thanks for the feedback! I've replaced the in-memory Map with a persistent Skeletor Collection, following the VCards pattern:

- Created bob.js model with isExpired() and getBlobURL() methods.
- Created bobs.js collection using initStorage() for IndexedDB persistence
- Collection is initialized on connected event and cleaned up on clearSession

BOB images now persist across page reloads and remain available even if the sender goes offline.

src/shared/texture/texture.js
sreeshanth-soma: Done! Created a new bob-views plugin that:

- Listens to afterMessageBodyTransformed event (following the omemo-views pattern)
- Moved the BOB image handling logic from texture.js to the new plugin
- Registered it in src/index.js and added to VIEW_PLUGINS whitelist

This properly separates the headless BOB storage from the view-layer rendering.

src/headless/plugins/bob/bob.js
```diff
@@ -0,0 +1,50 @@
...
+        const type = this.get("type");
+
+        if (!data || !type) return null;
+
+        try {
+            const binary = atob(data);
+            const bytes = new Uint8Array(binary.length);
+            for (let i = 0; i < binary.length; i++) {
+                bytes[i] = binary.charCodeAt(i);
+            }
```
jcbrand:
```suggestion
            const bytes = Uint8Array.from(atob(data), c => c.charCodeAt(0));                                                                
```

src/headless/plugins/bob/bobs.js
```diff
@@ -0,0 +1,56 @@
...
+    async initialize() {
+        const { session } = _converse;
+        const bare_jid = session.get("bare_jid");
+        const cache_key = `${bare_jid}-converse.bobs`;
+        initStorage(this, cache_key);
+
+        await this.fetchBOBs();
+
+        // Clean up expired entries
+        this.cleanupExpired();
```
jcbrand, line 27: Please remove comments like this. It's already clear from the code what's happening.

src/headless/plugins/bob/bobs.js
```diff
@@ -0,0 +1,56 @@
...
+        });
+        return deferred;
+    }
+
+    /**
+     * Remove expired BOB entries from the collection
+     */
+    cleanupExpired() {
+        const expired = this.filter((bob) => bob.isExpired());
+        expired.forEach((bob) => bob.destroy());
```
jcbrand:
```suggestion
        this.forEach((bob) => if (bob.isExpired()) bob.destroy());
```

src/headless/plugins/bob/plugin.js
```diff
@@ -0,0 +1,199 @@
...
+ * @description
+ * XEP-0231: Bits of Binary
+ * Handles receiving and caching small binary data (custom smileys, CAPTCHAs)
+ */
+import sizzle from 'sizzle';
+import converse from '../../shared/api/public.js';
+import _converse from '../../shared/_converse.js';
+import log from '@converse/log';
+
+const { Strophe, $iq } = converse.env;
```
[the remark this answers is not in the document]
jcbrand: Please pay attention to these security bot messages.

src/headless/plugins/bob/plugin.js
```diff
@@ -0,0 +1,212 @@
...
+import converse from '../../shared/api/public.js';
+import _converse from '../../shared/_converse.js';
+import log from '@converse/log';
+import BOB from './bob.js';
+import BOBs from './bobs.js';
+
+const { Strophe, $iq } = converse.env;
+
+// Maximum size for BOB data (8KB as per XEP-0231)
+const MAX_BOB_SIZE = 8192;
```
jcbrand: I think this can be made configurable.

Where you init the plugin:

```
        // Configure plugin settings
        api.settings.extend({
            max_bob_size: 8192
        });
```

src/headless/plugins/bob/plugin.js
```diff
@@ -0,0 +1,212 @@
...
+    dependencies: [],
+
+    initialize() {
+        const { api } = this._converse;
+
+        api.promises.add('BOBsInitialized');
+
+        // Export classes
+        const exports = { BOB, BOBs };
+        Object.assign(_converse, exports); // XXX DEPRECATED
```
jcbrand: Please don't add deprecated lines to new files.

src/headless/plugins/bob/plugin.js
```diff
@@ -0,0 +1,212 @@
...
+        Object.assign(_converse.exports, exports);
+
+        // Register namespace
```
jcbrand: Unnecessary comment

src/shared/chat/message-body.js
```diff
@@ -51,6 +51,7 @@ export default class MessageBody extends CustomElement {
             media_urls: this.model.get('media_urls'),
             mentions: this.model.get('references'),
             nick: this.model.chatbox.get('nick'),
+            from_jid: this.model.get('from'),
```
jcbrand: This line is no longer necessary

src/shared/texture/texture.js
```diff
@@ -56,6 +57,7 @@ export class Texture extends String {
      *  which create new Texture instances (as happens with XEP-393 styling directives).
      * @param {Object} [options]
      * @param {string} [options.nick] - The current user's nickname (only relevant if the message is in a XEP-0045 MUC)
+     * @param {string} [options.from_jid] - The JID of the message sender (needed for fetching BOB images)
```
jcbrand: The changes in this file are no longer necessary. Please look through your changes yourself before asking me to review. You could easily have seen this.

src/headless/plugins/bob/plugin.js
```diff
@@ -0,0 +1,212 @@
...
+                 */
+                has(cid) {
+                    const bobs = _converse.state.bobs;
+                    if (!bobs) return false;
+
+                    const bob = bobs.get(cid);
+                    if (!bob) return false;
+
+                    if (bob.isExpired()) {
+                        bob.destroy();
```
jcbrand: I don't think calling `has` should have the side effect of deleting expired blobs.

If should return `true`, even if the blob is expired.

Instead you can use `setInterval` to call a function every 10 minutes, to check for expired blobs and then delete them.

sreeshanth-soma:
@jcbrand  sorry for the ping can you please review?

src/headless/plugins/bob/bob.js
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    /**
+     * Check if this BOB entry has expired based on max_age
+     * @returns {boolean}
+     */
+    isExpired() {
+        const max_age = this.get("max_age");
+        if (!max_age) return false;
+        if (max_age === 0) return true;
```
jcbrand: There's a bug here. If `max_age === 0`, then `false` will always be returned on line 18 and line 19 will never execute.


```suggestion     
        if (!max_age) return max_age === 0 ? true : false;
```

src/headless/plugins/bob/bob.js
```diff
@@ -0,0 +1,46 @@
...
+        const type = this.get("type");
+
+        if (!data || !type) return null;
+
+        try {
+            const bytes = Uint8Array.from(atob(data), c => c.charCodeAt(0));
+            const blob = new Blob([bytes], { type });
+            return URL.createObjectURL(blob);
+        } catch (e) {
+            return null;
```
jcbrand, line 42: Log an error here so that we're informed when this fails.

src/headless/plugins/bob/plugin.js
```diff
@@ -0,0 +1,214 @@
...
+
+        api.settings.extend({
+            max_bob_size: 8192
+        });
+
+        const exports = { BOB, BOBs };
+        Object.assign(_converse.exports, exports);
+
+        api.listen.on('addClientFeatures', () => {
+            api.disco.own.features.add('urn:xmpp:bob');
```
jcbrand: Please register a new Strophe namespace `Strophe.NS.BOB` and use that instead of the string.

In the tests you can still use the string, but in the normal code we should use `Strophe.NS.BOB`.

See here for an example: https://github.com/conversejs/converse.js/blob/605b1e747c51334357edd80aa25052f8f22f8d4b/src/headless/plugins/omemo/index.js#L10

src/headless/plugins/bob/plugin.js
```diff
@@ -0,0 +1,214 @@
...
+converse.plugins.add('converse-bob', {
+    dependencies: [],
+
+    initialize() {
+        const { api } = this._converse;
+
+        api.promises.add('BOBsInitialized');
+
+        api.settings.extend({
+            max_bob_size: 8192
```
jcbrand: When adding a new setting, you need to document it in `docs/source/configuration.rst`

Please make sure it's in the proper alphabetical position.

src/headless/plugins/bob/plugin.js
```diff
@@ -0,0 +1,214 @@
...
+                const bobs = new _converse.exports.BOBs();
+                _converse.state.bobs = bobs;
+                await bobs.initialize();
+
+                _converse.state.bobs_cleanup_interval = setInterval(() => {
+                    if (_converse.state.bobs) {
+                        _converse.state.bobs.cleanupExpired();
+                    }
+                }, 600000);
+            } catch (e) {
```
jcbrand: What is the purpose of this `try/catch`? What potentially happens when it's not there?

Large `try/catch` sections are often a code-smell and as a general rule a `try/catch` (if used at all) should only wrap a small section of code that is known to throw errors under certain expected circumstances.

src/headless/plugins/bob/plugin.js
jcbrand: @sreeshanth-soma You marked this comment as resolved but I still see the line `bob.destroy()` which is a side-effect that I don't think should be in this function.

src/headless/plugins/bob/plugin.js
```diff
@@ -0,0 +1,214 @@
...
+
+                /**
+                 * Fetch BOB data via IQ-get
+                 * @param {string} cid
+                 * @param {string} from_jid
+                 * @returns {Promise<void>}
+                 */
+                async fetch(cid, from_jid) {
+                    const iq = $iq({
+                        type: 'get',
```
jcbrand: Please use the `stx` tagged template literal instead of `$iq` (or `$msg` or `$pres`).

There are numerous examples throughout the codebase.

src/headless/plugins/bob/plugin.js
```diff
@@ -0,0 +1,214 @@
...
+                 * Store BOB data in persistent cache
+                 * @param {string} cid
+                 * @param {string} data - Base64 encoded data
+                 * @param {string} type - MIME type
+                 * @param {number} max_age - Max age in seconds
+                 */
+                store(cid, data, type, max_age) {
+                    const bobs = _converse.state.bobs;
+                    if (!bobs) {
+                        log.warn('BOB collection not initialized');
```
jcbrand: You can make this function `async` and wait for the `BOBsInitialized` promise. Then you won't have race conditions where a blob gets stored before the collection has been initialized.

src/headless/plugins/bob/plugin.js
```diff
@@ -0,0 +1,214 @@
...
+                clearInterval(_converse.state.bobs_cleanup_interval);
+                delete _converse.state.bobs_cleanup_interval;
+            }
+            if (_converse.state.bobs) {
+                _converse.state.bobs.clearStore();
+                delete _converse.state.bobs;
+            }
+        });
+
+        Object.assign(api, {
```
jcbrand: The API should go into its own file please.

src/headless/plugins/bob/plugin.js
```diff
@@ -0,0 +1,214 @@
...
+                        return;
+                    }
+
+                    try {
+                        const size = atob(data).length;
+                        if (size > api.settings.get('max_bob_size')) {
+                            log.warn(`BOB data for ${cid} exceeds max size (${size} > ${api.settings.get('max_bob_size')})`);
+                            return;
+                        }
+                    } catch (e) {
```
jcbrand: The `try/catch` should not also encapsulate the `if` statement, it should only be around the section of code it's relevant to, which is the `atob` call.

src/headless/plugins/bob/plugin.js
```diff
@@ -0,0 +1,214 @@
...
+                        log.warn(`Invalid base64 data for BOB ${cid}`);
+                        return;
+                    }
+
+                    if (!type.startsWith('image/')) {
+                        log.warn(`BOB data for ${cid} has unsupported MIME type: ${type}`);
+                        return;
+                    }
+
+                    if (bobs.get(cid)) {
```
jcbrand: This should go to the top of the function to avoid unnecessary computation only to exit here.

src/headless/plugins/bob/plugin.js
```diff
@@ -0,0 +1,214 @@
...
+                        return;
+                    }
+
+                    bobs.create({
+                        cid,
+                        data,
+                        type,
+                        max_age: max_age || 86400,
```
jcbrand: In the function signature you can give `max_age` a default value of `86400` then you don't have to do the `||`  here.

src/headless/plugins/bob/plugin.js
```diff
@@ -0,0 +1,214 @@
...
+                },
+
+                /**
+                 * Get BOB data as Blob URL
+                 * @param {string} cid
+                 * @param {string} [from_jid] - JID to request from if not cached
+                 * @returns {Promise<string|null>} - Blob URL or null
+                 */
+                async get(cid, from_jid) {
+                    const bobs = _converse.state.bobs;
```
jcbrand: Here you can also wait for the `BlobsInitialized` promise.

src/headless/plugins/bob/plugin.js
```diff
@@ -0,0 +1,214 @@
...
+
+                    if (this.has(cid)) {
+                        const bob = bobs.get(cid);
+                        return bob.getBlobURL();
+                    }
+
+                    if (from_jid) {
+                        try {
+                            await this.fetch(cid, from_jid);
+                            return this.get(cid);
```
jcbrand: The `return` statement should be outside of the `try/catch`

src/headless/plugins/bob/plugin.js
```diff
@@ -0,0 +1,214 @@
...
+                    const iq = $iq({
+                        type: 'get',
+                        to: from_jid
+                    }).c('data', {
+                        xmlns: 'urn:xmpp:bob',
+                        cid: cid
+                    });
+
+                    const result = await api.sendIQ(iq);
+                    const data_el = result.querySelector('data[xmlns="urn:xmpp:bob"]');
```
jcbrand: Use `Strophe.NS.BLOB`

src/headless/plugins/bob/plugin.js
```diff
@@ -0,0 +1,214 @@
...
+
+                    if (!data_el) {
+                        throw new Error('No BOB data in response');
```
jcbrand: Also log the resulting stanza as an error so that we can see what it was.

src/headless/plugins/bob/plugin.js
```diff
@@ -0,0 +1,214 @@
...
+                    if (cid && data) {
+                        bob_data.push({ cid, data, type, max_age });
+                        api.bob?.store(cid, data, type, max_age);
+                    }
+                });
+
+                if (bob_data.length > 0) {
+                    attrs.bob_data = bob_data;
+                }
+            } catch (e) {
```
jcbrand: I don't think there should be a `try/catch` here

src/plugins/bob-views/index.js
```diff
@@ -0,0 +1,58 @@
...
+ * Handles rendering BOB images in message bodies
+ */
+import { api, converse, log } from '@converse/headless';
+import tplImage from 'shared/texture/templates/image.js';
+
+/**
+ * Handle BOB images in message body after transformation
+ * @param {import('shared/texture/texture.js').Texture} texture
+ */
+async function handleBOBImages(texture) {
```
jcbrand: Please put this function in a `utils.js` file.

src/headless/plugins/bob/plugin.js
jcbrand: Looks good thanks!

src/shared/texture/texture.js
jcbrand: Thanks

jcbrand:
@sreeshanth-soma Please squash all your commits into a single one. It makes it easier to rebase later.

Also, please try to run `prettier` on the new files you create (but only those files, we can't yet run it on the whole codebase).

Otherwise looks pretty good already!

Have you tested this code with any other XMPP clients that use BOB?

sreeshanth-soma:
@jcbrand Done, squashed all commits into a single one and rebased onto the latest master.

Also ran prettier on all the new files (api.js, bob.js, bobs.js, plugin.js, utils.js, and the test file).

Regarding testing with other clients, the implementation closely follows the XEP-0231 specification and is validated through unit tests covering storing, retrieval, max-age expiry, size and type validation, IQ-based fetching, and error handling.

docs/source/configuration.rst
```diff
@@ -1144,6 +1144,18 @@ Set this option to ``true`` to request all pages of archived messages, but be
 aware that this can have performance implications.
 
 
+max_bob_size
+------------
+
+* Default: ``8192``
+
+The maximum size (in bytes) of Base64-decoded data that may be stored via
```
jcbrand:
```suggestion
The maximum size (in bytes) of Base64-encoded data that may be stored via
```

src/headless/plugins/bob/tests/bob.js
```diff
@@ -0,0 +1,166 @@
...
+        it(
+            'extracts BOB data from incoming messages',
+            mock.initConverse(['chatBoxesInitialized', 'BOBsInitialized'], {}, async (_converse) => {
+                await mock.waitForRoster(_converse, 'current', 1);
+                const contact_jid = mock.cur_names[0].replace(/ /g, '.').toLowerCase() + '@montague.lit';
+
+                const cid = 'cid:sha1+abc123@bob.xmpp.org';
+                const bobData = 'iVBORw0KGgoAAAANSUhEUg==';
+
+                const stanza = $msg({
```
jcbrand: Please use the `stx` tagged template literal everywhere instead of `$msg` (or `$pres` or `$iq`).

src/plugins/bob-views/utils.js
```diff
@@ -0,0 +1,36 @@
...
+            if (blob_url) {
+                const template = tplImage({
+                    src: blob_url,
+                    href: null,
+                    onClick: texture.onImgClick,
+                    onLoad: texture.onImgLoad,
+                });
+                texture.addTemplateResult(m.index + texture.offset, m.index + m[0].length + texture.offset, template);
+            }
+        } catch (e) {
```
jcbrand: Another `try/catch` that shouldn't be there.

jcbrand (commented):
Thanks for your effort so far @sreeshanth-soma 

> Regarding testing with other clients, the implementation closely follows the XEP-0231 specification and is validated through unit tests covering storing, retrieval, max-age expiry, size and type validation, IQ-based fetching, and error handling.

It's important to test the code with actual usage and not only with automated tests. Often certain difficult bugs like race conditions arise only with actual usage and you won't know about them if you only run unit tests.

In the original ticket it says the Pidgin sends BOB images as custom emojis and it also shows a screenshot.

To know whether your code works properly, you'll have to send a similar message from Pidgin and see whether it renders in Converse.

I see in the screenshot that if BOBs aren't supported, the shortname `bob` will be shown. It would be interesting to see how Pidgin includes the shortname when it sends a `message` stanza with BOB images.

When you send a message with BOB images from Pidgin to Converse, you can log the resulting stanza in Converse, then you can copy it and put it in a test, so that we can have a reproducible test.

aliraza556:
@sreeshanth-soma Are you working on this?

sreeshanth-soma:
@aliraza556 thanks for your concern. Yes, I am working on this.

sreeshanth-soma:
Hey @jcbrand, thanks for the review and especially for pushing me to test with an actual client. That suggestion was spot on.

I set up Pidgin with custom smileys against a local ejabberd server and it actually exposed a real bug. Took me a few days to figure out what was going on, but here's what I found:

**Pidgin needs two things we weren't handling:**

1. It checks for `http://jabber.org/protocol/xhtml-im` in disco — without that, it just sends plain text and skips the `` tag entirely. Took me a while to figure out why Pidgin was ignoring BOB support, turns out it was the caps cache + missing XHTML-IM feature.

2. It doesn't include BOB data inline at all. It only sends `` and expects the receiver to fetch the data via IQ-get. We only handled the inline case before.

Here's the actual stanza I captured from Pidgin:

```xml
<message from="user1@localhost/..." to="user2@localhost" type="chat">
  <active xmlns="http://jabber.org/protocol/chatstates"/>
  <html xmlns="http://jabber.org/protocol/xhtml-im">
    <body xmlns="http://www.w3.org/1999/xhtml">
      <p><img alt="test" src="cid:sha1+6ac9f63af35688...@bob.xmpp.org" /></p>
    </body>
  </html>
  <body>test</body>
</message>
```

No `<data xmlns="urn:xmpp:bob">` — just the cid reference.

Both issues are fixed now. All headless tests passing.

**Pidgin sending custom emoji to Converse.js via BOB:**
image


This whole process, debugging the caps cache, figuring out Pidgin's disco requirements, tracing the IQ flow  taught me a ton about how these XEPs actually work together in practice. Appreciate the guidance on this, it really pushed me to dig deeper. Let me know if there's anything else!

jcbrand:
Well done @sreeshanth-soma, thanks for the effort, I'll review as soon as I have time again.

jcbrand:
@sreeshanth-soma Thanks again for your work on this PR.

I was worried that the only way that Pidgin was sending the BOB images was via XHTML-IM.
We don't support XHTML-IM and likely never will.

I've come to the conclusion that BOB is a niche XEP that isn't particularly relevant for a modern XMPP client. I also noticed that Gajim decided the same: https://dev.gajim.org/gajim/gajim/-/issues/4993

I'm sorry that you did this work and now it won't get merged, but I don't think it's worth having the extra code, complexity and maintenance burden for such a fringe feature.

I'll still pay you the bounty, because it's not your fault that the PR doesn't get merged.

Please email me (see my Github bio) and we can arrange how to pay the bounty.

sreeshanth-soma:
Thank you @jcbrand, I have sent you an email. I appreciate it.

Neustradamus:
@sreeshanth-soma: Very good job!

@jcbrand: It is really a bad news...

Can you revert your choice?

You can see what projects support XEP-0231 here:
- https://xmpp.org/extensions/#xep-0231-implementations

PR #3941, opened 2026-02-18
add adhoc completed command result and text-multi as merged lines of text

### LINKED ISSUE #3940: Results of multi-step ad-hoc commands in text-multi forms are not shown
**Describe the bug**
The results of a multi-step ad-hoc command which are forms with text-multi fields are not displayed. Instead only the "completed" text is shown.

**To Reproduce**
1. Click on the gear icon in Converse.js.
2. Click commands.
3. Enter an entity where relevant adhoc commands are supported (e.g. http://jabber.org/protocol/admin#get-online-users-list)
4. Run the command
5. Fill out the first step
6. Hit "Next" or "Complete"
7. The form closes and the "Completed" text is displayed at the top.

**Expected behavior**
The form should stay open and display the text-multi field which is the result of the command.

**Screenshots**

Image

Click Next or Complete:

Image

 - Desktop
 - Browser: Firefox
 - Converse.js version: 12.0.0
 - xmpp server: prosody 0.12.4

**Additional context**
I see that in the debug log the result iq does come through and it has the correct `type="result"` tag. It's just not being displayed by Converse.js.

bobpaw:
- If there is a response type 'result' then display that instead of clearing the command.
- Parse the command result also if the status is completed and there is result data.
- Parse jid-multi and text-multi fields into object with values property.
- Output text-multi into a text area with values merged with newlines (XEP-0004 says SHOULD).
    - At first I output as a table but I think this is closer to the spec.
- also output jid-multi the same way.
- update multi step adhoc test.
- update XFormField in types.ts
- update XFORM_TYPE_MAP.
- add test for multi-step forms that return a result form
- update form_textarea.js template to remove whitespace when no description is present.

Closes #3940.

### DIFF
--- a/src/plugins/adhoc-views/tests/adhoc.js
+++ b/src/plugins/adhoc-views/tests/adhoc.js
@@ -964,10 +964,10 @@ describe("Ad-hoc commands consisting of multiple steps", function () {
         sel = `iq[to="${entity_jid}"][type="set"] command`;
         iq = await u.waitUntil(() => IQ_stanzas.filter(iq => sizzle(sel, iq).length).pop());
 
-        expect(Strophe.serialize(iq)).toBe(
-            `<iq id="${iq.getAttribute("id")}" to="${entity_jid}" type="set" xmlns="jabber:client">`+
-                `<command action="execute" node="get-online-users-list" xmlns="http://jabber.org/protocol/commands"/>`+
-            `</iq>`
+        expect(iq).toEqualStanza(stx`
+            <iq id="${iq.getAttribute("id")}" to="${entity_jid}" type="set" xmlns="jabber:client">
+                <command action="execute" node="get-online-users-list" xmlns="http://jabber.org/protocol/commands"/>
+            </iq>`
         );
 
         const sessionid = "f4d477d3-d8b1-452d-95c9-fece53ef99ad";


bobpaw:
I'm having a little bit of trouble applying the correct styling to the table I created from the text-multi field:

https://github.com/bobpaw/converse.js/blob/331039e5ba9938c517555ec2eee6c8ed01741f67/src/templates/form_textmulti.js

bobpaw:
For reference, here is the currently output table:

Screenshot 2026-02-19 at 8 26 39 AM

I think it should either have white text or a light background? The only table styles I see are defaults imported from bootstrap:

https://github.com/bobpaw/converse.js/blob/331039e5ba9938c517555ec2eee6c8ed01741f67/src/shared/styles/website.scss#L13

bobpaw:
I'm wondering if the following style which uses "list-group" and "list-group-item" would be more consistent?

Screenshot 2026-02-19 at 3 53 26 PM

```diff
diff --git a/src/templates/form_textmulti.js b/src/templates/form_textmulti.js
index a36bbb1f0..e281701cb 100644
--- a/src/templates/form_textmulti.js
+++ b/src/templates/form_textmulti.js
@@ -5,19 +5,16 @@ export default (o) => {
     const id = u.getUniqueId();
     return html`
         <div class="mb-3">
-            <table class="table" name="${o.name}" id="${id}" data-var="${o.var}">
-                <thead class="thead-light">
-                    <th scope="col">
-                        ${o.label}
-                        ${(o.desc) ? html`<small class="form-text text-muted">${o.desc}</small>` : ''}
-                    </th>
-                </thead>
-                <tbody>
-                    ${o.values?.map(
-                        (value) => html`<tr><td>${value}</td></tr>`
-                    )}
-                </tbody>
-            </table>
+            <ul name="${o.name}" id="${id}" data-var="${o.var}" class="list-group">
+                <li class="list-group-item active">
+                    ${o.label}
+                    ${(o.desc) ? html`<small class="form-text text-muted">${o.desc}</small>` : ''}
+                </li>
+                ${o.values?.map(
+                    (value) => html`<li class="list-group-item">${value}</li>`
+                )}
+            </ul>
         </div>
     `;
 };
```

src/utils/html.js
```diff
@@ -382,7 +382,20 @@ export function xFormField2TemplateResult(xfield, options = {}) {
...
+                ? xfield.values.reduce((acc, x) => acc + '\n' + x) : ''
+        });
+
+    } else if (xfield['type'] === 'text-multi') {
+        return tplFormTextarea({
+            ...default_vals,
+            ...xfield,
+            value: (xfield !== undefined && xfield.values.length !== 0)
+                ? xfield.values.reduce((acc, x) => acc + '\n' + x) : ''
+        });
```
jcbrand: You could also just have a single `else if` block:

```
} else if (xfield['type'] === 'jid-multi' || xfield['type'] === 'text-multi') {
```

src/headless/shared/parsers.js
```diff
@@ -506,12 +506,23 @@ function parseXFormField(field, readonly, stanza) {
...
+        };
+    } else if (type === 'text-multi') {
+        const values = Array.from(field.querySelectorAll(':scope > value')).map((el) => el?.textContent);
+        return {
+            type,
+            label,
+            var: v,
             required: !!field.querySelector('required'),
+            values,
             ...result,
```
jcbrand: Looks like you can have a single `else if` block here:

```
 } else if (type === 'jid-multi' || type === 'text-multi') {
```

bobpaw:
> Yes that looks much better.
> 
> I don't see that code in the PR though.

After looking at the XEP I think `text-multi` fields basically need to be presented as a `<textarea>`. For the online users example I was looking at it looks nice because it's basically a list, but I think `text-multi` can be arbitrary multiline text?

It might be worth differentiating `jid-multi` here since those fields are pretty clearly always lists, but that would also require adding a UI element for editing (add row/remove row). This is the approach I have seen other clients take. And maybe individual `<input>` elements for each row.

jcbrand:
Thanks for your contribution so far @bobpaw 

> but I think text-multi can be arbitrary multiline text?

Yes, it should be a `textarea`, which is the UI element for arbitrary multiline text.

> It might be worth differentiating jid-multi here since those fields are pretty clearly always lists, but that would also require adding a UI element for editing (add row/remove row)

Yes I agree. If you're willing to do the work for this, Choices.js looks like a good candidate for implementing a UI element for `jid-multi`.

https://choices-js.github.io/Choices/
https://github.com/Choices-js/Choices

I'd be happy to review and provide guidance.

I think that work should be done in a separate PR though.

jcbrand:
@bobpaw Would you say this PR can be merged as it currently is? It looks basically complete to me, although a test to show that `response.type === 'result'` is properly handled would be great.

jcbrand:
Another alternative to choices.js is Tom Select (great name).

https://tom-select.js.org/

It appears to be a bit smaller, has a [Bootstrap5 stylesheet](https://github.com/orchidjs/tom-select/blob/master/src/scss/tom-select.bootstrap5.scss)

It has a plugin architecture, so if we drop unused plugins it might be even smaller.

bobpaw:
> @bobpaw Would you say this PR can be merged as it currently is? It looks basically complete to me, although a test to show that `response.type === 'result'` is properly handled would be great.

Let me see about adding this test and then yes!

bobpaw:
Choices.js and Tom Select both seem like good options. The version of Tom Select without plugins is about 14kb gzipped and minified, so that might be the better choice based on size alone.

bobpaw:
Added a new test case based on the original problematic adhoc command. Also updated the form_textarea template to remove the whitespace added when no description is present.

bobpaw:
@jcbrand this should be good to merge now!

src/plugins/adhoc-views/tests/adhoc.js
```diff
@@ -900,4 +910,157 @@ describe("Ad-hoc commands consisting of multiple steps", function () {
...
+
+        const items = adhoc_form.querySelectorAll('.list-group-item:not(.active)');
+        expect(items.length).toBe(1);
+        expect(items[0].textContent.trim()).toBe('Get List of Online Users');
+        items[0].querySelector('a').click();
+
+        sel = `iq[to="${entity_jid}"][type="set"] command`;
+        iq = await u.waitUntil(() => IQ_stanzas.filter(iq => sizzle(sel, iq).length).pop());
+
+        expect(Strophe.serialize(iq)).toBe(
```
jcbrand: Please use `toEqualStanza` isntead of `Strophe.serialize` and `toBe`.

src/plugins/adhoc-views/tests/adhoc.js
```diff
@@ -900,4 +910,157 @@ describe("Ad-hoc commands consisting of multiple steps", function () {
...
+                            <value>${entity_jid}</value>
+                        </field>
+                    </x>
+                </command>
+            </iq>`
+        ));
+
+        const dummy = await u.waitUntil(() => adhoc_form.querySelector('form form textarea'));
+
+        const form = await u.waitUntil(() => adhoc_form.querySelector('form form'));
```
jcbrand:
```suggestion
        await u.waitUntil(() => adhoc_form.querySelector('form form textarea'));
        const form = await u.waitUntil(() => adhoc_form.querySelector('form form'));
```

Side-note, seems kinda weird that there is a form inside a form. 🤔

src/plugins/adhoc-views/tests/adhoc.js
```diff
@@ -900,4 +910,157 @@ describe("Ad-hoc commands consisting of multiple steps", function () {
...
+        expect(form).toBeDefined();
+
+        const labels = Array.from(form.querySelectorAll('label'));
+        expect(labels.length).toBe(1);
+        expect(labels[0].textContent).toEqual('The list of all online users');
+
+        const textareas = Array.from(form.querySelectorAll('textarea'));
+        expect(textareas.length).toBe(1);
```
jcbrand, line 1063: Another potential improvement that could be made would be to not show the form input elements for `readonly` values.

I see we set the `readonly` attribute on the `textarea` which is good, but I think it would be a nicer UI if we didn't show a textarea at all (and same for other form elements).

Instead, for `text-multi` and `jid-multi` we could show something like the Bootstrap list-group:
https://getbootstrap.com/docs/5.3/components/list-group/

It would be a matter of updating `src/templates/form_textarea.js` to either render the `textarea` when its not readonly and otherwise to render a list-group (and then doing the same for any other relevant form elements).

WDYT?

But again, if you were up to doing that, I would rather it went into a new PR to making reviewing easier.

src/plugins/adhoc-views/tests/adhoc.js
bobpaw: Oops. I used the "May return a result form with readonly fields" as a template and it used the same expect clause.

src/plugins/adhoc-views/tests/adhoc.js
bobpaw: Yeah I was confused by the `'form form'` at first as well but it's just because the list of adhoc commands is a `<form>`.

src/plugins/adhoc-views/tests/adhoc.js
bobpaw: Yeah, I think it would be worth changing the display of result forms.

Another example: right now the `<field var="FORM_TYPE" type="hidden">` results in an invisible `<div>`.

I think `jid-multi` should be a list-group like you recommended. I think maybe `text-multi` might be better as a `<pre>` element? I'm going to open a new issue to track the discussion.

src/plugins/adhoc-views/tests/adhoc.js
jcbrand: Yes, the code is full of deprecated patterns and conventions which need to be replaced.

jcbrand:
Thanks @bobpaw! Merged manually.

Issue #3939, opened 2026-02-17
"Invitation to groupchat" prompt appears duplicatively


jstevans:
**Describe the bug**

I've just started hosting a Snikket server per [their quick-start guide](https://snikket.org/service/quickstart/). I've created a MUC, to which I've invited someone else and sent a message `"test"`.

I'm definitely a member of this MUC, but now I keep getting the "Invitation to a groupchat" prompt for that MUC.

**To Reproduce**

From https://chat.conversejs.org/:
1. join a newly-setup Snikket server
1. Create a MUC
2. Invite someone else
3. Send a message `"test"`
4. Refresh the browser page and select the MUC (does not repro 100% of the time)

**Expected behavior**

Chatroom shows and no invitation prompt is given

**Screenshots**

On some refreshes, the below invitation will appear (note the lack of an inviter at the beginning of the text):
<img width="505" height="247" alt="Image" src="https://github.com/user-attachments/assets/affbacfe-c982-4186-adea-34d38e3f8c0b" />

**Environment (please complete the following information):**
 - Desktop
 - Firefox
 - Converse.js version: 12.0.0

**Additional context**
I see the below XML, which IIUC corresponds to [this code](https://github.com/conversejs/converse.js/blob/02a50fc1c9ed4d8812242acbd31a6157957468e8/src/headless/plugins/muc/utils.js#L58-L68). Should there be a check [around here](https://github.com/conversejs/converse.js/blob/master/src/headless/plugins/muc/utils.js#L134) whether the user is already a member of the room/chatbox?

```xml
<message to='johnst@xmpp.<myserver>/converse.js-117166079' xmlns='jabber:client'>
    <result queryid='25822713-35d2-4582-829a-90c3df793191' xmlns='urn:xmpp:mam:2'
        id='019c63cc-b024-74fd-a4ae-49620c17e05d'>
        <forwarded xmlns='urn:xmpp:forward:0'>
            <delay xmlns='urn:xmpp:delay' stamp='2026-02-16T00:14:44.772090Z' />
            <message to='friend@xmpp.<myserver>' id='842b4efc-a480-486a-8dc0-1d066d3fd823'
                from='johnst@xmpp.<myserver>/converse.js-117166079' xmlns='jabber:client'
                xml:lang='en'>
                <x jid='<my-muc>@groups.xmpp.<myserver>' xmlns='jabber:x:conference' />
            </message>
        </forwarded>
    </result>
</message>
```

[...]

PR #3942, opened 2026-02-20
fix(#3939): Don't show invitations to groupchats in which the user is already present

jstevans:

### DIFF
--- a/src/headless/plugins/muc/tests/muc.js
+++ b/src/headless/plugins/muc/tests/muc.js
@@ -157,10 +157,12 @@ describe("Groupchats", function () {
             await mock.openAndEnterMUC(_converse, muc_jid, 'romeo');
             const model = _converse.chatboxes.get(muc_jid);
 
+            expect(api.rooms.get(model.muc_jid)).not.toBeNull();
+
             spyOn(_converse.exports, 'onDirectMUCInvitation').and.callThrough();
             spyOn(api, 'hook').and.callThrough();
             const apiRoomsGet = api.rooms.get;
-            spyOn(api.rooms, 'get').and.callFake(apiRoomsGet(...arguments));
+            spyOn(api.rooms, 'get').and.callFake(function () { return apiRoomsGet(...arguments); });
 
             expect(model.session.get('connection_status')).toBe(converse.ROOMSTATUS.ENTERED);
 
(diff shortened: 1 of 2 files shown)


src/headless/plugins/muc/utils.js
```diff
@@ -131,7 +131,14 @@ export async function onDirectMUCInvitation(message) {
         reason = x_el.getAttribute('reason');
 
     let result;
-    if (api.settings.get('auto_join_on_invite')) {
+
+    const { chatboxes } = _converse.state;
+    const rooms = chatboxes.filter((m) => m.get('type') === CHATROOMS_TYPE && !m.get('closed'));
```
jstevans: @jcbrand I tried switching this to `api.rooms.get(room_jid)` but kept getting back `null` when it executed in my tests.

I see other examples where it's used that way, but perhaps I'm passing something incorrectly?

Or is there some setup I'm missing to properly simulate joining a room?

src/headless/plugins/muc/tests/muc.js
```diff
@@ -148,5 +148,46 @@ describe("Groupchats", function () {
...
+            const muc_jid = 'coven@chat.shakespeare.lit';
+            await mock.openAndEnterMUC(_converse, muc_jid, 'romeo');
+            const model = _converse.chatboxes.get(muc_jid);
+
+            spyOn(_converse.exports, 'onDirectMUCInvitation').and.callThrough();
+            spyOn(api, 'hook').and.callThrough();
+            const apiRoomsGet = api.rooms.get;
+            spyOn(api.rooms, 'get').and.callFake(function () {
+                apiRoomsGet(...arguments)
+            });
```
jcbrand: You're missing a `return` statement, so the room doesn't actually get returned.

```suggestion
            const apiRoomsGet = api.rooms.get;
            spyOn(api.rooms, 'get').and.callFake(apiRoomsGet(...arguments));
```

src/headless/plugins/muc/utils.js
jcbrand: I think the issue is with how you mock the function. You're missing a `return` statement, so `undefined`
is returned.

See here: https://github.com/conversejs/converse.js/pull/3942/changes#r2839235015

jcbrand:
Thanks @jstevans! I've merged manually.

Issue #3509, opened 2024-09-11
Pin conversation(s) to top


roughnecks:
Hello,

haven't seen this on the tracker yet but now that we have conversations ordered by name, it would also be cool to be able to pin some of those to the top, so that you can have your favorite chats at hand's reach.

It's a bit of a pain to have to scroll the list of conversations to reach some on the bottom, so this fix could mitigate the frustration partly.

Thanks

marclaporte:
@roughnecks Any idea if a XEP exists for this? If I toggle between a desktop client and Converse.js, I'd like the same behavior.

roughnecks:
I think this one is it? https://xmpp.org/extensions/xep-0469.html

Neustradamus:
There is this PR:
- https://github.com/conversejs/converse.js/pull/4088

It replaces:
- https://github.com/conversejs/converse.js/pull/3949

PR #3949, opened 2026-02-28
Feature: Allow pinning conversations to the top (XEP-0469)

mercihabam:
This is an implementation of https://github.com/conversejs/converse.js/issues/3509, which allows users to pin bookmarked MUC rooms to the top.

Here is a quick overview of how it looks:
Screenshot 2026-03-14 at 15 34 08
Screenshot 2026-03-14 at 15 34 33
Screenshot 2026-03-14 at 15 34 48


mercihabam:
Before adding tests for the many changes, I would appreciate a review from the maintainers. Thanks for taking the time to look into this!

mercihabam:
I went ahead and added tests for the changes. Please review this when you get a chance. Thanks!

mercihabam:
@jcbrand could you please take a look at this when you get the chance? Thanks.

jcbrand:
@mercihabam Please rebase your PR onto the latest master branch.

mercihabam:
@jcbrand that is now done.

mercihabam:
@jcbrand don't forget this one please. 😊

jcbrand:
I won't forget, but I'm first finishing off the emoji reactions and popular emojis work. Then I'll make a new release and then I'll get back to all the outstanding PRs.

jcbrand:
@mercihabam There is a failing test which looks like it might be related to your changes:

```
Chrome 147.0.0.0 (Linux 0.0.0) A MUC domain group is collapsible FAILED
	Expected true to be falsy.
	    at <Jasmine>
	    at src/plugins/roomslist/tests/grouplists.js:118:52
	    at async UserContext.<anonymous> (src/headless/tests/mock.js:625:13)
Chrome 147.0.0.0 (Linux 0.0.0): Executed 479 of 625 (1 FAILED) (0 secs / 46.029 secs)
Chrome 147.0.0.0 (Linux 0.0.0) A MUC domain group is collapsible FAILED
	Expected true to be falsy.
	    at <Jasmine>
	    at src/plugins/roomslist/tests/grouplists.js:118:52
	    at async UserContext.<anonymous> (src/headless/tests/mock.js:625:13)
 ```
 
 Please also squash your commits into one and rebase onto the latest `master` branch.

**Comment by mercihabam** [comment]
> @mercihabam There is a failing test which looks like it might be related to your changes:

I did not see that one because some tests are skipped on my instance; I'm still figuring out why. But now I have fixed that, rebased onto master, and squashed the commits. Thanks!

File: src/plugins/bookmark-views/components/templates/pin-list.js [hunk 1]
```diff
@@ -0,0 +1,43 @@
...
+ */
+
+import { constants } from "@converse/headless";
+import { __ } from "i18n";
+import { html } from "lit";
+import { tplRoomItem } from "shared/roomslist/templates/room-item";
+import '../styles/pin-list.scss';
+
+/**
+ * @param {BookmarksPinView} el
```
jcbrand:
```suggestion
 * @param {import('plugins/bookmark-views/components/bookmarks-pin-list').BookmarksPinView} el
```

It's better to avoid `@typedef` at the top of the file if possible since those types get re-exported from the file if they're defined at module level (know JSDoc/types wart).

src/headless/plugins/bookmarks/tests/bookmarks.js
```diff
@@ -555,4 +555,149 @@ describe('A bookmark', function () {
...
+            const IQ_stanzas = api.connection.get().IQ_stanzas;
+
+            // Now pin the bookmark
+            const bookmark = state.bookmarks.findWhere({ jid: muc_jid });
+            expect(bookmark).toBeTruthy();
+            await state.bookmarks.pinBookmark(bookmark);
+
+
+            const sent_stanza = await u.waitUntil(() =>
+                IQ_stanzas.filter((s) => sizzle('publish[node="urn:xmpp:bookmarks:1"] conference[name="The Play"] extensions pinned', s).length).pop()
```
jcbrand:
```suggestion
                IQ_stanzas.filter((s) => sizzle('`publish[node="Strophe.NS.BOOKMARKS2"] conference[name="The Play"] extensions pinned`, s).length).pop()
```

We assign the XML namespaces to `Strophe.NS`.  Use that everywhere instead of the hardcoded string value. That makes it easier to upgrade when there's a version bump in the namespace.

src/headless/plugins/bookmarks/collection.js
```diff
@@ -334,6 +336,52 @@ class Bookmarks extends Collection {
         const { chatboxes } = _converse.state;
         return this.filter((b) => !chatboxes.get(b.get('jid')));
     }
+
+    /**
+     * 
+     * @param {Bookmark} bookmark
+     */
+    pinBookmark(bookmark) {
+        const extensions = [...bookmark.get('extensions'), '<pinned xmlns="urn:xmpp:bookmarks-pinning:0"/>'];
```
jcbrand:
```suggestion
        const extensions = [...bookmark.get('extensions'), `<pinned xmlns="Strophe.NS.BOOKMARKS_PINNING"/>`];
```

You'll need to defined `Strophe.NS.BOOKMARKS_PINNING` in src/headless/plugins/bookmarks/plugin.js

src/plugins/bookmark-views/components/bookmarks-pin-list.js
```diff
@@ -0,0 +1,38 @@
...
+    initialize() {
+        const bare_jid = _converse.session.get('bare_jid');
+        const id = `converse.bookmarks-pin-list-model-${bare_jid}`;
+        this.model = new BookmarksPinListModel({ id });
+        _converse.state.bookmarks_pin_list = this.model;
+
+        initStorage(this.model, id);
+        this.model.fetch();
+
+        this.handleEvents();
```
jcbrand: I think this method should be renamed.

```suggestion
        this.addEventListeners();
```

You're not really handling events, you're just adding listeners for events.

src/plugins/bookmark-views/components/model.js
```diff
@@ -0,0 +1,9 @@
+import { constants, Model } from "@converse/headless";
+
+export default class BookmarksPinListModel extends Model {
```
jcbrand: I don't think a new type is necessary. You can just do `new Model({ toggle_state: constants.OPENED });` instead.

src/plugins/controlbox/templates/controlbox.js
```diff
@@ -47,6 +47,9 @@ export default (el) => {
             ? html`<converse-user-profile></converse-user-profile>
                   <div class="controlbox-pane">
                       <converse-headlines-feeds-list class="controlbox-section"></converse-headlines-feeds-list>
+                      <div id="converse-bookmarks-pin" class="controlbox-section">
+                        <converse-bookmarks-pin></converse-bookmarks-pin>
+                    </div>
```
jcbrand: Would be a bit cleaner if this was possible:

```suggestion
                    <converse-bookmarks-pin class="controlbox-section"></converse-bookmarks-pin>
```

src/plugins/bookmark-views/components/bookmarks-pin-list.js
```diff
@@ -0,0 +1,38 @@
...
+        rooms.sort((a, b) => (a.getDisplayName().toLowerCase() <= b.getDisplayName().toLowerCase() ? -1 : 1));
+        return rooms;
+    }
+
+    render() {
+        return tplBookmarksPinList(this);
+    }
+}
+
+api.elements.define('converse-bookmarks-pin', BookmarksPinView);
```
jcbrand: I think `converse-pinned-bookmarks` and `PinnedBookmarksView` are better names.

```suggestion
api.elements.define('converse-pinned-bookmarks', PinnedBookmarksView);
```

src/plugins/roomslist/view.js
```diff
@@ -52,7 +56,7 @@ export class RoomsList extends CustomElement {
     /** @returns {import('@converse/headless').MUC[]} */
     getRoomsToShow() {
         const { chatboxes } = _converse.state;
-        const rooms = chatboxes.filter((m) => m.get('type') === CHATROOMS_TYPE && !m.get('closed'));
+        const rooms = chatboxes.filter((m) => m.get('type') === CHATROOMS_TYPE && !m.get('closed') && !m.get('pinned'));
```
jcbrand: This change looks relevant to the pinned bookmarks list, but not the normal groupchats list.

If a bookmark is pinned, will it show in both lists? I think it should only show in the `Pinned Bookmarks` list.

src/headless/plugins/bookmarks/collection.js
```diff
@@ -251,6 +252,7 @@ class Bookmarks extends Collection {
         const { chatboxes } = _converse.state;
         const groupchat = chatboxes.get(bookmark.get('jid'));
         groupchat?.save('bookmarked', true);
+        groupchat?.save('pinned', bookmark.pinned);
```
jcbrand: If possible, it's preferable that there is a single source of truth regarding whether a groupchat is pinned or not, and that should be the bookmark.

Otherwise you can get inconsistencies between `groupchat.get('pinned')` and `bookmark.get('pinned')`

Every bookmarked groupchat should have a `bookmark` attribute, so this should be doable.

If you need to re-render the groupchat when a bookmark is pinned or unpinned, you can trigger an event on the groupchat that it can listen to. We follow this same pattern with VCard changes.

https://github.com/conversejs/converse.js/blob/f951289e91054695ed8e9305868766042995a313/src/headless/plugins/vcard/utils.js#L87

jcbrand (commented):
Thanks @mercihabam, please see my review comments.

src/plugins/roomslist/view.js
mercihabam: Sure, that's the reasoning here; showing the bookmark only in one list whether it's pinned or not. The way I'm seeing this should be changed so that the normal groupchats list doesn't have to deal with the `pinned` state is to separate the two states at the Chatboxes collection level. Would that be better?

src/headless/plugins/bookmarks/collection.js
mercihabam: Great, thanks for the code reference. The `renderIfRelevantChange()` method of the `RoomsList` object is handling the re-rendering of the group chat when pinning or unpinning a bookmark. Should this change to the VCard pattern, or do that in parallel?

jcbrand: If you trigger a new event `bookmark:change` (similar to `vcard:change`), then that's a different event from `change` and the two events won't fire in parallel because the bookmark model changing won't trigger `change` on the `chatboxes` model.

So I don't think you need to do anything to the `renderIfRelevantChange` function.

You can probably just do this:

```
        this.listenTo(chatboxes, 'bookmark:change', () => this.requestUpdate());
```

src/plugins/roomslist/view.js
jcbrand: Looking at this again, the code makes sense and I misunderstood what was happening here.

However, there's a deeper issue here. I would like there to be a single source of truth, similar to what we do with VCards, but with VCards we have the `ModelWithVCard` mixin which makes sure that we have a `.vcard` attribute stored on the respective model.

We don't have the same thing for bookmarks and MUCs. So to have a single source of truth, we might have to add a `ModelWithBookmark` mixin, similar to the `ModelWithVCard`.

That way, we can then know that a MUC will have a `bookmark` attribute pointing to the bookmark instance and we can read `pinned` from that, and not from the MUC model itself.

Then the line above changes like this:

```suggestion
        const rooms = chatboxes.filter(
            (m) => m.get('type') === CHATROOMS_TYPE &&
                !m.get('closed') &&
                !m.bookmark?.get('pinned')
        );
```

Does that make sense?

jcbrand: If you do make this change, please keep your commits separate, don't squash, so that we can get back to the old state if necessary.

mercihabam: Yes, that makes sense and is much better.

mercihabam:
@jcbrand I have refactored this as discussed. A few tests are now failing, but I will fix them soon. Thanks!

jcbrand:
Closed in favor of #4088

PR #3980, opened 2026-03-22
feat: require current password when changing password (#3120)

### LINKED ISSUE #3120: Ask user to provide current password in order to reset password

FloodExLLC:
## Summary

- Adds a "Current password" field to the password reset form (closes #3120)
- Verifies the current password against the stored connection password when available, preventing unauthorized password changes when someone has access to the browser
- Adds error message display when the current password is incorrect

## Changes

- `src/plugins/profile/password-reset.js`: Added `current_password_error` property and `checkCurrentPassword()` method
- `src/plugins/profile/templates/password-reset.js`: Added "Current password" input field with error display
- `src/plugins/profile/tests/password-reset.js`: Added test for incorrect current password, updated helper to include current password

## Test plan

- [x] Existing password reset tests pass with updated helper (uses correct password `secret`)
- [x] New test verifies that submitting with wrong current password shows error and sends no IQ stanzas
- [x] Form requires current password field (marked `required`)

?? Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/plugins/profile/password-reset.js
+++ b/src/plugins/profile/password-reset.js
@@ -11,12 +11,14 @@ class PasswordReset extends CustomElement {
     static get properties () {
         return {
             passwords_mismatched: { type: Boolean },
+            current_password_error: { type: Boolean },
             alert_message: { type: String }
         }
     }
 
     initialize () {
         this.passwords_mismatched = false;
+        this.current_password_error = false;
         this.alert_message = '';
     }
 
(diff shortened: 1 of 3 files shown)


src/plugins/profile/password-reset.js
```diff
@@ -33,9 +35,22 @@ class PasswordReset extends CustomElement {
         return this.passwords_mismatched
     }
 
+    checkCurrentPassword (ev) {
+        const form_data = new FormData(ev.target.form ?? ev.target);
+        const current_password = form_data.get('current_password');
+        const connection = api.connection.get();
+        if (connection?.pass && current_password !== connection.pass) {
```
jcbrand, line 42: This isn't going to work in all cases. When you reconnected via BOSH or XEP-0198 then the password won't be set on the connection object.

jcbrand: To properly test the existing password, you might have to create a new Strophe.Connection object (using the subclass in Converse is likely overkill) and then test logging in with that connection and the provided password.

PR #3983, opened 2026-04-02
Fix #1851: Sort open groupchats alphabetically when grouped by domain

### LINKED ISSUE #1851: Open Groupchats no longer in alphabetical order
**Describe the bug**
The list of Open Groupchats used to be in alphabetical order (based on (bookmark?) name). Now it appears to be randomly ordered, possibly by join order.

**To Reproduce**
1. Have a bunch of bookmarks with autojoin set.
2. Connect with Converse.js
3. Observe *Open Groupchats*
4. Wonder what they're supposed to be ordered by.

**Expected behavior**
The list ought to be ordered by something. Alphabetical or by activity would be nice. Random ordering makes it difficult to find chats.

**Environment:**
 - Desktop
 - Firefox 72.0.2 on Windows
 - Converse.js version 6.0.0 (fullscreen mode)

cj-l920:
## Problem Summary
Fixes #1851: Open Groupchats not sorted in any order

When the `muc_grouped_by_domain` configuration option is enabled, rooms within each domain group were not being sorted alphabetically. This made it difficult for users to find specific groupchats in the list.

## Solution Approach
Added alphabetical sorting to the `tplRoomDomainGroupList` function in `src/plugins/roomslist/templates/roomslist.js`.

The fix ensures that:
1. Rooms are sorted by their display name (case-insensitive) before being grouped by domain
2. The existing domain sorting behavior is maintained
3. The sort is stable and consistent

## Changes Made
- Modified `tplRoomDomainGroupList()` to sort rooms alphabetically using `getDisplayName()`
- Used case-insensitive comparison (`toLowerCase()`) for consistent ordering
- Added a copy of the rooms array to avoid mutating the original

## Test Evidence
- [x] Code compiles without errors
- [x] Sorting logic follows the same pattern used in `getRoomsToShow()` in view.js
- [x] Case-insensitive sorting ensures "Room A" and "room a" are ordered correctly
- [x] Original array is not mutated (uses spread operator copy)

## Example
Before (random order within domain):
example.com
Zebra Chat
Alpha Room
Beta Group
After (alphabetically sorted):
example.com
Alpha Room
Beta Group
Zebra Chat
Fixes #1851

### DIFF
--- a/src/plugins/roomslist/templates/roomslist.js
+++ b/src/plugins/roomslist/templates/roomslist.js
@@ -103,9 +103,15 @@ function tplRoomDomainGroup (el, domain, rooms) {
  * @param {MUC[]} rooms
  */
 function tplRoomDomainGroupList (el, rooms) {
-    // The rooms should stay sorted as they are iterated and added in order
+    // Sort rooms alphabetically by display name first
+    const sorted_rooms = [...rooms].sort((a, b) => {
+        const nameA = a.getDisplayName().toLowerCase();
+        const nameB = b.getDisplayName().toLowerCase();
+        return nameA <= nameB ? -1 : 1;
+    });
+
     const grouped_rooms = new Map();
-    for (const room of rooms) {
+    for (const room of sorted_rooms) {
         const roomdomain = room.get('jid').split('@').at(-1).toLowerCase();
         if (grouped_rooms.has(roomdomain)) {
             grouped_rooms.get(roomdomain).push(room);


src/plugins/roomslist/templates/roomslist.js
```diff
@@ -103,9 +103,15 @@ function tplRoomDomainGroup (el, domain, rooms) {
  * @param {MUC[]} rooms
  */
 function tplRoomDomainGroupList (el, rooms) {
-    // The rooms should stay sorted as they are iterated and added in order
+    // Sort rooms alphabetically by display name first
+    const sorted_rooms = [...rooms].sort((a, b) => {
+        const nameA = a.getDisplayName().toLowerCase();
+        const nameB = b.getDisplayName().toLowerCase();
+        return nameA <= nameB ? -1 : 1;
+    });
```
jcbrand, line 111: The rooms are already sorted in `getRoomsToShow`, so it's not clear to me that this is a real issue.

https://github.com/conversejs/converse.js/blob/6b1ad62ec4e6bd4570975362eb94a97198fa1d51/src/plugins/roomslist/view.js#L53

PR #3995, opened 2026-04-23
Fix avatars not showing when not using HTTPS

### LINKED ISSUE #2374: Avatars not showing when not using HTTPS
The rendering of avatars from vCards uses `crypto.subtle`. This feature is available only in secure contexts (HTTPS), in some or all supporting browsers ([source](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/subtle)).

This results in avatars not being shown when the Converse webapp is opened on a non-secure context.

Screenshot from 2020-12-24 17-18-50
Screenshot from 2020-12-24 17-18-35

jiafeimao0p:
## Summary
Fixes the issue where avatars are not showing when Converse.js is opened on a non-secure context (non-HTTPS).

## Problem
The rendering of avatars from vCards uses `crypto.subtle`, which is only available in secure contexts (HTTPS). This causes avatars to not be displayed in non-HTTPS environments.

## Solution
Check for `window.isSecureContext` before using the `crypto.subtle` API, and provide a fallback for non-HTTPS environments.

## Files Changed
- `src/headless/plugins/vcard/parsers.js` - Check secure context before SHA-1 hash
- `src/headless/plugins/vcard/api.js` - Check secure context when setting avatar
- `src/headless/utils/color.js` - Check secure context in colorize function

## Testing
The fallback uses base64 substring or simple string hash as pseudo-hash when `crypto.subtle` is unavailable, allowing avatars to display in non-HTTPS contexts.

Fixes #2374

### DIFF
--- a/src/headless/plugins/vcard/api.js
+++ b/src/headless/plugins/vcard/api.js
@@ -50,11 +50,19 @@ export default {
                 // Optimistically update the vcard with image data. Otherwise some servers (e.g. Ejabberd)
                 // could send a XEP-0153 vcard:update presence which would cause us to refetch the vcard again.
                 const buffer = u.base64ToArrayBuffer(data.image);
-                const hash_ab = await crypto.subtle.digest('SHA-1', buffer);
+                // Check if crypto.subtle is available (requires secure context/HTTPS)
+                let image_hash;
+                if (window.isSecureContext) {
+                    const hash_ab = await crypto.subtle.digest('SHA-1', buffer);
+                    image_hash = u.arrayBufferToHex(hash_ab);
+                } else {
+                    // Fallback for non-HTTPS contexts: use base64 as pseudo-hash
+                    image_hash = data.image.substring(0, 32);
+                }
                 vcard.save({
                     image: data.image,
                     image_type: data.image_type,
-                    image_hash: u.arrayBufferToHex(hash_ab),
+                    image_hash: image_hash,
                 });
             }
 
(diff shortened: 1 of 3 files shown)


src/headless/plugins/vcard/api.js
```diff
@@ -50,11 +50,19 @@ export default {
...
                 const buffer = u.base64ToArrayBuffer(data.image);
-                const hash_ab = await crypto.subtle.digest('SHA-1', buffer);
+                // Check if crypto.subtle is available (requires secure context/HTTPS)
+                let image_hash;
+                if (window.isSecureContext) {
+                    const hash_ab = await crypto.subtle.digest('SHA-1', buffer);
+                    image_hash = u.arrayBufferToHex(hash_ab);
+                } else {
+                    // Fallback for non-HTTPS contexts: use base64 as pseudo-hash
+                    image_hash = data.image.substring(0, 32);
```
jcbrand, line 60: The sha1 hash is used to keep track of whether VCard avatars have changed, and is a requirement of [XEP-0153](https://xmpp.org/extensions/xep-0153.html).

If you change it to using base64, then you're breaking compliance with XEP-0153 and you're also signaling to other XMPP clients that the avatar has changed.

This might even cause a continuous loop of avatar fetching because the sha1 hash will never match the base64 value.

The only proper way to do this that I can think of would be to dynamically load a sha1 hash library and use that if `crypto.subtle` is not available.

jcbrand:
Thanks for your PR @jiafeimao0p, please see my review comment.

PR #4013, opened 2026-05-01
chore: replace older Strophe.Builder calls with stx

jcbrand:


src/headless/plugins/blocklist/tests/blocklist.js
```diff
@@ -5,7 +5,6 @@ const { u, stx } = converse.env;
 
 describe('A blocklist', function () {
     beforeEach(() => {
```
linkmauve, line 7: Shouldn’t you also remove that `beforeEach()` block, now that it does nothing?

PR #4037, opened 2026-05-20
fix: keep audio play button available while loading

### LINKED ISSUE #3382: audio player accessibility
**Describe the bug**
Screen reader users can't play audio. 

**To Reproduce**
1. Use a screen reader and keyboard to go to the play button on an audio file in a chat, such as at https://freespoken.nz/bible-radio/
2. Press Enter
3. Discover that the play button is unavailable. 

**Expected behavior**
The audio file should play. 

 - Desktop or Mobile?  Windows 11 with NVDA, or JAWS
 - Browser [e.g. Chrome, Firefox] Not sure
 - Converse.js version [e.g. 4.2.0] 10.1.7, I think

**Additional context**
You can see the comments of the person who actually tested it in the replies to this toot https://theres.life/@TMakarios/112278231413032952

rafael226:
Fixes #3382

## Summary
- keep the custom audio player's Play button enabled while audio metadata is still loading
- preserve the disabled state for actual audio errors
- add a regression test that verifies the loading Play button remains focusable and keyboard-activates playback

## Tests
- npm run dev
- npx eslint src/plugins/chatview/tests/message-audio.js src/shared/texture/components/audio-player.js
- npm test -- --single-run with only the new focused test temporarily enabled: 1 SUCCESS

Note: the full Karma suite currently fails later in unrelated existing MAM/MUC tests (src/plugins/muc-views/tests/mam.js) after the new audio test has already passed.

### DIFF
--- a/src/plugins/chatview/tests/message-audio.js
+++ b/src/plugins/chatview/tests/message-audio.js
@@ -23,7 +23,7 @@ describe('A Chat Message', function () {
     it(
         'will render audio streams using the accessible audio player',
         mock.initConverse(['chatBoxesFetched'], { fetch_url_headers: true }, async function (_converse) {
-            spyOn(window, 'fetch').and.callFake(async () => {
+            spyOn(window, 'fetch').and.callFake(() => {
                 return new Response('', {
                     status: 200,
                     headers: {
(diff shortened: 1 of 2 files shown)


jcbrand:
Hi @rafael226, can you please explain the purpose of this PR?

Is it possible to play already while the player is still loading?

rafael226:
Hi @jcbrand, yes. The purpose is to avoid blocking the user's play intent while the audio element is still loading metadata.

The PR doesn't try to force playback before the browser has enough data. It only keeps the custom Play button enabled/focusable during the loading state, so a mouse click or keyboard Enter can call `audio.play()`. The browser will then either start when enough data is available, or reject/handle playback according to its normal media rules.

Before this change, `is_loading` made the Play button disabled, so the control disappeared from the keyboard interaction path even though loading is a temporary state, not an error. The button still remains disabled for actual audio errors via `has_error`.

So the practical intent is: loading should show a spinner/status, but should not make the Play control unavailable.

PR #4074, opened 2026-06-03
Hide trusted login option when persistent storage is unavailable

### LINKED ISSUE #2405: Private tab - detect missing storage and disable/hide "trusted device" toggle
...else you are allowed to login and then it fails to do anything: `ERROR: No available storage method found.`

alan747271363-art:
﻿## Summary
- Add a persistent storage availability check before showing the trusted-device login checkbox.
- Preserve the existing `allow_user_trust_override: 'off'` behavior so the checkbox can still be shown unchecked by default.
- Add login form coverage and document the private browsing behavior.

Fixes #2405.

## Validation
- `npm ci`
- `$env:DROP_DEBUGGER='false'; npx.cmd rspack build --config rspack/rspack.headless.js --mode=development; npx.cmd rspack build --config rspack/rspack.build.js --mode=development; npm.cmd run build:converse-min-css`
- `npx.cmd eslint src/headless/utils/storage.js src/plugins/controlbox/templates/loginform.js src/plugins/controlbox/tests/login.js`
- `npm.cmd run types:check`
- `npx.cmd karma start karma.login.conf.js --single-run` (temporary local config containing only `src/plugins/controlbox/tests/login.js`; removed before commit)

Note: `npm.cmd test -- --single-run` hit an existing unrelated timeout in `src/plugins/chatview/tests/message-images.js` before reaching the login form tests.

### DIFF
--- a/src/headless/utils/storage.js
+++ b/src/headless/utils/storage.js
@@ -30,7 +30,11 @@ function storeUsesIndexedDB(type) {
 export function isPersistentStorageAvailable() {
     const store = settings.get('persistent_store');
     if (store === 'sessionStorage') {
-        return typeof window.sessionStorage !== 'undefined';
+        try {
+            return typeof globalThis.sessionStorage !== 'undefined';
+        } catch {
+            return false;
+        }
     } else if (store === 'BrowserExtLocal' || store === 'BrowserExtSync') {
         return true;
     }
--- a/src/headless/utils/storage.js
+++ b/src/headless/utils/storage.js
@@ -25,6 +25,26 @@ function storeUsesIndexedDB(type) {
 }
 
 /**
+ * @returns {boolean}
+ */
+export function isPersistentStorageAvailable() {
+    const store = settings.get('persistent_store');
+    if (store === 'sessionStorage') {
+        return typeof window.sessionStorage !== 'undefined';
+    } else if (store === 'BrowserExtLocal' || store === 'BrowserExtSync') {
+        return true;
+    }
+
+    const driver =
+        store === 'localStorage'
+            ? BrowserStorage.localForage.LOCALSTORAGE
+            : store === 'IndexedDB'
+              ? BrowserStorage.localForage.INDEXEDDB
+              : undefined;
+    return driver ? BrowserStorage.localForage.supports(driver) : true;
+}
+
+/**
  * @param {string} id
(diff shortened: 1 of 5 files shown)


src/plugins/controlbox/templates/loginform.js
```diff
@@ -1,12 +1,13 @@
...
-import { _converse, api, constants } from '@converse/headless';
+import { _converse, api, constants, converse } from '@converse/headless';
 import tplSpinner from 'templates/spinner.js';
 import { CONNECTION_STATUS_CSS_CLASS } from '../constants.js';
 import { __ } from 'i18n';
 import 'shared/components/brand-heading.js';
 import 'shared/components/footer.js';
 
 const { ANONYMOUS, EXTERNAL, LOGIN, PREBIND, CONNECTION_STATUS } = constants;
+const { u } = converse.env;
```
jcbrand: You can import `u` directly from `@converse/headless`.

src/headless/utils/storage.js
```diff
@@ -24,6 +24,26 @@ function storeUsesIndexedDB(type) {
     return type === 'persistent' && settings.get('persistent_store') === 'IndexedDB';
 }
 
+/**
+ * @returns {boolean}
+ */
+export function isPersistentStorageAvailable() {
+    const store = settings.get('persistent_store');
+    if (store === 'sessionStorage') {
+        return typeof window.sessionStorage !== 'undefined';
```
jcbrand: I'd like `@converse/headless` to eventually be usable in NodeJS.

For that to happen, we need to stop using `window`. Additionally, Safari in private browsing mode might throw a `SecurityError` so a `try/catch` needs to be used.

Something like this:

```
  if (store === 'sessionStorage') {
      try {
          return typeof globalThis.sessionStorage !== 'undefined';
      } catch {
          return false;
      }
  }
```

jcbrand (commented):
Hi @alan747271363-art Thanks for your PR.

Please see the review comments.

alan747271363-art:
Thanks for the review. I pushed an update that imports \u\ directly from \@converse/headless\ and changes the session storage availability check to use \globalThis.sessionStorage\ inside a try/catch, so it stays safer for headless/Node-like environments and restricted browser storage modes.

jcbrand:
Thanks!

PR #4075, opened 2026-06-04
fix: strip trailing slash from assets_path to prevent double-slash URLs

### LINKED ISSUE #4041: `assets_path` ending with `/` causes double-slash asset URLs and emoji.json fetch failure
I found the following while upgrading to 13.0.0 (but it likely affects other versions, too - I did not check).

Converse generates asset URLs with a double slash when `assets_path` ends with `/`.

For example, with the following configuration:

```javascript
assets_path: "/inverse/dist/"
```

Converse requests:
```
https://xmpp.igniterealtime.org:7483/inverse/dist//emoji.json
```

On my webserver, this results in:
```
HTTP 400: Ambiguous URI empty segment
```

The browser console shows:
```
GET https://xmpp.igniterealtime.org:7483/inverse/dist//emoji.json
[HTTP/1.1 400 Bad Request]

Failed to load emoji.json: Error: Failed to fetch emoji.json
```

Replacing the double slash with a single slash makes the request succeed.

The documentation currently shows `/dist/` as the default value for `assets_path`, which also ends with a trailing slash.

Converse should normalize asset URLs to avoid generating `//` in paths when `assets_path` ends with `/`.

khushigoel0101:
Fixes #4041

## Problem
When `assets_path` ends with `/` (e.g. `/dist/`), Converse generates 
double-slash URLs like `/dist//emoji.json` causing HTTP 400 errors.

## Fix
Strip trailing slash from `assets_path` before building asset URLs in:
- src/headless/plugins/emoji/api.js
- src/plugins/notifications/index.js
- src/shared/chat/utils.js

## Checklist
- [x] Added changelog entry in CHANGES.md
- [x] Added a test in src/plugins/chatview/tests/emojis.js

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ## 14.0.0 (Unreleased)
 
+- Fix: Strip trailing slash from `assets_path` to prevent double-slash asset URLs (fixes #4041)
+
 ### Backwards incompatible changes
 
 - `api.modal.show` no longer takes a modal instance as first parameter
--- a/src/headless/plugins/emoji/api.js
+++ b/src/headless/plugins/emoji/api.js
@@ -17,7 +17,7 @@ const emojis = {
 
             let json;
             try {
-                const path = api.settings.get('assets_path');
+                const path = api.settings.get('assets_path').replace(/\/$/, '');
                 const response = await fetch(`${path}/emoji.json`);
                 if (!response.ok) throw new Error('Failed to fetch emoji.json');
                 json = await response.json();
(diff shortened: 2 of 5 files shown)


jcbrand:
Hi @khushigoel0101, thank you for making a PR!

A better approach is to normalize the path when the setting is written, instead of doing it in multiple places when the setting is read.

Done here: https://github.com/conversejs/converse.js/commit/22f5f01368947aa2a97b3539455fc46de28d77a9

jcbrand:
Longer term we should be using a library like Zod or Valibot for the settings, but that's a bigger task.

khushigoel0101:
Thank you for the feedback. I understand now that normalizing at the point where the setting is written is much better than fixing it in multiple places. I learned a lot from this process. Looking forward to contributing more! :)

PR #4089, opened 2026-06-18
Feat/calls

pounceandmiss:
Introduce support for 1:1 audio calls over Jingle ([XEP-0167](https://xmpp.org/extensions/xep-0167.html)), set up with Jingle Message Initiation ([XEP-0353](https://xmpp.org/extensions/xep-0353.html)) and STUN/TURN discovery via [XEP-0215](https://xmpp.org/extensions/xep-0215.html). 
Tested on Firefox and Chromium, tested against Conversations and Dino.
Calls are initiated using the existing call button which is now enabled by default. 
Screenshot_2026-06-18_03-22-02


jcbrand:
Wow! Updates #447

pounceandmiss:
One thing I didn't add is a ringtone - not sure where/what sound to take.

jcbrand:
Thanks for your effort with this PR. I currently don't have any bandwidth to do a review, but I'll do so as soon as I can.

PR #4091, opened 2026-06-18
Clear the background class list before setting theme attributes.

kicken:
When changing the theme via the settings API, the background element did not clear out the previous theme class, resulting in theme CSS conflicts.

This change removes all `theme-` classes before applying the new theme, resolving the issue.


jcbrand:
Thanks!

Issue #326, opened 2015-03-07
XEP-0077: Ability to change XMPP account password


nodiscc:
Hello, I'd like my users to be able to change their XMPP passwords using converse.js. Is this already implemented/possible? AFAIK it depends on [XEP-0077](http://xmpp.org/extensions/xep-0077.html#usecases-changepw) which is already (partially?) implemented for in-band registration.

Any help is welcome

jcbrand:
In-band-registration is already supported, as you know, but not yet the resettting of passwords.

It's something I'd like to add but there's no timeline or deadline for it.

This feature will require some UI changes. We'll need to add a way to let users configure their accounts.
I'm not yet exactly sure what's the best way of exposing this option. I'm thinking of adding a "configure" icon next to the "close" icon (similar to the configure icon of chat rooms) which will render a configuration form where users can do things like change their passwords.

nodiscc:
XEP-0077 adds 3 available actions for the user: register, change password, and [cancel account](http://xmpp.org/extensions/xep-0077.html#usecases-cancel). I don't know if you want to add account cancellation, but even then it should fit in the main view (I'd keep account management next to roster management like this:)

![converseui](https://cloud.githubusercontent.com/assets/3950390/6545522/ccc1f26a-c58b-11e4-91c8-ddce5203c29a.png)

On clicking the gear icon, unfold this:

 ![manage](https://cloud.githubusercontent.com/assets/3950390/6545523/d03e2fc6-c58b-11e4-957f-2790c09f78f6.png)

jcbrand:
Thanks, I really like your suggestion. It's simple and logically grouped.

I think I'll go with that.

nodiscc:
I don't use the software anymore, cleaning up my issues list

licaon-kter:
Why close this if it's not fixed?

PandaScience:
Would love to see this feature as well (at least pw change). Any chance this will come in the near future?

PR #4099, opened 2026-06-27
feat: require current password when resetting password

### LINKED ISSUE #3120: Ask user to provide current password in order to reset password

lequangsang01:
## Summary

This PR addresses issue #3120 by adding a current password field to the password reset form. Currently, anyone with access to the browser can change the user's password without knowing the current one, which is a security concern.

## Changes

- Added a "Current password" input field to the password reset form
- Updated the XMPP IQ stanza to include the `<old_password>` field when submitting the password change request
- Added handling for `not-authorized` error responses (shown when the current password is incorrect)
- Updated existing tests to include the current password field
- Added a new test case for incorrect current password scenarios

## Technical Details

The form now requires three fields:
1. **Current password** (required) - authenticates the user before allowing password change
2. **New password** (required) - the new password to set
3. **Confirm new password** (required) - ensures the user typed the new password correctly

When the form is submitted, the XMPP registration query includes both the new password and the old password, allowing the server to validate the current credentials before processing the change.

## Related Issues

- Fixes #3120
- References #326

### DIFF
--- a/src/plugins/profile/templates/password-reset.js
+++ b/src/plugins/profile/templates/password-reset.js
@@ -3,12 +3,28 @@ import { html } from 'lit';
 
 export default el => {
     const i18n_submit = __('Submit');
+    const i18n_current_password = __('Current password');
     const i18n_passwords_must_match = __('The new passwords must match');
     const i18n_new_password = __('New password');
     const i18n_confirm_password = __('Confirm new password');
 
     return html`<form class="converse-form passwordreset-form" method="POST" @submit=${ev => el.onSubmit(ev)}>
         ${el.alert_message ? html`<div class="alert alert-danger" role="alert">${el.alert_message}</div>` : ''}
+
+        <div class="py-2">
+            <label for="converse_password_reset_current" class="form-label">${i18n_current_password}</label>
+            <input
+                class="form-control"
+                type="password"
+                value=""
+                name="current_password"
+                required="required"
+                id="converse_password_reset_current"
+                autocomplete="current-password"
+                minlength="8"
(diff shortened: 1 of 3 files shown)


src/plugins/profile/templates/password-reset.js
```diff
@@ -3,13 +3,29 @@ import { html } from 'lit';
...
+            <label for="converse_password_reset_current" class="form-label">${i18n_current_password}</label>
+            <input
+                class="form-control"
+                type="password"
+                value=""
+                name="current_password"
+                required="required"
+                id="converse_password_reset_current"
+                autocomplete="current-password"
+                minlength="8"
```
kaushanikoner20-hub, line 24: Should we remove `minlength="8"` from the current-password field? Existing accounts may have passwords shorter than eight characters, and client-side validation would prevent those users from changing their password even if the server accepts it.

kaushanikoner20-hub (commented):
Thanks for the implementation. I reviewed the form, the XMPP request, error handling, and test coverage. I left one question about client-side minimum-length validation for existing passwords.

jcbrand:
Thanks for taking a look at #3120. Unfortunately I can't merge this, because the approach isn't allowed by XEP-0077, and as written it wouldn't actually protect anyone. 

`<old_password/>` is not a valid element in the `jabber:iq:register` namespace

`old_password` exists in XEP-0077 only as a data form field var, registered under the `jabber:iq:register:changepassword` `FORM_TYPE` (https://xmpp.org/extensions/xep-0077.html#registrar-formtypes-changepassword)
  
So the spec's actual mechanism is a server-driven challenge (via data-form).

PR #4102, opened 2026-07-03
doap: Add status

yurtpage:
The doap.xml does not always shows a completion status of the implementation
https://xmpp.org/software/converse-js/
I checked manually dozen of XEP and it looks like all of them are completed or almost completed. So I put the completed status to all XEPs that didn't had the status.


jcbrand:
Thanks!

Issue #4110, opened 2026-07-06
Intermittently missing new message count icon


dca00:
**Describe the bug**
The new message count icon in the roster is missing intermittently. This is net new in 13.

**To Reproduce**
1. Wait to get messages
2. See only the red icon in the tab but no green icon in the roster

**Expected behavior**
There should always be the red icon in the tab and the green new message count icon in the roster, when the user is not in the chat with the originator of the new messages.

**Screenshots**

**Environment (please complete the following information):**
 - Desktop or Mobile? Desktop
 - Browser [Firefox]
 - Converse.js version [e.g. 13.1.0]

**Additional context**
It may be depending on the originating client or totally random, I haven't made an observation yet.

abhiramvsmg:
Hey, I'd like to take a shot at fixing this. I'll dig into how the tab icon and roster badge get updated for incoming messages and see where they diverge. Will update here once I have something.

abhiramvsmg:
The Problem: The tab icon and the roster badge update through different paths. The tab icon works perfectly because it listens directly to the chat's unread counter. The roster badge, however, depends on a linked contact object that loads asynchronously. If a message comes in before that contact object finishes loading, the badge update is silently skipped and never catches up.

How we fixed it: In 
model-with-contact.js , we added a call to updateContactUnreadCounter right when the contact object finishes loading in  setModelContact.
This closes the timing window: even if a message comes in early, the badge synchronizes as soon as the contact is ready.

How we tested it: Because public XMPP server registration was down, we wrote an automated integration test in  chat.js . The test intercepts the contact fetch, delays it, sets the unread count, and verifies that the unread count synchronizes to the contact badge immediately after the contact resolves.

PR #4111, opened 2026-07-08
Fix intermittent missing new message count badge (Issue #4110)

abhiramvsmg:
This fixes a timing race condition where the roster badge fails to update with new unread messages if they arrive before the linked roster contact object has finished loading asynchronously.

Changes :

Fix : Added a call to `this.updateContactUnreadCounter()` inside `setModelContact` in `src/headless/shared/model-with-contact.js` to sync the unread counter as soon as the contact resolves.

Test : Added a regression integration test in `src/headless/plugins/chat/tests/chat.js` that mock-delays the contact load, sets the unread count, and asserts that the contact's unread badge synchronizes correctly upon resolution.

Changelog : Added an entry for Issue #4110 in `CHANGES.md`.

### DIFF
--- a/src/headless/shared/model-with-contact.js
+++ b/src/headless/shared/model-with-contact.js
@@ -67,6 +67,7 @@ export default function ModelWithContact(BaseModel) {
 
             if (contact) {
                 this.contact = contact;
+                this.updateContactUnreadCounter();
                 this.set('nickname', contact.get('nickname'));
 
                 this.listenTo(this.contact, 'vcard:add', (changed) => {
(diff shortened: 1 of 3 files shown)


src/headless/shared/model-with-contact.js
```diff
@@ -67,6 +67,7 @@ export default function ModelWithContact(BaseModel) {
 
             if (contact) {
                 this.contact = contact;
+                this.updateContactUnreadCounter();
```
jcbrand, line 70: Further down in this `if` statement the `contact:add` event is triggered and the `ModelWithContact` model already listens for that event and then calls `updateContactUnreadCounter()`.

So this change here look unnecessary.

Also, the test passes without this fix.

jcbrand (changes requested):

PR #4116, opened 2026-07-17
fix: ask for current password in password reset flow

### LINKED ISSUE #3120: Ask user to provide current password in order to reset password

gtmdev-br:
## Summary

This PR adds a current password field to the password reset form, as requested in #3120. The change ensures that users must provide their current password before being allowed to reset it, preventing unauthorized password changes by someone with access to the browser.

## Changes

1. **Template** (`src/plugins/profile/templates/password-reset.js`):
   - Added a new "Current password" input field with proper `autocomplete="current-password"` attribute

2. **Component** (`src/plugins/profile/password-reset.js`):
   - Added `current_password_error` state property
   - Added validation to ensure current password is provided
   - Handle `not-authorized` error from XMPP server when current password is incorrect

3. **Tests** (`src/plugins/profile/tests/password-reset.js`):
   - Updated `submitPasswordResetForm` helper to include current password
   - Added new test case for incorrect current password scenario

## Testing

All 4 password reset tests pass:
- allows you to reset your password
- informs you if you cannot reset your password due to registration not being supported
- informs you if you're not allowed to reset your password
- informs you if the current password is incorrect (NEW)

Closes #3120

### DIFF
--- a/src/plugins/profile/password-reset.js
+++ b/src/plugins/profile/password-reset.js
@@ -11,13 +11,15 @@ class PasswordReset extends CustomElement {
     static get properties () {
         return {
             passwords_mismatched: { type: Boolean },
-            alert_message: { type: String }
+            alert_message: { type: String },
+            current_password_error: { type: String }
         }
     }
 
     initialize () {
         this.passwords_mismatched = false;
         this.alert_message = '';
+        this.current_password_error = '';
     }
 
     render () {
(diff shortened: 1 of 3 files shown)


jcbrand:
Thanks for tackling #3120, but I'm closing this.

The current password is collected in the form and never actually sent or verified anywhere and the `not-authorized` mapping doesn't match XEP-0077.

PR #4164, opened 2026-09-11
fix(ui): keep fullscreen chats and heading controls on-screen

mikejw:
## Summary
- After the app-switcher rename, narrow fullscreen CSS targeted a nested `converse-app-chat` that does not exist, so opening a room left the roster at `100vw` and the chat off-screen.
- Size the chat column beside the switcher (`flex: 1; min-width: 0`) and use `100%` instead of `100vw` so rooms overlay the roster and heading/section hamburgers stay on screen.
- Give enabled buttons a pointer cursor and turn the room back control into a real button.

## Test plan
- [ ] Fullscreen, viewport under ~768px: click a room and land in the chat (back arrow in the header).
- [ ] Hamburgers visible on Groupchats / Contacts and on the room heading; menus open and are clickable.
- [ ] Back arrow returns to the roster and shows a pointer.
- [ ] Widen past ~768px: roster + chat side by side still works.

This change was made with AI assistance.


jcbrand:
Merged manually, thanks!

